PR #35, opened 2024-04-03
Remove emit_propagated_event

dmulcahey:


dmulcahey:
line wasn't added it was modified.

PR #39, opened 2024-04-05
Remove sending HS color commands to lights

TheJulianJES:
## Proposed change

This removes sending "move to hue/saturation" commands to lights supporting HS when the "always prefer xy mode" option is disabled. By default, it is already disabled.

The config option is now also removed and ZHA only sends XY color commands to color lights.
There are no benefits to addressing the light with HS color calls, only downsides:
- color is off in many/all lights I've tested with compared to XY (many colors can't be produced)
- doesn't match the colors in the HA color picker at all (tested with Philips Hue lights)
- all color lights support XY, only some support HS, even less support enhanced HS
- the config option to always prefer XY mode (over HS when possible) is already enabled by default, so HS is never used in default ZHA installations (due to the mentioned downsides)
- Z2M only seems to use XY color commands when addressing Zigbee lights via HA, only via their own web UI can you send a predefined set of HS colors (or via direct MQTT messages)
- supporting XY, HS, and enhanced HS color commands brings a bunch of (unnecessary) maintenance work (+ groups)

Do note that HA will convert HS args given to `light.turn_on` to XY colors, as we set "supported color modes" correctly.
So, the functionality of being able to set the color of lights via HS colors doesn't break. They just get converted to XY first.

## Further considerations

The unfinished test `test_zha_group_light_entity` is still unfinished and the HS attribute cache initialization probably doesn't make much sense to test now.

This also requires removing the `CONF_ALWAYS_PREFER_XY_COLOR_MODE` option from the ZHA websocket API in HA.
I've not checked how HA deals with the fact if somebody has set `CONF_ALWAYS_PREFER_XY_COLOR_MODE` to `False` in their config. If it breaks, we should fix that.. and remove the option from the config.

For the few people that (accidentally) disabled the "Always prefer XY color mode" option in ZHA, the state might be shown incorrect in HA until the light is initialized/polled. This should happen shortly after HA startup.
Even if the light is in "HS color mode" then, polling will revert the light color mode attribute to XY as expected and use the correct XY values.


TheJulianJES:
Updated the PR. I did not test anything yet, but everything should still work (and it's somewhat covered at this point?).
There are a few more removals now (e.g. removing HS properties in the cluster handler). We can also keep them for now if that's preferred.

- ZHA in HA will also need to be adapted to remove `hs_color` references in the [light class](https://github.com/home-assistant/core/blob/dev/homeassistant/components/zha/light.py).
- The config option in the frontend should also be removed. We might have to change existing configs too (or just ignore?)
- There are some incomplete light tests around HS lights that should be finished in the future (with XY) or be removed
- ...more?

As a note, you can still pass HS colors to the `light.turn_on` service. HA should convert them to XY for ZHA depending on the color capabilities we report.
So, there should be no real loss from removing "ZCL HS support". Referring back to the PR description, they seem to be completely off compared to the HA color picker.

puddly:
Tested this locally and it works fine:

```diff
diff --git a/homeassistant/components/zha/const.py b/homeassistant/components/zha/const.py
index 3986a99cf3f..0e62837782e 100644
--- a/homeassistant/components/zha/const.py
+++ b/homeassistant/components/zha/const.py
@@ -43,7 +43,6 @@ CONF_CUSTOM_QUIRKS_PATH = "custom_quirks_path"
 CONF_DEFAULT_LIGHT_TRANSITION = "default_light_transition"
 CONF_ENABLE_ENHANCED_LIGHT_TRANSITION = "enhanced_light_transition"
 CONF_ENABLE_LIGHT_TRANSITIONING_FLAG = "light_transitioning_flag"
-CONF_ALWAYS_PREFER_XY_COLOR_MODE = "always_prefer_xy_color_mode"
 CONF_GROUP_MEMBERS_ASSUME_STATE = "group_members_assume_state"
 
 CONF_ENABLE_IDENTIFY_ON_JOIN = "enable_identify_on_join"
diff --git a/homeassistant/components/zha/helpers.py b/homeassistant/components/zha/helpers.py
index 35a794e8631..12e95d10c7e 100644
--- a/homeassistant/components/zha/helpers.py
+++ b/homeassistant/components/zha/helpers.py
@@ -140,7 +140,6 @@ from .const import (
     CONF_ALARM_ARM_REQUIRES_CODE,
     CONF_ALARM_FAILED_TRIES,
     CONF_ALARM_MASTER_CODE,
-    CONF_ALWAYS_PREFER_XY_COLOR_MODE,
     CONF_BAUDRATE,
     CONF_CONSIDER_UNAVAILABLE_BATTERY,
     CONF_CONSIDER_UNAVAILABLE_MAINS,
@@ -1156,7 +1155,6 @@ CONF_ZHA_OPTIONS_SCHEMA = vol.Schema(
         ),
         vol.Required(CONF_ENABLE_ENHANCED_LIGHT_TRANSITION, default=False): cv.boolean,
         vol.Required(CONF_ENABLE_LIGHT_TRANSITIONING_FLAG, default=True): cv.boolean,
-        vol.Required(CONF_ALWAYS_PREFER_XY_COLOR_MODE, default=True): cv.boolean,
         vol.Required(CONF_GROUP_MEMBERS_ASSUME_STATE, default=True): cv.boolean,
         vol.Required(CONF_ENABLE_IDENTIFY_ON_JOIN, default=True): cv.boolean,
         vol.Optional(
@@ -1231,7 +1229,6 @@ def create_zha_config(hass: HomeAssistant, ha_zha_data: HAZHAData) -> ZHAData:
         enable_light_transitioning_flag=zha_options.get(
             CONF_ENABLE_LIGHT_TRANSITIONING_FLAG
         ),
-        always_prefer_xy_color_mode=zha_options.get(CONF_ALWAYS_PREFER_XY_COLOR_MODE),
         group_members_assume_state=zha_options.get(CONF_GROUP_MEMBERS_ASSUME_STATE),
     )
     device_options: DeviceOptions = DeviceOptions(
diff --git a/homeassistant/components/zha/strings.json b/homeassistant/components/zha/strings.json
index 5d81556564a..8b6754a2422 100644
--- a/homeassistant/components/zha/strings.json
+++ b/homeassistant/components/zha/strings.json
@@ -178,7 +178,6 @@
       "title": "Global Options",
       "enhanced_light_transition": "Enable enhanced light color/temperature transition from an off-state",
       "light_transitioning_flag": "Enable enhanced brightness slider during light transition",
-      "always_prefer_xy_color_mode": "Always prefer XY color mode",
       "group_members_assume_state": "Group members assume state of group",
       "enable_identify_on_join": "Enable identify effect when devices join the network",
       "default_light_transition": "Default light transition time (seconds)",
diff --git a/tests/components/zha/data.py b/tests/components/zha/data.py
index eb135c7e8fe..b50c7a18a06 100644
--- a/tests/components/zha/data.py
+++ b/tests/components/zha/data.py
@@ -23,12 +23,6 @@ BASE_CUSTOM_CONFIGURATION = {
                 "required": True,
                 "default": True,
             },
-            {
-                "type": "boolean",
-                "name": "always_prefer_xy_color_mode",
-                "required": True,
-                "default": True,
-            },
             {
                 "type": "boolean",
                 "name": "group_members_assume_state",
@@ -62,7 +56,6 @@ BASE_CUSTOM_CONFIGURATION = {
             "enhanced_light_transition": True,
             "default_light_transition": 0,
             "light_transitioning_flag": True,
-            "always_prefer_xy_color_mode": True,
             "group_members_assume_state": False,
             "enable_identify_on_join": True,
             "consider_unavailable_mains": 7200,
@@ -94,12 +87,6 @@ CONFIG_WITH_ALARM_OPTIONS = {
                 "required": True,
                 "default": True,
             },
-            {
-                "type": "boolean",
-                "name": "always_prefer_xy_color_mode",
-                "required": True,
-                "default": True,
-            },
             {
                 "type": "boolean",
                 "name": "group_members_assume_state",
@@ -154,7 +141,6 @@ CONFIG_WITH_ALARM_OPTIONS = {
             "enhanced_light_transition": True,
             "default_light_transition": 0,
             "light_transitioning_flag": True,
-            "always_prefer_xy_color_mode": True,
             "group_members_assume_state": False,
             "enable_identify_on_join": True,
             "consider_unavailable_mains": 7200,
```

TheJulianJES:
Nice! Thanks for testing. We might also need to migrate old configs in HA where the key is still present, as config validation will fail otherwise(?)
That should be relatively simple though, we can go ahead with this one if we can get to the HA part in time.

TheJulianJES:
The changes to HA above should be enough now, as https://github.com/home-assistant/core/pull/125710 removes unused keys from the HA config, so we'd not fail validation if the option is present.

I think we could go ahead with this now.

TheJulianJES:
When merging these changes into HA, it's important to add this commit to HA: https://github.com/home-assistant/core/pull/124804/commits/8fb7789ea8ddb6ed2a287aed5010374c0452f6c9
AND to remove all pre-existing `hs_color` references in the HA light class.
Especially important is to remove passing `hs_color` in `restore_external_state_attributes`, as stuff will break otherwise. That's missing in the above commit.

PR #43, opened 2024-04-06
Use config dataclasses instead of HA dicts

dmulcahey:


tests/test_discover.py
```diff
@@ -502,7 +498,7 @@ async def test_device_override(
 
 
 async def test_quirks_v2_entity_discovery(
-    zha_gateway: Gateway,
+    zha_gateway: Gateway,  # pylint: disable=unused-argument
```
puddly, line 501: Maybe we should disable `unused-argument` and `redefined-outer-name` for any file in `tests/`?

dmulcahey: on the fence. the `redefined-outer-name` one saved me a couple times when converting. I am mainly adding the hints to the files when I know they aren't an issue so I don't get distracted by them in vscode.  I'll defer to consensus here.

PR #48, opened 2024-05-08
OTA fixes (and more)

puddly:
This PR tracks minor fixes I had to make to get things running.

Corresponding Core branch: https://github.com/dmulcahey/home-assistant/compare/dm/zha-external-zha...puddly:core:puddly/zha-external-zha

### DIFF
--- a/zha/application/platforms/button/__init__.py
+++ b/zha/application/platforms/button/__init__.py
@@ -132,6 +132,7 @@ class IdentifyButton(Button):
 
     _attr_device_class = ButtonDeviceClass.IDENTIFY
     _attr_entity_category = EntityCategory.DIAGNOSTIC
+    _attr_translation_key = "identify"
     _command_name = "identify"
     _kwargs = {}
     _args = [DEFAULT_DURATION]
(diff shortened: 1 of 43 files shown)


puddly:
Some entity names are a little messed up:

image

zha/application/platforms/button/__init__.py
```diff
@@ -132,6 +132,7 @@ def create_platform_entity(
 
     _attr_device_class = ButtonDeviceClass.IDENTIFY
     _attr_entity_category = EntityCategory.DIAGNOSTIC
+    _attr_translation_key = "identify"
```
TheJulianJES: I don't think this should have a translation key for HA. It already has the "identify" device class which is enough for the name. Same as battery entities (only use device class, not another explicit "Battery" name/translation).

puddly: You're right, this can be removed without issue.

I think originally I was setting `_attr_name = None`, which again defaulted to the device name. Fixed by not setting the attribute unless it has a value.

dmulcahey (approved):

PR #58, opened 2024-07-01
Implement external state restoration

puddly:
A few platforms in ZHA uses instance attributes and complex logic to derive their state, relying on Home Assistant to persist it and restore it after startup:

- I've exposed necessary information for some platforms (`light` and `cover`) for Home Assistant to store this information as extra state attributes, allowing it to be restore properly. I believe this was actually a bug in the past.
- Other platforms could be converted to use the ZCL attribute cache, such as `binary_sensor` and `lock`. These effectively perform operations that change a ZCL attribute so the attribute cache can be used to maintain state.
- Others rely on directly computing `state` from other information and unfortunately will have to be directly restored (`siren`).

In the future, we need to explore a way to store this extra state. I think zigpy can provide a hook. Alternatively, we can invert the storage logic and have ZHA implement a database driver for zigpy.

### DIFF
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -1078,24 +1078,15 @@ class Light(PlatformEntity, BaseLight):
         effect: str | None,
     ) -> None:
         """Restore extra state attributes that are stored outside of the ZCL cache."""
-        if state is not None:
-            self._state = state
-        if off_with_transition is not None:
-            self._off_with_transition = off_with_transition
-        if off_brightness is not None:
-            self._off_brightness = off_brightness
-        if brightness is not None:
-            self._brightness = brightness
-        if color_temp is not None:
-            self._color_temp = color_temp
-        if xy_color is not None:
-            self._xy_color = xy_color
-        if hs_color is not None:
-            self._hs_color = hs_color
-        if color_mode is not None:
-            self._color_mode = color_mode
-        if effect is not None:
-            self._effect = effect
+        self._state = state
+        self._off_with_transition = off_with_transition
+        self._off_brightness = off_brightness
+        self._brightness = brightness
+        self._color_temp = color_temp
+        self._xy_color = xy_color
+        self._hs_color = hs_color
+        self._color_mode = color_mode
+        self._effect = effect
 
 
 @STRICT_MATCH(
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -61,9 +61,9 @@ from zha.application.platforms.light.const import (
     LightEntityFeature,
 )
 from zha.application.platforms.light.helpers import (
+    ZCL_TO_ENTITY_COLOR_MODE,
     brightness_supported,
     filter_supported_color_modes,
-    zcl_color_mode_to_entity_color_mode,
 )
 from zha.application.registries import PLATFORM_ENTITIES
 from zha.debounce import Debouncer
@@ -791,9 +791,10 @@ class Light(PlatformEntity, BaseLight):
                     self._effect = EFFECT_COLORLOOP
 
             if self._color_cluster_handler.color_mode is not None:
-                self._color_mode = zcl_color_mode_to_entity_color_mode(
+                self._color_mode = ZCL_TO_ENTITY_COLOR_MODE[
                     self._color_cluster_handler.color_mode
-                )
+                ]
+
         self._external_supported_color_modes = filter_supported_color_modes(
             self._supported_color_modes
         )
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -1093,10 +1093,15 @@ class Light(PlatformEntity, BaseLight):
 
             self.maybe_emit_state_changed_event()
 
-    def restore_extra_state_attributes(
-        self, *, off_with_transition: bool | None, off_brightness: int | None
+    def restore_external_state_attributes(
+        self,
+        *,
+        state: bool | None,
+        off_with_transition: bool | None,
+        off_brightness: int | None,
     ) -> None:
         """Restore extra state attributes that are stored outside of the ZCL cache."""
+        self._state = state
         self._off_with_transition = off_with_transition
         self._off_brightness = off_brightness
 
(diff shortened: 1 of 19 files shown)


zha/zigbee/cluster_handlers/lighting.py
```diff
@@ -65,7 +63,7 @@ class ColorClusterHandler(ClusterHandler):
         Color.AttributeDefs.options.name: True,
     }
 
-    @cached_property
```
dmulcahey: Is color_capabilities changing?

puddly: Ah, I think this was a hack to get the unit test to pass. Fixed, thanks!

zha/application/platforms/light/__init__.py
```diff
@@ -1190,6 +1231,12 @@ def available(self) -> bool:
         """Return entity availability."""
         return self._available
 
+    def _persist_color_mode(self, color_mode: ZclColorMode) -> None:
+        """Persist the color mode."""
+
+        # FIXME: Groups use raw clusters, not cluster handlers
```
TheJulianJES: Is "FIXME" needed here?
At least persisting the color mode shouldn't be an issue for groups. When initializing, the LightGroup should `update()` its status by checking all the member statuses which correctly persist/restore the color mode (either being persisted through ZCL cache or if it were to be restored from HA. Although the `__init__` `update()` call might be done a bit too early?).

(I need to find the code for how the group is notified of child updates.. -> there might be another issue if a debouncer isn't used)

There's also no attribute cache for groups, since that wouldn't make a ton of sense, so I'm not sure the "raw clusters" vs "cluster handlers" comment makes much sense for this use-case(?) Or am I missing something?

zha/application/platforms/light/__init__.py
```diff
@@ -1064,6 +1093,18 @@ def _assume_group_state(self, update_params) -> None:
...
 
+    def restore_external_state_attributes(
+        self,
+        *,
+        state: bool | None,
+        off_with_transition: bool | None,
+        off_brightness: int | None,
+    ) -> None:
+        """Restore extra state attributes that are stored outside of the ZCL cache."""
+        self._state = state
```
TheJulianJES: When `state` is passed from HA to this, we'd need to make sure it's actually a `bool | None`, not a `str`. 
HA uses `"off"`, `"on"`, `"unknown"`, `"unavailable"`. So, [`STATE_ON`, ...](https://github.com/home-assistant/core/blob/61f1c8d9631bab1a4236eee405c5b0b03e8be22e/homeassistant/const.py#L463-L492)

For reference, current code for restoring on/off state in HA Core ZHA [here.](https://github.com/home-assistant/core/blob/61f1c8d9631bab1a4236eee405c5b0b03e8be22e/homeassistant/components/zha/light.py#L849).

zha/application/platforms/light/__init__.py
```diff
@@ -787,20 +803,32 @@ def __init__(
...
+        self._external_supported_color_modes = filter_supported_color_modes(
+            self._supported_color_modes
+        )
+
+        if self._color_mode == ColorMode.UNKNOWN:
+            if len(self._external_supported_color_modes) == 1:
+                self._color_mode = next(iter(self._external_supported_color_modes))
             else:
-                self._color_mode = ColorMode.XY
+                # Light supports color_temp + hs, determine which mode the light is in
```
TheJulianJES: The comment seems to be outdated (before and now). This is a leftover from when HA only provided HS and we converted it to XY manually.

```suggestion
                # Light supports color_temp + xy, determine which mode the light is in
```


(Also, it seems like we completely ignore the light possibly being in HS color mode here. But seeing as it's an opt-in option, I'd not bother with it for now.
I'd just get rid of HS support entirely sooner or later, since all colors are "broken" there and Z2M also doesn't offer it for HA: just converts HS to XY to send to light, which we also do by default.)

zha/application/platforms/light/__init__.py
TheJulianJES: Also, I think `effect` also needs to be restored.

TheJulianJES: Furthermore, I'd also consider restoring `brightness`, `color_temp`, `xy_color`, (`hs_color`, ~~`color_mode`~~) for now.

I'd prefer a different solution too, especially since it's only really needed for the older Hue lights to not break if HA is restarted before they had a chance to be polled after a color change, but seeing as [Core ZHA currently does it](https://github.com/home-assistant/core/blob/61f1c8d9631bab1a4236eee405c5b0b03e8be22e/homeassistant/components/zha/light.py#L846-L865) and we have the "framework" for it now, maybe also restore that?

zha/application/platforms/light/__init__.py
puddly: The fact that cluster handlers are bypassed is a bug, IMO, since the parent class expects the cluster handler to *be* one, not a cluster instance. We filter most commands through cluster handlers but groups bypass that and wire straight into zigpy. It might be good in the future to create a group cluster handler.

zha/application/platforms/light/__init__.py
puddly: It's converted into a `bool` within the Core code: <https://github.com/home-assistant/core/blob/27bd52d26572b4a7d329c44a1b40a6271c5af6d9/homeassistant/components/zha/light.py#L166-L168>. I'm hoping static type checking will pick up invalid invocations.

puddly:
I've foregone using the ZCL cache and instead just implement `restore_external_state_attributes` for the required entities. This makes this PR much simpler.

zha/application/platforms/number/__init__.py
```diff
@@ -141,14 +141,13 @@ def native_step(self) -> float | None:
         """Return the value step."""
         return self._analog_output_cluster_handler.resolution
 
-    @functools.cached_property
+    @property
```
TheJulianJES: To confirm, you changed this intentionally because the ZCL `description` attribute could theoretically change?

puddly: Our use of `property` vs `cached_property` is a little inconsistent right now so I think either one works. Unit tests aren't affected so I'll revert the change.

TheJulianJES (approved):
Also looks good IMO

PR #59, opened 2024-07-02
Remove some mypy excludes

puddly:
Typing is not actually enabled because we do not expose `py.typed`.

Progress:

- [ ] `arg-type`
- [ ] `assignment`
- [ ] `attr-defined`
- [x] `call-arg`
- [x] `dict-item`
- [x] `index`
- [x] `misc`
- [ ] `no-any-return`
- [ ] `no-untyped-call`
- [ ] `no-untyped-def`
- [x] `override`
- [x] `return-value`
- [ ] `union-attr`
- [x] `var-annotated`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,13 +63,13 @@ install_types = true
 non_interactive = true
 
 disable_error_code = [
-    "no-untyped-def",   # 348
-    "attr-defined",     # 172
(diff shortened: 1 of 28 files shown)


puddly:
Save for deleted constants, these changes should all be lib-internal.

dmulcahey (approved):
:shipit:

PR #60, opened 2024-07-05
Emit state change events for sensors not tied to an attribute

puddly:
Sensors not tied to an attribute (such as the Sinope thermostat HVAC mode) did not emit an update event when their state changed.

This fixes all commented out ZHA unit tests.


TheJulianJES (approved):
Looks good

PR #63, opened 2024-07-06
Add `is_on` property to siren

TheJulianJES:
This adds a `is_on` property to the siren platform to align it with the rest of the on/off platforms we have.
The change will allow us to address this comment: https://github.com/home-assistant/core/pull/120190#discussion_r1667428899

The added property is used in the `state` property, like done by other platforms (and thus covered by tests already).

### DIFF
--- a/zha/application/platforms/siren.py
+++ b/zha/application/platforms/siren.py
@@ -113,8 +113,13 @@ class Siren(PlatformEntity):
     def state(self) -> dict[str, Any]:
         """Get the state of the siren."""
         response = super().state
-        response["state"] = self._attr_is_on
+        response["state"] = self.is_on
         return response
+
+    @property
+    def is_on(self) -> bool:
+        """Return true if the entity is on."""
+        return self._attr_is_on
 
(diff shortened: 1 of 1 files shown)


puddly (approved):
LGTM!

In the future, I wonder if it would be good to drop `state` and instead expose individual state attributes (i.e. `state` -> `is_on`)?

TheJulianJES:
Currently, we need the "whole state" for `maybe_emit_state_changed_event` (and maybe elsewhere?): <https://github.com/zigpy/zha/blob/378e53641d02a450a99a67dccf79e97324196d77/zha/application/platforms/__init__.py#L249-L250>

The naming of the individual entries seems to be a bit inconsistent though. Sometimes, the "on/off" state is called [`"on"`](https://github.com/zigpy/zha/blob/378e53641d02a450a99a67dccf79e97324196d77/zha/application/platforms/light/__init__.py#L161). But most of the times, like here, it's just `"state"`. I guess because that's the only real state of the entity, while light has more "states".

PR #71, opened 2024-07-09
Fix supported_features type on for non group lights

dmulcahey:
SSIA


puddly (approved):

puddly:
We should also start pulling shared features into the base entity subclass. Otherwise, I think type checking won't help.

PR #73, opened 2024-07-09
Expose functionality shared only by all members of a light group

puddly:
The main change is intersecting the features instead of unioning them: a light group should expose only the common functionality.


puddly:
This may need a broader discussion. We union together color modes and effects, I think this behavior is not correct and we should instead intersect.

TheJulianJES:
Hmm, it would definitely be a breaking change.
For example, I could see combining brightness only lights with RGB(-CCT) lights to be somewhat "common". You'd still want the color control for the lights that support it if they're grouped.

PR #77, opened 2024-07-10
Use main zigpy CI branch

puddly:
CI was running from `@dm/update-ci-03272024`, not `@main`.


TheJulianJES:
That's probably because of https://github.com/zigpy/workflows/pull/15.
I think it should be fine to merge, but haven't tested it with the other zigpy repos.

puddly:
I run most of them with `xdist` locally anyways, it should be good.

PR #78, opened 2024-07-11
Fix light groups not restoring `off_brightness`

TheJulianJES:
### Background info

At the moment, we do not restore light group state at all, which is fine generally.
However, `off_brightness` and `off_with_transition` are not calculated from the member state at the moment.
So, we need to restore them to make sure that a group light turned off with a transition still turns on to the correct brightness after a restart (and not completely dim).

(yet to this PR in the "real-world", but it should be fine and is unit-tested at least)

### Change

For that reason, this PR restores `off_brightness` and `off_with_transition` for light groups.
The test is also modified to test for the restored `off_brightness` and verifies `off_with_transition` is still `False` after the restore.

### DIFF
--- a/tests/test_light.py
+++ b/tests/test_light.py
@@ -823,7 +823,8 @@ async def test_zha_group_light_entity(
     # test that the lights were created and are off
     assert bool(entity.state["on"]) is False
 
-    # Group entities do not support state restoration
+    # Group entities do not support state restoration,
+    # except for off_brightness and off_with_transition
     entity.restore_external_state_attributes(
         state=True,
         off_with_transition=False,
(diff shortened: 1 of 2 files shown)


tests/test_light.py
```diff
@@ -837,6 +838,8 @@ async def test_zha_group_light_entity(
     )
 
     assert bool(entity.state["on"]) is False
+    assert bool(entity.state["off_with_transition"]) is False
```
TheJulianJES, line 841: `off_with_transition is False` is also tested here, but it's `False` by default anyway, even if not restored.
If we'd want to fully verify that it also restores correctly too, we'd need to modify the test further (or split it out into a separate one), as the code below expects the `on` cluster command on the `OnOff` cluster. That doesn't happen if `off_with_transition` is `True` though.
Still, the modified test checks that `off_with_transition` stays `False`. (and the `off_brightness` check below is fully correct and would fail if it weren't restored. It would be `None` before this PR.)

IMO, it's fine like this, but can change if wanted.

dmulcahey (approved):

PR #79, opened 2024-07-12
Rename update to async_update for consistency

dmulcahey:
SSIA

### DIFF
--- a/zha/application/platforms/__init__.py
+++ b/zha/application/platforms/__init__.py
@@ -421,7 +421,7 @@ class GroupEntity(BaseEntity):
             _LOGGER,
             cooldown=update_group_from_member_delay,
             immediate=False,
-            function=self.update,
+            function=self.async_update,
         )
         self._group.register_group_entity(self)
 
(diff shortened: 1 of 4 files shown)


puddly:
Does this need any changes in Core?

dmulcahey:
> Does this need any changes in Core?

I still have to look but I don’t think so.

TheJulianJES (approved):
Shouldn't need any changes in core: [homeassistant/components/zha/entity.py#L137-L141](https://github.com/home-assistant/core/blob/162b734be756d48e5e36ec828e6703110b7e2f59/homeassistant/components/zha/entity.py#L137-L141)
(and don't see `update()` being called anywhere else either)

dmulcahey:
> Shouldn't need any changes in core: [homeassistant/components/zha/entity.py#L137-L141](https://github.com/home-assistant/core/blob/162b734be756d48e5e36ec828e6703110b7e2f59/homeassistant/components/zha/entity.py#L137-L141) (and don't see `update()` being called anywhere else either)

thx. I just double checked as well and I agree. 

image

PR #81, opened 2024-07-13
Fix bindable cluster matching

dmulcahey:
The `BINDABLE_CLUSTERS` registry was being redefined in the `helpers` module because of a circular import issue. This was an oversight from the conversion and caused the registry to always be empty.

### DIFF
--- a/tests/test_application_helpers.py
+++ b/tests/test_application_helpers.py
@@ -108,5 +108,4 @@ async def test_get_matched_clusters(
     assert matches[0].target_ieee == zha_device.ieee
     assert matches[0].target_ep_id == 1
 
-    matches = await get_matched_clusters(not_bindable_zha_device, remote_zha_device)
-    assert not matches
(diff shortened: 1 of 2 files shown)


tests/test_application_helpers.py
```diff
@@ -0,0 +1,112 @@
...
+    assert len(matches) == 1
+    assert (
+        matches[0].source_cluster
+        == remote_zha_device.device.endpoints[1].out_clusters[OnOff.cluster_id]
+    )
+    assert matches[0].target_ieee == zha_device.ieee
+    assert matches[0].target_ep_id == 1
+
+    matches = await get_matched_clusters(not_bindable_zha_device, remote_zha_device)
+    assert not matches
```
TheJulianJES: Could also do this here, but also fine how it currently is:
```suggestion
    assert not await get_matched_clusters(not_bindable_zha_device, remote_zha_device)
```

TheJulianJES (approved):

PR #86, opened 2024-07-15
Allow translations for enum values in EnumSensor and ZHAEnumSelectEntity entities

Caius-Bonus:
This is the counterpart of the original in homeassistant/core: https://github.com/home-assistant/core/pull/109309

### DIFF
--- a/zha/application/platforms/select.py
+++ b/zha/application/platforms/select.py
@@ -485,7 +485,7 @@ class AqaraT2RelaySwitchType(ZCLEnumSelectEntity):
     _unique_id_suffix = "switch_type"
     _attribute_name = "switch_type"
     _enum = T2RelayOppleCluster.SwitchType
-    _attr_translation_key: str = "switch_type"
+    _attr_translation_key: str = "relay_switch_type"
 
 
 @CONFIG_DIAGNOSTIC_MATCH(
@@ -549,7 +549,7 @@ class InovelliSwitchTypeEntity(ZCLEnumSelectEntity):
     _unique_id_suffix = "switch_type"
     _attribute_name = "switch_type"
     _enum = InovelliSwitchType
-    _attr_translation_key: str = "switch_type"
+    _attr_translation_key: str = "fan_switch_type"
 
 
 class InovelliFanSwitchType(types.enum1):
--- a/zha/application/platforms/select.py
+++ b/zha/application/platforms/select.py
@@ -63,6 +63,7 @@ class EnumSelectEntity(PlatformEntity):
     _attr_entity_category = EntityCategory.CONFIG
     _attribute_name: str
     _enum: type[Enum]
+    _translation_keys: dict[str, str]
 
     def __init__(
         self,
@@ -75,7 +76,10 @@ class EnumSelectEntity(PlatformEntity):
         """Init this select entity."""
         self._cluster_handler: ClusterHandler = cluster_handlers[0]
         self._attribute_name = self._enum.__name__
-        self._attr_options = [entry.name.replace("_", " ") for entry in self._enum]
+        self._translation_keys = {
+            entry.name.lower(): entry.name for entry in self._enum
+        }
+        self._attr_options = [entry.name.lower() for entry in self._enum]
         super().__init__(unique_id, cluster_handlers, endpoint, device, **kwargs)
 
     @functools.cached_property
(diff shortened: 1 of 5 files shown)


zha/application/platforms/select.py
```diff
@@ -75,7 +76,10 @@ def __init__(
         """Init this select entity."""
         self._cluster_handler: ClusterHandler = cluster_handlers[0]
         self._attribute_name = self._enum.__name__
-        self._attr_options = [entry.name.replace("_", " ") for entry in self._enum]
+        self._translation_keys = {
+            entry.name.lower(): entry.name for entry in self._enum
+        }
+        self._attr_options = [entry.name.lower() for entry in self._enum]
```
puddly, line 82: Will this be a breaking change for entities that do not yet implement translation keys?

Caius-Bonus: In homeassistant/core, I have written translation keys for all enums. But unfortunately, regardless of this specific code fragment, the internal state will change from the current form (titles with the first letter being a capital) to snake_case without any capitals. Automations that depend on this will seize to function.

puddly: Hmm. If it's going to be a breaking change we should get these strings into the exact state we want them (with no compatibility shims) and then prominently mark the breaking change as such.

Caius-Bonus: Currently, the only thing we do is .lower() as most Enums have capitals in them. Do you see that as a compatibility shim? Changing that requires changes in at least: zigpy/zha, zigpy/zigpy and zigpy/zha-device-handlers. There might be more users of zigpy/zigpy that rely on the names in the Enums.

Caius-Bonus: Secondly, if we want to create final names, we probably should change the CamelCase Enums to snake_case (CamelCase looks pretty bad in lowercase), which aren't many, but they do exist in zigpy/zha and zigpy/zha-device-handlers. 
Thirdly, maybe this is also a nice opportunity to move the Enums still present in zigpy/zha to zigpy/zha-device-handlers, while changing the naming where appropriate.

puddly: I've spoken to a few Core members and they said explained that unfortunately there is no migration path, this must be a breaking change.

The good news is that I looked over the [affected entities](https://github.com/home-assistant/core/pull/109309/files) once more and truthfully, they don't seem like something someone is going to automate with, making this a breaking change that would affect very very few people.

@TheJulianJES @dmulcahey thoughts?

TheJulianJES: Yeah, seems good.

puddly: To make this a one-time breaking change, I wonder if we can add a unit test to validate all enums in quirks to make sure they conform to the expected format?

TheJulianJES: I mean, yeah, but isn't the expected format just anything (ASCII characters) and underscores?
Like, we just lower-case all enum values here. They can't contain spaces, so.. not sure what we really want to check.

puddly: `SomeValue` would become `somevalue` instead of `some_value`. I think in zigpy we use a mix of `UPPERCASE_WITH_UNDERSCORES`, `CamelCase`, and `Whatever_case_this_is`.

TheJulianJES: Ok, I see. They're all still valid at least 😄 
So, for HA state/translation keys, we probably always want `some_value`.

Which style do we want to use for the actual enums in general? Also `some_value`?
Or do we want to leave them as-is for now and just convert all types you mentioned to `some_value` in ZHA? Should be possible at least.. (We could change the actual enums to `some_value` (or similar) later, without breaking HA state again then.)

Caius-Bonus: When leaving the actual enums as-is, from "SomeValue" we would get "somevalue" using ``.lower()``. So when changing the actual enums later on to "some_value", we would still have a breaking change. Although we could also do something like: ``re.sub(r"_+", r"_", re.sub(r"([A-Z])", r"_\1", "SomeValue").strip("_").lower())``.

Edit: fix multiple underscores for occurrences of "Some_Value".

puddly: Here's all of the enum members defined in quirks and zigpy: https://gist.github.com/puddly/2190cc8450bbcb6200afd6b87e516252.

I think the function in there should work, I didn't see any mistakes for this mapping:

```python
{
    "FooBar": "foo_bar",
    "FOOBAR": "foobar",
    "FOO_BAR": "foo_bar",
    "foo_bar": "foo_bar",
    "FOOBar": "foo_bar"
}
```

TheJulianJES: Yeah, let's use this.

PR #87, opened 2024-07-16
Fix Starkvind fan capabilities and modes being incorrectly exposed

freundTech:
### Breaking change (for end-users in HA)
The IKEA Starkvind fan modes have been updated. Automations that used to set the fan to 10% speed to switch it to "auto mode" should now be updated to set the preset to "auto."

In auto mode, the actual fan speed is now properly displayed.

For more information, see the PR in the ZHA repo: https://github.com/zigpy/zha/pull/87

### Proposed change

Fixes https://github.com/home-assistant/core/issues/97440.
Previously, Starkvind exposed 10 speed settings and no modes, where 10% corresponded to auto mode and 20%-100% corresponded to fixed speeds.

This patch correctly exposes auto mode as a mode. It also adds support for showing the actual fan speed while auto mode is enabled.

Starkvind supports 9 fan speeds. Because 9 doesn't neatly fit into 100% I cheated a bit and divided the 100% into 10% increments, where trying to set the fan to 10% sets it to 20% instead. I believe that this gives the overall better user experience compared to having 11.11% increments. This also keeps the previous fixed speeds in HA, instead of changing them all.
The 5 speed modes present on the physical interface of the device correspond to HA speed settings 20%, 40%, 60% and 100%.

### Additional information

~~This depends on https://github.com/home-assistant/core/pull/122033 to work.~~ That PR is no longer needed.
Previous PR: https://github.com/home-assistant/core/pull/114854


puddly:
You can fix most of the CI errors by running pre-commit:

```bash
pip install pre-commit
pre-commit install
pre-commit run --all-files
```

Ignoring the fan speed attribute values, just so that I understand correctly, the fan supports `auto` and then five unnamed speeds that you've mapped to 20-100%?

freundTech:
> You can fix most of the CI errors by running pre-commit

Done. Both pytest and pre-commit should now pass.

> the fan supports auto and then five unnamed speeds that you've mapped to 20-100%?

The fan supports 9 speeds, of which only 5 are exposed on the physical interface. The other four speeds are halfway between those speeds and used in auto mode, as well as exposed in the official app.

zha/application/platforms/fan/__init__.py
```diff
@@ -375,28 +373,15 @@ def async_update(self, _: Any = None) -> None:
...
-
-
 @MULTI_MATCH(
     cluster_handler_names="ikea_airpurifier",
     models={"STARKVIND Air purifier", "STARKVIND Air purifier table"},
 )
 class IkeaFan(Fan):
     """Representation of an Ikea fan."""
 
+    _attr_supported_features = FanEntityFeature.SET_SPEED | FanEntityFeature.PRESET_MODE
```
dmulcahey: not used?

zha/application/platforms/fan/__init__.py
```diff
@@ -375,28 +373,15 @@ def async_update(self, _: Any = None) -> None:
         self.maybe_emit_state_changed_event()
 
 
-IKEA_SPEED_RANGE = (1, 10)  # off is not included
-IKEA_PRESET_MODES_TO_NAME = {
```
dmulcahey, line 378: Do these really not exist? I don't have the device so I can't check for myself.

zha/application/platforms/fan/__init__.py
```diff
@@ -415,23 +400,66 @@ def __init__(
...
+        if preset_mode is not None:
+            await self.async_set_preset_mode(preset_mode)
+        elif speed is not None:
+            await self.async_set_percentage(self.speed_to_percentage(speed))
+        elif percentage is not None:
+            await self.async_set_percentage(percentage)
+        else:
+            await self.async_set_preset_mode("auto")
+
+    async def async_set_percentage(self, percentage: int) -> None:
```
dmulcahey, line 449: why are % and mode mixed in here?

zha/application/platforms/fan/__init__.py
freundTech: I think so. Yes.
Left over from adapting from my pre-zha-split version. Will remove.

zha/application/platforms/fan/__init__.py
freundTech: According to https://developers.home-assistant.io/docs/core/entity/fan 

> Preset modes should not include named (manual) speed settings as these should be represented as percentages.

These modes are basically just aliases for speeds.
The only real mode is auto.

zha/application/platforms/fan/__init__.py
freundTech: The starkvind zigbee interface is a bit weird. It has two attrubites: `fan_mode` and `fan_speed`, which are pretty similar.

`fan_mode` is r/w and controls both the mode and speed while `fan_speed` is read-only and displays the speed. Most of the time the values of `fan_mode` and `fan_speed` are identical, except for when `fan_mode` is set to auto. In that case `fan_speed` still reports the actual speed that the auto mode has set.

freundTech: Also see this PR for a longer explanation: https://github.com/zigpy/zha-device-handlers/pull/3088

zha/application/platforms/fan/__init__.py
freundTech: Done

freundTech:
Improved test coverage, cleaned up some code and addressed one of the review comments.

zha/application/platforms/fan/__init__.py
```diff
@@ -107,7 +107,7 @@ def speed_count(self) -> int:
         return int_states_in_range(self.speed_range)
 
     @functools.cached_property
-    def supported_features(self) -> int:
+    def supported_features(self) -> FanEntityFeature:
```
freundTech: The HA side uses `value in supported_features` to check if a feature is supported. The `in` operator is only valid on `IntFlag` Enums, not on `int` itself. This is already typed like this for cover, climate, siren and other entity types.

freundTech:
I rebased on top of dev to fix the merge conflicts caused by #100.

Also the HA part of this PR isnt needed anymore, as it has been addressed by https://github.com/home-assistant/core/pull/122813.

freundTech:
Rebased again because of the conflict caused by #125

freundTech:
Do you think it will be possible to get this into HA 2024.9 now that the zha rework has landed in 2024.8?

zha/application/platforms/fan/__init__.py
```diff
@@ -421,20 +413,54 @@ def __init__(
...
+        return 1, 10
 
     @property
-    def default_on_percentage(self) -> int:
-        """Return the default on percentage."""
-        return int(
-            (100 / self.speed_count) * self.preset_name_to_mode[PRESET_MODE_AUTO]
+    def percentage(self) -> int | None:
+        """Return the current speed percentage."""
+        self.debug(f"fan_speed is {self._fan_cluster_handler.fan_speed}")
```
TheJulianJES: I think we can remove this debug line. Seems a bit spammy, as this seems to be called often, for some reason. (Not related to this PR though)
We also already get multiple debug lines elsewhere when the `fan_speed` attribute changes.
```suggestion
```

zha/application/platforms/fan/__init__.py
```diff
@@ -421,20 +413,54 @@ def __init__(
...
+        # Starkvind turns on in auto mode by default.
+        if speed is None and percentage is None and preset_mode is None:
+            await self.async_set_preset_mode("auto")
+        else:
+            await super().async_turn_on(speed, percentage, preset_mode)
+
+    async def async_set_percentage(self, percentage: int) -> None:
+        """Set the speed percentage of the fan."""
+
+        self.debug(f"Set percentage {percentage}")
```
TheJulianJES: Eh, I think we can remove this too. We probably don't need to log what this method is called with every time.
```suggestion
```

zha/zigbee/cluster_handlers/manufacturerspecific.py
```diff
@@ -409,6 +409,11 @@ def fan_mode(self) -> int | None:
         """Return current fan mode."""
         return self.cluster.get("fan_mode")
 
+    @property
+    def fan_speed(self) -> int | None:
+        """Return current fan mode."""
```
TheJulianJES:
```suggestion
        """Return current fan speed."""
```

TheJulianJES:
Testing the PR right now. The behavior and code of the the PR seems good to me. 

> where trying to set the fan to 10% sets it to 20% instead. I believe that this gives the overall better user experience compared to having 11.11% increments

This behavior is slightly weird, I could see a bug report coming along the lines of "I can't set my STARKVIND to 10% anymore". But on the other hand, I also don't like 11.11% increments.. 
Changing the increments might also be more of a breaking change than not doing that. So, let's keep it like this for now.
Eh, and 10% meant "auto" before, so, *should* be fine.

Still, for anyone using the Starkvind in automations, this PR could be a breaking change, as setting it to 10% no longer sets it to auto mode. Maybe, this should be mentioned in the "breaking changes" section when ZHA is bumped in HA.
Automations just have to be updated to the proper "auto" fan mode now.

Lastly, I wonder if we can/should add a translation key for the "auto" fan mode on the HA side after this:
image

TheJulianJES:
I've added a "breaking change" text in the PR description. This should be copy-pasted by whoever bumps the lib in HA.
Also applied the small suggestions made above.

TheJulianJES (approved):
LGTM. Thanks!

PR #99, opened 2024-07-26
Add support for EFFECT_OFF

### LINKED ISSUE #98: Add `EFFECT_OFF` for Home Assistant
Home Assistant integrations can implement `EFFECT_OFF` instead of a "null" effect for the effect state now.
ZHA should implement this.
Reference: [https://developers.home-assistant.io/blog/2024/01/23/light-color-mode-changes/](https://developers.home-assistant.io/blog/2024/01/23/light-color-mode-changes/#a-special-effect-effect_off-which-means-no-effect--turn-off-effect-has-been-added)

This will also allows us to streamline/improve the "members assume group state" code here slightly:  https://github.com/zigpy/zha/blob/3a4ac1ef9e7b5334c3dacf95d66e65edc8f117cc/zha/application/platforms/light/__init__.py#L1085-L1086

dmulcahey:
Fixes: #98

### DIFF
--- a/tests/test_light.py
+++ b/tests/test_light.py
@@ -2029,7 +2029,7 @@ async def test_light_state_restoration(
         color_temp=None,
         xy_color=None,
         color_mode=None,
-        effect=None,  # Effect is the only `None` value actually restored
+        effect=None,
     )
 
     assert entity.state["on"] is True
@@ -2037,4 +2037,4 @@ async def test_light_state_restoration(
     assert entity.state["color_temp"] == 500
     assert entity.state["xy_color"] == (1, 2)
     assert entity.state["color_mode"] == ColorMode.XY
-    assert entity.state["effect"] is None
+    assert entity.state["effect"] == "colorloop"
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -906,6 +906,7 @@ class Light(PlatformEntity, BaseLight):
             effect = update_params.get(ATTR_EFFECT)
 
             supported_modes = self._supported_color_modes
+            effect_list = self._effect_list
 
             # unset "off brightness" and "off with transition"
             # if group turned on this light
@@ -944,7 +945,7 @@ class Light(PlatformEntity, BaseLight):
                 self._color_temp = color_temp
             if xy_color is not None and ColorMode.XY in supported_modes:
                 self._xy_color = xy_color
-            if effect is not None and self._effect_list and effect in self._effect_list:
+            if effect is not None and effect_list and effect in effect_list:
                 self._effect = effect
 
             self.maybe_emit_state_changed_event()
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -53,6 +53,7 @@ from zha.application.platforms.light.const import (
     DEFAULT_MIN_TRANSITION_MANUFACTURERS,
     DEFAULT_ON_OFF_TRANSITION,
     EFFECT_COLORLOOP,
+    EFFECT_NONE,
     FLASH_EFFECTS,
     SUPPORT_GROUP_LIGHT,
     ColorMode,
@@ -123,7 +124,7 @@ class BaseLight(BaseEntity, ABC):
         self._off_with_transition: bool = False
         self._off_brightness: int | None = None
         self._effect_list: list[str] | None = None
-        self._effect: str | None = None
+        self._effect: str = EFFECT_NONE
         self._supported_color_modes: set[ColorMode] = set()
         self._external_supported_color_modes: set[ColorMode] = set()
         self._zha_config_transition: int = self._DEFAULT_MIN_TRANSITION_TIME
@@ -175,7 +176,7 @@ class BaseLight(BaseEntity, ABC):
         return self._effect_list
 
     @property
-    def effect(self) -> str | None:
+    def effect(self) -> str:
         """Return the current effect."""
         return self._effect
 
@@ -462,7 +463,7 @@ class BaseLight(BaseEntity, ABC):
                 start_hue=0,
             )
             t_log["color_loop_set"] = result
-            self._effect = None
+            self._effect = EFFECT_NONE
 
         if flash is not None:
             result = await self._identify_cluster_handler.trigger_effect(
@@ -664,7 +665,7 @@ class Light(PlatformEntity, BaseLight):
         if self._color_cluster_handler:
             self._min_mireds: int = self._color_cluster_handler.min_mireds
             self._max_mireds: int = self._color_cluster_handler.max_mireds
-        effect_list = []
+        effect_list = [EFFECT_NONE]
 
         light_options = device.gateway.config.config.light_options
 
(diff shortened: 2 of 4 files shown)


dmulcahey:
@puddly @TheJulianJES do we want to do this color mode part: https://developers.home-assistant.io/blog/2024/01/23/light-color-mode-changes/#requirements-on-color_mode-are-less-strict-when-a-light-is-rendering-an-effect

TheJulianJES:
Can't comment on unchanged lines, but the assume option for effects should be changed to be like the others.
So, adding the `is not None` check. Line 1383 and following currently.
image

---

> do we want to do this color mode part

We could do that, yeah. (would prefer a separate PR though)
IIRC, changing the color during a colorloop effect doesn't impact the bulb, so we could set the color modes to on_off and brightness during that. Brightness changes still affect it, I think.

(On a semi-related note, I'd also really like to support Hue effects in the future. Will require changes to both quirks and zha.)

zha/application/platforms/light/const.py
```diff
@@ -91,6 +91,7 @@ class ColorMode(StrEnum):
 EFFECT_COLORLOOP: Final[str] = "colorloop"
 EFFECT_RANDOM: Final[str] = "random"
 EFFECT_WHITE: Final[str] = "white"
+EFFECT_NONE: Final[str] = "None"
```
TheJulianJES: Shouldn't this be `EFFECT_OFF = "off"` (not `EFFECT_NONE`)?
At least, that's what HA uses: [home-assistant/core/blob/homeassistant/components/light/__init__.py#L228-L233](https://github.com/home-assistant/core/blob/7820bcf2186c9e9d815c91c724e5187809b64d4b/homeassistant/components/light/__init__.py#L228-L233)

dmulcahey: Yep not sure wth I was looking at 😂. Good catch

zha/application/platforms/light/__init__.py
```diff
@@ -1044,8 +1045,8 @@ def _assume_group_state(self, update_params) -> None:
             if hs_color is not None and ColorMode.HS in supported_modes:
                 self._hs_color = hs_color
             # the effect is always deactivated in async_turn_on if not provided
-            if effect is None:
-                self._effect = None
+            if effect is None or effect == EFFECT_OFF:
+                self._effect = EFFECT_OFF
             elif self._effect_list and effect in self._effect_list:
                 self._effect = effect
```
TheJulianJES: I believe this still needs some changes, though I haven't tested the PR yet.
This might be what we actually want here:
```suggestion
            if effect is not None and effect in self._effect_list:
                self._effect = effect
```
(It's `None` if the `effect` wasn't touched on the group. In that case, we do not want to modify the member lights. If it's not None and in the effect list, the member will have applied that effect, so we want to set that state too. This will include `EFFECT_OFF`.)

EDIT: `_effect_list` can be `None` though, so maybe:
```suggestion
            if effect is not None and self._effect_list and effect in self._effect_list:
                self._effect = effect
```

zha/application/platforms/light/__init__.py
```diff
@@ -1062,7 +1063,7 @@ def restore_external_state_attributes(
         xy_color: tuple[float, float] | None,
         hs_color: tuple[float, float] | None,
         color_mode: ColorMode | None,
-        effect: str | None,
+        effect: str,
```
TheJulianJES: `restore_external_state_attributes` can still get `None` from HA, right? E.g. if the light is off or if the `core.restore_state` JSON file is deleted from HA(?)

I think we should revert this change and make sure that `effect is not None` before restoring it below (currently line 1087).

---

"Why should we not restore `None` effect? anymore"
Well, it'll now behave similar to all other state attributes. `None` means "no data" instead of "no effect" now. So, we should follow the same behavior as we do for the other state attributes.
For those, we do not restore `None` because some state attributes may already be populated from the zigpy cache, though it may be slightly outdated for non-attribute reporting lights.

zha/application/platforms/light/__init__.py
```diff
@@ -1397,7 +1398,7 @@ def restore_external_state_attributes(
...
```
TheJulianJES: Same here, this should be reverted IMO.
```suggestion
        effect: str | None,
```

TheJulianJES:
I've rebased the PR to fix the conflicts and addressed the comments I just made above. There's some more context in those comments on why the changes are needed IMO.

tests/test_light.py
```diff
@@ -2084,12 +2084,12 @@ async def test_light_state_restoration(
         xy_color=None,
         hs_color=None,
         color_mode=None,
-        effect=None,  # Effect is the only `None` value actually restored
```
TheJulianJES, line 2032: The workaround of restoring `effect` even if `None` is no longer needed/present, so the tests are adjusted.

zha/application/platforms/light/__init__.py
```diff
@@ -710,7 +711,7 @@ def __init__(
         if self._color_cluster_handler:
             self._min_mireds: int = self._color_cluster_handler.min_mireds
             self._max_mireds: int = self._color_cluster_handler.max_mireds
-        effect_list = []
+        effect_list = [EFFECT_OFF]
```
TheJulianJES, line 668: I believe changing this here might cause the effect option to always appear in HA now, even if the light does not actually support effects (other than off). (Haven't tested though.)

If the light does not support any effects, we do not want to expose `EFFECT_OFF`, correct? (Or do we? Does it matter? Does HA remove not show a single "off" effect in that case maybe?)
We should have a look at this before merging.

Should be addressed with https://github.com/zigpy/zha/pull/99/commits/aa870ef62f908aef14cb1081dbc76fa71a361cd8 (again, not tested, EDIT: reverted for now, see below. Might not matter.)

TheJulianJES: Should be addressed with https://github.com/zigpy/zha/pull/99/commits/aa870ef62f908aef14cb1081dbc76fa71a361cd8
(again, not tested)

TheJulianJES: Ok, I'm not actually sure that is needed, as `_supported_features` will not contain `EFFECT` already.
We may be able to revert that commit. It would also still return `EFFECT_OFF` in `effect`, even though that's not in the supported effects lists with that commit.
Without checking, it's not clear to me what behavior HA expects/wants for the `effect` and `effect_list` property when the "effect" feature isn't supported.

Looking at [HA light code](https://github.com/home-assistant/core/blob/dev/homeassistant/components/light/__init__.py), we may be able to safely revert https://github.com/zigpy/zha/commit/aa870ef62f908aef14cb1081dbc76fa71a361cd8 (?) Both options should really be tested..
EDIT: Reverted for now.

TheJulianJES: Ok, it works fine with that commit reverted.

TheJulianJES:
I think this PR should be ready now. The missing coverage is on existing lines, something to improve in the future.

---

Also discovered an existing (long-standing) issue when turning off effects via changing the color, but it's unrelated to this PR. Filed https://github.com/zigpy/zha/issues/210. Also something to be fixed in the future. The effect code on/off should be cleaned up a bit with it. Looks like nobody reported this before.

PR #108, opened 2024-07-29
Don't hide handler exceptions behind debug

elupus:
Don't swallow exceptions during configure of clusters in debug logs.

Catching generic Exceptions will hide programming errors. We could hide specific exceptions if we know their cause, other exceptions should be fixed.

### DIFF
--- a/tests/test_cluster_handlers.py
+++ b/tests/test_cluster_handlers.py
@@ -847,8 +847,8 @@ async def test_ep_cluster_handlers_configure(cluster_handler) -> None:
         assert ch.async_configure.call_count == 1
         assert ch.async_configure.await_count == 1
 
-    assert ch_3.debug.call_count == 2
-    assert ch_5.debug.call_count == 2
+    assert ch_3.warning.call_count == 2
+    assert ch_5.warning.call_count == 2
 
 
 async def test_poll_control_configure(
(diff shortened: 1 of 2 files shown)


puddly:
I would suggest `DeliveryError` and `asyncio.TimeoutError` be hidden. Otherwise, logs get very spammy.

elupus:
DeliveyError is a real error. We failed at doing something. Same with timeout. Hiding these type of things is why zha is so flakey at times.

puddly:
I don't think that results in anything actionable, though. What does a user do if they get a delivery error during startup? This is where `async_initialize` is called. All of this is device state polling and nothing breaks if the (likely identical) state of an entity isn't read at startup.

puddly:
That being said, more logging during device *joining* (and increased retries) would be good. We actually swapped it from a warning to a debug statement a few months ago because of log spam.

elupus:
This is also used during the configure step which can be triggered by user on a button. Manual configure should be verbose to indicate what worked and what failed.

It should be actionable to create a bug report on the issue.

I do agree that during normal startup: IF we have cached values, we could ignore some errors like timeouts. But we can only do it if we have cached values and user have not requested a reconfigure.

puddly:
> It should be actionable to create a bug report on the issue.

Definitely not. Delivery errors are unavoidable. Some devices are aggressive with sleeping after joining ad just are hard to reach. Some devices are hard to reach because the network has poor connectivity and the delivery error is genuine.

> I do agree that during normal startup: IF we have cached values, we could ignore some errors like timeouts.

It would be best to isolate this change then to just those scenarios. Otherwise, such verbose and user-visible logging creates a support burden.

elupus:
Converted to draft until we have some consensus. Also trying out https://github.com/zigpy/zha/pull/120 which might be a better alternative (can't decide yet).

PR #117, opened 2024-07-31
Remove unused instance mock code

elupus:
This code is no longer used and is very slow in general.


puddly (approved):
Thanks!

I think this was carried over from ZHA. We ran into a few bugs due to excessive mocking and I think this was added to remove bugs resulting from invalid function and property use.

PR #119, opened 2024-07-31
Make sure gateway fixture does not reload quirk on each run

elupus:
The code in quirks setup would still run since it does not try to protect for multiple invocations internally.

I can't fully patch out the executor and keep test coverage unless we move this quirks loading out of the gateway object.


puddly:
This speeds tests up by 2 seconds for me (5%). I think the executor impact is pretty negligible, last I remember it adds about 0.3ms per invocation.

puddly (approved):

PR #120, opened 2024-07-31
RFC: Let cluster config/init bubble up exception groups

elupus:
Let cluster config/init bubble up exception groups
 - Only skip errors when re-joining

This shows that we have several faults we currently don't detect properly.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -25,7 +25,7 @@ import zigpy.profiles
 from zigpy.quirks import get_device
 import zigpy.types
 from zigpy.zcl.clusters.general import Basic, Groups
-from zigpy.zcl.foundation import Status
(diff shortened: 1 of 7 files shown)


elupus:
So i've gotten the tests in line i think. There where several cases where our tests where throwing unexpected exceptions due to unexpected responses on data. This exposes those errors and adds some solutions to them.

What do you guys think?

elupus:
Ps. Some of the fixes should be broken out into separate changes perhaps.

puddly:
I think some of the tests are also failing because we are (slowly) migrating away from using lists (and lists-of-lists) to represent command responses to `t.Struct` subclasses that concurrently support a tuple interface (`rsp[0]`) in addition to normal attribute lookup (`rsp.status`). I think the unit tests right now have a lot of places that return lists-of-lists (`return [[]]`) that bubble up to code that uses attributes.

PR #121, opened 2024-08-01
Split quirks loading from network and app initialization

dmulcahey:
This PR moves the loading of quirks so that we can split the quirk loading from the initialization of the network and the application. There are a couple HA users that hit an issue in the beta where device triggers from quirks aren't available early enough and if the integration fails to set up on the first shot (flaky stick, poor nwk conditions etc.) it prevents automations from finding triggers for ZHA devices:

```
Logger: homeassistant.components.automation.drukknop_slaapkamer_noor
Source: components/automation/__init__.py:842
integration: Automation (documentation, issues)
First occurred: 02:53:39 (1 occurrences)
Last logged: 02:53:39

Got error 'Unable to find trigger ('remote_button_short_press', 'remote_button_short_press')' when setting up triggers for Drukknop slaapkamer Noor
```

There is a corresponding HA PR: https://github.com/home-assistant/core/pull/123027

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -323,6 +323,7 @@ class TestGateway:
     async def __aenter__(self) -> Gateway:
         """Start the ZHA gateway."""
         self.zha_gateway = await Gateway.async_from_config(self.zha_data)
+        await self.zha_gateway.async_initialize()
         await self.zha_gateway.async_block_till_done()
         await self.zha_gateway.async_initialize_devices_and_entities()
         INSTANCES.append(self.zha_gateway)
--- a/zha/application/gateway.py
+++ b/zha/application/gateway.py
@@ -215,7 +215,13 @@ class Gateway(AsyncUtilMixin, EventBase):
     async def async_from_config(cls, config: ZHAData) -> Self:
         """Create an instance of a gateway from config objects."""
         instance = cls(config)
-        await instance.async_initialize()
+
+        if config.config.quirks_configuration.enabled:
+            await instance.async_add_executor_job(
+                setup_quirks,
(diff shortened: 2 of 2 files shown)


TheJulianJES (approved):

elupus (commented):
Why is this part pf the gateway code at all? Quirks are global.

puddly:
@elupus Custom quirks are reloaded from disk. This is also sort of a remnant of ZHA in Core.

elupus:
Sure, but it does not need to be done inside the gateway class. Could be done outside of that by the caller constructor of the gateway object.

elupus:
What i mean is, if we are changing it, it is better to remove it from gateway and make it a caller responsibility. Makes testing easier.

dmulcahey:
I don't have a strong opinion either way. My initial view was that the gateway should essentially function as a standalone application. Take HA out of the picture and you should be able to take the gateway and just use it. I already have plans to eventually have an add-on and we are also used by another automation platform (they use all the libs and hand rolled what is essentially this lib now... hopefully the convert too). Initial thinking was anything that would be common across all would be done in 1 spot 🤷🏻‍♂️

puddly:
> Could be done outside of that by the caller constructor of the gateway object.

Quirk loading is also somewhat tied to the lifecycle of `Gateway`: quirks should always be loaded *before* `Gateway` is constructed and preferably re-loaded before `Gateway` is re-initialized a second time. Quirks loading is pretty much a part of startup and I think it should be hidden from upstream application so I think it would be preferable to keep it a part of `Gateway`.

PR #128, opened 2024-08-05
Support new zigpy OTA update notification schema

puddly:
- Corresponding zigpy PR: https://github.com/zigpy/zigpy/pull/1397
 - Corresponding Core branch: https://github.com/home-assistant/core/compare/dev...puddly:puddly/zha-update-versions

This PR adds support for passing more granular version metadata for OTA updates from zigpy, allowing us to show all current and past versions for an OTA update. This will allow (once Home Assistant supports it) downgrading firmwares from the UI, selecting images from multiple sources, and so on.

This changes OTA update entity semantics a little: the OTA entity is now "unknown" if we do not have an update for the device, as opposed to turning in "on".


zha/application/platforms/update.py
```diff
@@ -231,16 +238,41 @@ def handle_cluster_handler_attribute_updated(
...
+            self._attr_release_summary = (
+                latest_firmware.metadata.changelog
+                if latest_firmware.metadata.changelog
+                else None
+            )
+            self._attr_release_notes = (
+                latest_firmware.metadata.release_notes
+                if latest_firmware.metadata.release_notes
+                else None
+            )
```
TheJulianJES: We might be able to do this:
```suggestion
            self._attr_release_summary = latest_firmware.metadata.changelog or None
            self._attr_release_notes = latest_firmware.metadata.release_notes or None
```

PR #130, opened 2024-08-05
Clean up cluster handler attribute_updated handling

### LINKED ISSUE #126: Improve `attribute_updated` in cluster handlers
The `attribute_updated` stuff in the cluster handlers should be improved. There's some duplicate code that can be avoided and it's one of the last places where the old `_value_attribute` stuff is still used.
It's also a bit confusing if only one attribute change actually emits an "attribute updated" event on a cluster. Maybe we should always emit the events and just check for the attribute in the places where we're subscribing? (the config switches do this already) It shouldn't cause any performance issues really..

dmulcahey:
Fixes: #126

### DIFF
--- a/zha/zigbee/cluster_handlers/general.py
+++ b/zha/zigbee/cluster_handlers/general.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 from collections.abc import Coroutine
 from dataclasses import dataclass
+from datetime import datetime
 from typing import TYPE_CHECKING, Any, Final
 
 from zhaquirks.quirk_ids import TUYA_PLUG_ONOFF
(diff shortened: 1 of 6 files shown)


zha/zigbee/cluster_handlers/general.py
```diff
@@ -367,6 +365,8 @@ def attribute_updated(self, attrid: int, value: Any, _: Any) -> None:
         self.debug("received attribute: %s update with value: %s", attrid, value)
         if attrid == self.CURRENT_LEVEL:
             self.dispatch_level_change(SIGNAL_SET_LEVEL, value)
+        else:
+            super().attribute_updated(attrid, value, _)
```
puddly: If we're using the attribute update time param (`_`), we should give it a name

PR #144, opened 2024-08-06
Avoid creating binary sensor for unsupported attributes

elupus:
Some of these are virtual attributes updated but "toggle" events the the OnOffClusterHandler, but not really supported by the device itself.

### DIFF
--- a/zha/application/platforms/binary_sensor/__init__.py
+++ b/zha/application/platforms/binary_sensor/__init__.py
@@ -5,12 +5,13 @@ from __future__ import annotations
 from dataclasses import dataclass
 import functools
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Self, Any
 
 from zhaquirks.quirk_ids import DANFOSS_ALLY_THERMOSTAT
 from zigpy.quirks.v2 import BinarySensorMetadata
 
 from zha.application import Platform
+from zha.application.const import ENTITY_METADATA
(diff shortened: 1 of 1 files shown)


zha/application/platforms/binary_sensor/__init__.py
```diff
@@ -61,6 +62,33 @@ class BinarySensor(PlatformEntity):
...
+        endpoint: Endpoint,
+        device: Device,
+        **kwargs: Any,
+    ) -> Self | None:
+        """Entity Factory.
+
+        Return entity if it is a supported configuration, otherwise return None
+        """
+        cluster = cluster_handlers[0].cluster
+        if ENTITY_METADATA not in kwargs and (
```
elupus, line 79: No idea why this is here, it was in one of the other handler like this.

elupus:
Okey,. this is not going to fly. The whole stuff with handle_on_off_output_cluster_exception makes things weird here. We hook up the server cluster handler to the client  (out) cluster. 
How much will break if we drop that? 

As far as i understand there are some weird Tuya devices that make use of this strangeness as well as perhaps some LUMI stuff (https://github.com/zigpy/zha/blob/9fe83455574b8aa3f29cb5e722f3b312e6615f79/tests/zha_devices_list.py#L3088)

elupus:
I've opened https://github.com/zigpy/zha/issues/145 to try to figure out a better solution.

PR #148, opened 2024-08-08
Remove group translation keys

puddly:
Groups shouldn't use translation keys to begin with, as groups (even without names) will always have a string name.

Before:
image

After:
image

https://github.com/home-assistant/core/issues/123373


dmulcahey (approved):
Awesome.

PR #151, opened 2024-08-08
Pass the local timezone to `Gateway`

### LINKED ISSUE #150: Time in ZHA is UTC despite setting local time in HA
I display the local time with my Sinopé thermostats but since HA 2024.8.0 my thermostats are displaying UTC. Seem's to happen with the decoupling of ZHA from core.

https://github.com/zigpy/zha/blob/db2bf30f7399c54d0004f06302404b139b887abc/zha/application/platforms/climate/__init__.py#L536

puddly:
Fixes https://github.com/zigpy/zha/issues/150 (once the Core side is implemented).

The default behavior will be to use UTC. Maybe we can instead use the machine time zone?

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -12,3 +12,4 @@ pytest
 zigpy>=0.65.2
 ruff
 looptime
+freezegun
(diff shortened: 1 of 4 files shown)


puddly:
Core PR: https://github.com/home-assistant/core/compare/03ba8f61...puddly:core:puddly/zha-bump-20240808

TheJulianJES (approved):
ZHA and Core changes look good

PR #153, opened 2024-08-09
Port Sinope changes over from core/zha

ckm2k1:
Icon diff: 

```diff
diff --git a/homeassistant/components/zha/icons.json b/homeassistant/components/zha/icons.json
index 9b060e8105..bb4224f2bd 100644
--- a/homeassistant/components/zha/icons.json
+++ b/homeassistant/components/zha/icons.json
@@ -94,6 +94,12 @@
       },
       "keypad_lockout": {
         "default": "mdi:lock"
+      },
+      "off_led_color": {
+        "default": "mdi:palette-outline"
+      },
+      "on_led_color": {
+        "default": "mdi:palette"
       }
     },
     "sensor": {
@@ -158,6 +164,9 @@
       },
       "hooks_locked": {
         "default": "mdi:lock"
+      },
+      "double_up_full": {
+        "default": "mdi:gesture-double-tap"
       }
     }
   },
```

### DIFF
--- a/tests/zha_devices_list.py
+++ b/tests/zha_devices_list.py
@@ -7351,7 +7351,21 @@ DEVICES = [
         DEV_SIG_DEV_NO: 102,
         SIG_MANUFACTURER: "Sinope Technologies",
         SIG_MODEL: "SW2500ZB",
-        SIG_NODE_DESC: b"\x02@\x80|\x11RR\x00\x00\x00R\x00\x00",
+        SIG_NODE_DESC: {
+            "logical_type": 1,
+            "complex_descriptor_available": 0,
+            "user_descriptor_available": 1,
+            "reserved": 0,
+            "aps_flags": 0,
+            "frequency_band": 8,
+            "mac_capability_flags": 142,
+            "manufacturer_code": 4508,
+            "maximum_buffer_size": 71,
+            "maximum_incoming_transfer_size": 43,
+            "server_mask": 10752,
+            "maximum_outgoing_transfer_size": 43,
+            "descriptor_capability_field": 0,
+        },
         SIG_ENDPOINTS: {
             1: {
                 SIG_EP_TYPE: 257,
--- a/tests/zha_devices_list.py
+++ b/tests/zha_devices_list.py
@@ -7347,4 +7347,94 @@ DEVICES = [
             },
         },
     },
+    {
+        DEV_SIG_DEV_NO: 102,
+        SIG_MANUFACTURER: "Sinope Technologies",
+        SIG_MODEL: "SW2500ZB",
+        SIG_NODE_DESC: b"\x02@\x80|\x11RR\x00\x00\x00R\x00\x00",
+        SIG_ENDPOINTS: {
+            1: {
+                SIG_EP_TYPE: 257,
+                DEV_SIG_EP_ID: 1,
+                SIG_EP_INPUT: [0, 2, 3, 4, 5, 6, 8, 1794, 2820, 2821, 65281],
+                SIG_EP_OUTPUT: [3, 4, 10, 25],
+                SIG_EP_PROFILE: 260,
+            }
+        },
+        DEV_SIG_EVT_CLUSTER_HANDLERS: ["1:0x0019"],
+        DEV_SIG_ENT_MAP: {
+            ("button", "00:11:22:33:44:55:66:77-1-3"): {
+                DEV_SIG_CLUSTER_HANDLERS: ["identify"],
+                DEV_SIG_ENT_MAP_CLASS: "IdentifyButton",
+                DEV_SIG_ENT_MAP_ID: "button.sinope_technologies_dm2500zb_identify",
+            },
+            ("light", "00:11:22:33:44:55:66:77-1"): {
+                DEV_SIG_CLUSTER_HANDLERS: ["level", "on_off"],
+                DEV_SIG_ENT_MAP_CLASS: "Light",
+                DEV_SIG_ENT_MAP_ID: "light.sinope_technologies_dm2500zb_light",
+            },
+            ("sensor", "00:11:22:33:44:55:66:77-1-2820"): {
+                DEV_SIG_CLUSTER_HANDLERS: ["electrical_measurement"],
+                DEV_SIG_ENT_MAP_CLASS: "PolledElectricalMeasurement",
+                DEV_SIG_ENT_MAP_ID: "sensor.sinope_technologies_dm2500zb_power",
+            },
+            ("sensor", "00:11:22:33:44:55:66:77-1-2820-apparent_power"): {
+                DEV_SIG_CLUSTER_HANDLERS: ["electrical_measurement"],
+                DEV_SIG_ENT_MAP_CLASS: "ElectricalMeasurementApparentPower",
+                DEV_SIG_ENT_MAP_ID: (
+                    "sensor.sinope_technologies_dm2500zb_apparent_power"
+                ),
+            },
+            ("sensor", "00:11:22:33:44:55:66:77-1-2820-rms_current"): {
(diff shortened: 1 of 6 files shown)


dmulcahey:
run `pre-commit install` locally and you can fix the lint issues before they get to the PR.

zha/application/platforms/select.py
```diff
@@ -820,3 +820,55 @@ class DanfossViewingDirection(ZCLEnumSelectEntity):
...
+    },
+)
+class SinopeLightLEDOnColorSelect(ZCLEnumSelectEntity):
+    """Representation of the marker LED color on-state color of Sinope light switches."""
+
+    _unique_id_suffix = "on_led_color"
+    _attribute_name = "on_led_color"
+    _attr_translation_key: str = "on_led_color"
+    _enum = SinopeLightLedColors
+    _attr_icon: str = "mdi:palette"
```
ckm2k1: @dmulcahey 

I can't seem to get HA to respect the `_attr_icon` attribute in the UI. It's showing the default entity icon instead. 
Any idea what I'm missing here? 

image

dmulcahey: Looks like a bug: <https://github.com/zigpy/zha/blob/f015eab0514dc33972d97535958fc78aeafce853/zha/application/platforms/__init__.py#L139>

dmulcahey: https://github.com/zigpy/zha/pull/171

ckm2k1: Thanks for the quick patch!

puddly: Core has moved to `icons.json` to specify entity icons. You should be able to add your entity's icon to the same file: <https://github.com/home-assistant/core/blob/dev/homeassistant/components/zha/icons.json>. If you update the PR with a diff, we can include it when the ZHA library is bumped.

ckm2k1: Something like this? 

```diff
diff --git a/homeassistant/components/zha/icons.json b/homeassistant/components/zha/icons.json
index 9b060e8105..bb4224f2bd 100644
--- a/homeassistant/components/zha/icons.json
+++ b/homeassistant/components/zha/icons.json
@@ -94,6 +94,12 @@
       },
       "keypad_lockout": {
         "default": "mdi:lock"
+      },
+      "off_led_color": {
+        "default": "mdi:palette-outline"
+      },
+      "on_led_color": {
+        "default": "mdi:palette"
       }
     },
     "sensor": {
@@ -158,6 +164,9 @@
       },
       "hooks_locked": {
         "default": "mdi:lock"
+      },
+      "double_up_full": {
+        "default": "mdi:gesture-double-tap"
       }
     }
   },
```

I'm assuming I should remove the `_attr_icon` from the relevant entities.

ckm2k1: Tested the change locally with the diff above, works great!

image

zha/zigbee/cluster_handlers/manufacturerspecific.py
```diff
@@ -496,3 +499,67 @@ class DanfossDiagnosticClusterHandler(DiagnosticClusterHandler):
...
+            "DM2550ZB-G2",
+        )
+
+        _LOGGER.debug(
+            "matching sinope device to cluster handler %s", cluster.endpoint.model
+        )
+
+        return (
+            cluster.endpoint.model in switches
+            or cluster.endpoint.device.model in switches
```
dmulcahey: this is redundant: https://github.com/zigpy/zigpy/blob/f1dddb45d6e7e0cc8a6fb4a0098a9b33149cf024/zigpy/endpoint.py#L331

zha/zigbee/cluster_handlers/manufacturerspecific.py
```diff
@@ -496,3 +499,64 @@ class DanfossDiagnosticClusterHandler(DiagnosticClusterHandler):
...
+
+        elif self.cluster.endpoint.model in ["SW2500ZB", "SW2500ZB-G2"]:
+            self.ZCL_INIT_ATTRS = shared_attr.copy()
+
+    REPORT_CONFIG = (
+        AttrReportConfig(
+            attr="action_report",
+            config=(
+                REPORT_CONFIG_MIN_INT_IMMEDIATE,
+                REPORT_CONFIG_MIN_INT_IMMEDIATE,
```
dmulcahey, line 536: Do we really want 0 for min and max?

dmulcahey:
Overall looks good. Just needs tests for the `SinopeManufacturerClusterHandler`.

zha/zigbee/cluster_handlers/manufacturerspecific.py
ckm2k1: I'm fairly new to the Zigbee spec so I probably got this wrong. 

My understanding is that it specifies how frequently the device should send updates about the `action_report` attribute, which for Sinope switches is how it communicates any button presses on the switch, especially non-single-press, like "long press" or "double tap".  

I figured 0 made sense because it'll be responsive when you press the switch. Do you think this should be a larger value on the Max side?

dmulcahey: the maximum interval is for sending a report even if the value hasn't changed. The way this is configured at the moment fits this case I believe:

"If this value is set to 0x0000, and the minimum reporting interval field does not equal 0xffff there SHALL be no periodic reporting, but change based reporting SHALL still be operational."

min and max are both 0 so it looks like it will be change based reporting (this may be fine?) 

image

ckm2k1: That was my reasoning too. I wrote `action_report` handling in the Sinope quirk to fire a ZHA event on change, which you wouldn't want to fire periodically, especially when hooked up to an automation.

tests/zha_devices_list.py
```diff
@@ -5919,4 +5919,94 @@
             },
         },
     },
+    {
+        DEV_SIG_DEV_NO: 102,
+        SIG_MANUFACTURER: "Sinope Technologies",
+        SIG_MODEL: "SW2500ZB",
+        SIG_NODE_DESC: b"\x02@\x80|\x11RR\x00\x00\x00R\x00\x00",
```
ckm2k1: Is there an easy way to pull node descriptions from a device? 
I copied the above from one of the IKEA devices, but I can't imagine that's correct.

puddly: Once https://github.com/zigpy/zha/pull/177 is merged, you can use the node descriptor dictionary from the "Signature" tab of the device management UI:

image

puddly:
Apparently you can close PRs via other PRs, same as issues 😄.

ckm2k1:
@dmulcahey @puddly
I don't see how to spin up the manufacturer specific cluster during tests. I've added a Sinope light switch in `/tests/zha_devices_list.py`, and all tests are passing, but the cluster handler isn't instantiated. 

Could you point me in the right direction?

dmulcahey:
> @dmulcahey @puddly I don't see how to spin up the manufacturer specific cluster during tests. I've added a Sinope light switch in `/tests/zha_devices_list.py`, and all tests are passing, but the cluster handler isn't instantiated.
> 
> Could you point me in the right direction?

this is because the manufacturer cluster with the attribute definitions doesn't exist in ZHA so when the base class attempts to process this logic: https://github.com/ckm2k1/zha/blob/80b80909dd49a094f2d8341772be40e3a1e70625/zha/zigbee/cluster_handlers/__init__.py#L206 it bombs out and logs a warning with a key error. We are discussing ways to make these failures much more apparent. For now you have 2 options: define the quirk fully in the test w/ the custom cluster and its attributes or define `_value_attribute` in the cluster handler so the lookup doesn't have to be processed.

dmulcahey (approved):

PR #154, opened 2024-08-09
Use a generic device name for coordinators without information

puddly:
Not sure if it's better to swap the model and manufacturer: `Generic EZSP by Zigbee Coordinator`. Or use other strings. Open to ideas!


dmulcahey (approved):
Change itself looks good. these strings are fine IMO.

puddly:
Home Assistant displays both concatenated in the frontend, without `by`:

image

But also with `by`:

image

puddly:
Yeah, I think @dmulcahey's suggestion looks better:

image
image

PR #155, opened 2024-08-09
Fix group entity API inconsistency with async_update

dmulcahey:
Fixes: https://github.com/home-assistant/core/issues/123472


TheJulianJES (approved):
So that’s why we had `update()` before for groups 😄

dmulcahey:
> So that’s why we had `update()` before for groups 😄

I knew I had a reason when I changed it... LOL

PR #159, opened 2024-08-10
Fix invalid uniqueness constraint on unique_id

### LINKED ISSUE #157: Valve device switch.id_on_off is not set in switch entity
Since last HA update 2024.8.0, Sinopé  valve devices lost there on/off switch. valve device are loaded by sinope quirks in switch.py.
We can't open or close the valve since the update.
There are only bynary_sensor and sensor that are created for the valve,
There should be a switch.valve_id_on_off created to be able to open and close the valve

dmulcahey:
Oh boy... where to start... 

fixes: https://github.com/zigpy/zha/issues/157
related: https://github.com/home-assistant/core/issues/123387

# Back story
An incorrect assumption was made when converting ZHA to a lib that the `unique_id` of an entity in HA was globally unique. This is not the case. `unique_id` is only unique per platform. This caused unique id clashes that can be seen as detailed in the linked issues. 

image

# Fix
To fix the issue the `platform_entities` dict is now keyed by a tuple: (platform, unique_id) and all areas that leverage the dict are updated to account for the change.

### DIFF
--- a/zha/zigbee/device.py
+++ b/zha/zigbee/device.py
@@ -29,7 +29,7 @@ from zigpy.zcl.foundation import (
 import zigpy.zdo.types as zdo_types
 from zigpy.zdo.types import RouteStatus, _NeighborEnums
 
-from zha.application import discovery
+from zha.application import Platform, discovery
 from zha.application.const import (
     ATTR_ARGS,
     ATTR_ATTRIBUTE,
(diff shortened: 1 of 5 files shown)


zha/zigbee/device.py
```diff
@@ -672,7 +672,7 @@ def extended_device_info(self) -> ExtendedDeviceInfo:
             **self.device_info.__dict__,
             active_coordinator=self.is_active_coordinator,
             entities={
-                unique_id: platform_entity.info_object
+                unique_id[1]: platform_entity.info_object
                 for unique_id, platform_entity in self.platform_entities.items()
```
puddly: Maybe we can use `for (platform, unique_id), platform_entity in self.platform_entities.items()` to avoid `unique_id[1]`?

dmulcahey: actually can't we just use platform_entity.unique_id?

puddly: Agreed, `{entity.unique_id: entity.info_object for entity in self.platform_entities.values()}` is much better!

dmulcahey:
```
           entities={
                platform_entity.unique_id: platform_entity.info_object
                for platform_entity in self.platform_entities.values()
            },
```

TheJulianJES (approved):

PR #164, opened 2024-08-12
Fix group entity availability handling

dmulcahey:
fixes: https://github.com/home-assistant/core/issues/123518#issuecomment-2284411430

### DIFF
--- a/tests/test_light.py
+++ b/tests/test_light.py
@@ -920,6 +920,36 @@ async def test_zha_group_light_entity(
     await zha_gateway.async_block_till_done()
     assert bool(entity.state["on"]) is True
 
+    assert entity.state["available"] is True
+
+    device_light_1.on_network = False
+    device_light_2.on_network = False
+    await asyncio.sleep(0.1)
+    await zha_gateway.async_block_till_done()
+
+    assert entity.state["available"] is False
+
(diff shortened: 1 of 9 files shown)


tests/test_light.py
```diff
@@ -920,6 +920,36 @@ async def test_zha_group_light_entity(
...
+
+    device_light_1.available = False
+    device_light_2.available = False
+    await asyncio.sleep(0.1)
+    await zha_gateway.async_block_till_done()
+
+    assert entity.state["available"] is False
+
+    device_light_1.available = True
+    device_light_2.available = True
```
TheJulianJES: Do we also want to test that the group entity is still available when only one of the members is available?
The test currently always sets all entities to either unavailable or available.
(Testing with half availability would make sure that our group `available` function works properly.)

We could just modify the availability of one light here, since we're basically testing the same for the second time (just after the members being unavailable previously).

dmulcahey: sure it's an additional 2 lines in each test... gimme a sec

TheJulianJES (approved):

dmulcahey:
gonna clean up the tests some more

dmulcahey:
done

PR #181, opened 2024-08-26
[RFC] - Leverage device diagnostics files for tests

dmulcahey:
This PR is a bit of an experiment currently. The changes aim to make testing easier and to clean up discovery tests too.

Goals:

1. create test devices from device diagnostic files with very little code:
    ```python
        zha_device = await zha_device_from_file(file_path)
    ```
    or if you want to do something to the zigpy_device before joining it to ZHA:
    ```python
        zigpy_device = zigpy_device_from_json(
            "tests/data/devices/lumi/lumi.sen_ill.mgl01.json"
        ) 
        zha_device = await device_joined(zigpy_device)
    ```
 
2. clean up discovery tests and remove the current unmaintainable device list from tests and hacks such as this: 

    https://github.com/zigpy/zha/blob/b13c1f82ae2ac068aac9bd6d1186d6371fa6e890/tests/test_discover.py#L74C1-L88C82
 
4. make discovery tests explode if there is a problem executing any logic in cluster handlers so errors aren't hidden

The PR creates 2 new fixtures: `zigpy_device_from_json` and `zha_device_from_file`. They allow for the creation of test devices from device diagnostic data. 

I am currently playing with adding additional data to the diagnostic output here: https://github.com/dmulcahey/zha-device-exporter/blob/master/custom_components/zha_device_exporter/__init__.py

This can be added as a custom component to HA and you can add: `zha_device_exporter:` to your configuration.yaml to test the changes. The component defines a single service that dumps all of the modified diagnostic data files to the folder `zha-device-export` in the HA config directory. 

The code itself is still messy and needs a bit of polish / deduplication.

**Future enhancements:**

Once we have binding tables and attribute reporting configurations stored in zigbee.db we can add that data to the diagnostics file as well. This will allow us to assert the bindings and report configurations for each cluster in the discovery tests as well.


puddly:
We'll iterate on the specific format for the JSON files in a future release, to make it compatible with Home Assistant's diagnostics. For now this is going to really help make sure we don't modify existing entities when making library changes.

PR #188, opened 2024-08-30
Implement smarter sensor polling

dmulcahey:
This PR changes the way that polling sensors work. `PollableSensor` has been updated to use the `GlobalUpdater` and a heuristic has been added to identify when the device is reporting attributes so that it can disable the polling since it shouldn't be needed. 

This accomplishes 3 things: 
1. devices that report will stop polling after the report threshold is reached (it would be good to be able to persist this across restarts)
2. polling of non reportable devices now is limited so that only 3 devices will poll at any given time (gather with limited concurrency)
3. less polling tasks get created (1 vs # of devices previously)

TODO: handle previously known device models so they avoid the work


zha/application/helpers.py
```diff
@@ -394,34 +402,49 @@ def stop(self):
             self._updater_task_handle = None
         _LOGGER.debug("global updater stopped")
 
-    def register_update_listener(self, listener: Callable):
+    def register_update_listener(self, listener: Callable | Callable[[], Awaitable]):
```
dmulcahey: not sure if the typing is right on this...

puddly:
```suggestion
    def register_update_listener(self, listener: Callable[[], Coroutine[None, None, None]]) -> None:
```

dmulcahey:
closing for now... will explore other avenues

PR #190, opened 2024-09-02
Implement device scanning

### LINKED ISSUE #115: Implement a full device scan
https://github.com/mdeweerd/zha-toolkit/blob/main/custom_components/zha_toolkit/scan_device.py

We should include this functionality in ZHA. We should schedule this post device join.

dmulcahey:
fixes: #115

This is VERY WIP at the moment. Just getting some ideas into code to see what things look like. Feel free to drop comments / ideas here as I work through this.

### DIFF
--- a/zha/zigbee/cluster_handlers/__init__.py
+++ b/zha/zigbee/cluster_handlers/__init__.py
@@ -11,12 +11,20 @@ import logging
 from typing import TYPE_CHECKING, Any, Final, ParamSpec, TypedDict
 
 import zigpy.exceptions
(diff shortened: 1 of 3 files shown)


dmulcahey:
Closing because this scanner is built on the removed `ClusterHandler` APIs and is not a usable implementation against current `dev`. It also discards the scan results and reads only the first page of the binding table.

The diagnostic use case in #115 would need a new implementation over current cluster APIs, with structured results and complete pagination. Rebasing this PR would not resolve those design gaps.

PR #197, opened 2024-09-07
Add AnalogInput sensor, use description attribute for fallback_name

### LINKED ISSUE #160: Query and use device values when creating entities
Many Zigbee clusters provide a Description attribute (0x001C). We should use this, if provided, for the entity name and id. In addition, the multistate cluster provides the State Text attribute (0x000E), as we add entities for this cluster, we should use the provided values for the friendly state.

prairiesnpr:
This ensures that the description attribute (0x001C), if present, is used for the fallback_name and sets the translation_key to None. The entity id and name will use the fallback_name and align with the devices value. This is mostly useful for devices using general clusters, if not set, the current functionality is maintained.

Due to testing requirements, this now also updates the cluster handlers for AnalogInput, MultistateInput, and BinaryInput.

Closes: https://github.com/zigpy/zha/issues/160

image

image

Multistate still requires https://github.com/zigpy/zigpy/pull/1456 to be fully implemented, since the LVLists aren't serialized, we create a default value for now.
image

AnalogOutput is putting the description in the name twice.
image

This is due to this logic in core, should we add similar logic to core for the rest of the Entities and remove it from ZHA, keep the logic in ZHA, or have it mixed?

https://github.com/home-assistant/core/blob/f1e4229b23ef7837bb4e46877f55d6533f8d4607/homeassistant/components/zha/number.py#L52C9-L52C58


puddly:
I wonder: would it be feasible to instead set the translation key to the ZCL `description`? This could allow the description to be localized.

prairiesnpr:
> I wonder: would it be feasible to instead set the translation key to the ZCL `description`? This could allow the description to be localized.

That was the approach I started with, unfortunately, when the translation key wasn't found, it would return None for the name and id. I understood it should then fallback, but that's not the result I observed in testing.

zha/application/platforms/sensor/const.py
```diff
@@ -390,3 +390,36 @@ class SensorDeviceClass(enum.StrEnum):
     SensorDeviceClass.ENUM,
     SensorDeviceClass.TIMESTAMP,
 }
+
+
+class AnalogInputStateClass(enum.Enum):
```
prairiesnpr: The ZCL specifies units for each value of ApplicationType, currently we are using units as specified in EngineeringUnits. Both attributes are optional, do we leave this as is, where sitting a ApplicationType only sets the device class, should it also set the units, if so, what should happen if both attributes are set?

zha/application/platforms/sensor/__init__.py
```diff
@@ -502,16 +515,85 @@ def formatter(self, value: int) -> str | None:
...
+        cluster_handlers: list[ClusterHandler],
+        endpoint: Endpoint,
+        device: Device,
+        **kwargs: Any,
+    ) -> None:
+        """Init this sensor."""
+        super().__init__(unique_id, cluster_handlers, endpoint, device, **kwargs)
+        engineering_units = self._cluster_handler.engineering_units
+        self._attr_native_unit_of_measurement = UNITS.get(engineering_units)
+        self._attr_state_class = SensorStateClass.MEASUREMENT
```
prairiesnpr: Should this always be measurement? I can see a situation where total or total increasing makes sense, but I don't see a good way to choose based on the attributes we have available.

TheJulianJES (commented):
There are some restrictions on what we can provide as a `translation_key` to Home Assistant. AFAIK, we are not allowed to provide translation keys blindly, so not directly from the `description` attribute, for example.

We'd need to map all known `description` attribute values to translation keys (and fall-back to a default or untranslated name otherwise).

So, I'm wondering if it would make sense to split this PR into two: one with the translation key changes and one with the added `AnalogInput` and `MultistateInput` sensors.

prairiesnpr:
> There are some restrictions on what we can provide as a `translation_key` to Home Assistant. AFAIK, we are not allowed to provide translation keys blindly, so not directly from the `description` attribute, for example.
> 
> We'd need to map all known `description` attribute values to translation keys (and fall-back to a default or untranslated name otherwise).
> 
> So, I'm wondering if it would make sense to split this PR into two: one with the translation key changes and one with the added `AnalogInput` and `MultistateInput` sensors.

We aren't modifying the `translation_key` here, if a description attribute is provided, we set the key to None and then provide a `fallback_name`, that way we can use arbitrary values without having a matching `translation_key`.  This follows the same logic currently used for custom quirks https://github.com/home-assistant/core/blob/f1e4229b23ef7837bb4e46877f55d6533f8d4607/homeassistant/components/zha/entity.py#L46

As far as splitting the PR, it was originally planned to have multiple PRs, one for the `fallback_name` functionality and then additional PRs for each cluster. I ran into issues with writing tests there, since we really needed them to both be present to get the required coverage. That said, I'm not opposed to splitting it, if we have a plan on how best to proceed.

zha/application/platforms/sensor/const.py
puddly: > If the ApplicationType attribute is implemented, and is set to a value with a defined physical unit, the physical unit defined in ApplicationType takes priority over EngineeringUnits.

I think we should keep things as simple as possible. The spec is expansive and we should only implement the bare minimum to support real devices.

zha/application/platforms/sensor/__init__.py
```diff
@@ -502,16 +515,85 @@ def formatter(self, value: int) -> str | None:
...
+        cluster_handlers: list[ClusterHandler],
+        endpoint: Endpoint,
+        device: Device,
+        **kwargs: Any,
+    ) -> None:
+        """Init this sensor."""
+        super().__init__(unique_id, cluster_handlers, endpoint, device, **kwargs)
+        if self._cluster_handler.cluster.get("state_text"):
+            self._enum = enum.Enum(  # type: ignore [misc]
+                "state_text", self._cluster_handler.cluster.get("state_text")
```
puddly:
```suggestion
                "state_text", self._cluster_handler.cluster["state_text"]
```

zha/application/platforms/sensor/__init__.py
```diff
@@ -502,16 +515,85 @@ def formatter(self, value: int) -> str | None:
...
+            )
+        elif self._cluster_handler.cluster.get("number_of_states") is not None:
+            self._enum = enum.Enum(  # type: ignore [misc]
+                "state_text",
+                [
+                    (f"state_{i+1}", i + 1)
+                    for i in range(
+                        self._cluster_handler.cluster.get("number_of_states")
```
puddly:
```suggestion
                        self._cluster_handler.cluster["number_of_states"]
```

zha/application/platforms/sensor/const.py
```diff
@@ -390,3 +390,36 @@ class SensorDeviceClass(enum.StrEnum):
...
+    """State class for AnalogInput Types."""
+
+    TEMPERATURE = (0x00, SensorDeviceClass.TEMPERATURE)
```
puddly: Let's move this enum to zigpy. We generally don't want to define ZCL enums and types in the ZHA library itself. From the spec:

image

I think we can define this easily in zigpy:

```python
class AnalogInputType(t.enum8):
    Temp_In_Degrees_C = 0x00
    ...


ANALOG_INPUT_TYPES = {
    AnalogInputType.Temp_In_Degrees_C: TempInDegreesC,
}


class TempInDegreesC(t.enum16):
    Two_Pipe_Entering_Water_Temperature = 0x0000
    ...


class ApplicationType(t.uint32_t, t.Struct):
    group: t.uint8_t
    type: AnalogInputType
    index: t.uint16_t
```

And then ZHA will only deal with creating a mapping between the ZCL spec units and the HA device classes.

prairiesnpr:
This will need a zigpy release before tests will pass.

dmulcahey:
I think we should update the test json files... that's a lot to delete. Also, should we start these off as disabled by default?

puddly:
I've pushed a whole host of changes to this branch, apologies! There were a few bugs that needed fixing in ZHA (and thus some tweaks to Core):
- Enum entities now implement `_attr_options`, which enumerate the possible entries. This is added to the info dict for the entity and will need to be exposed to HA Core.
- The `description` attribute is *required* for both types of entities to be auto-created. This is unfortunate but something that we need to do until ZHA discovery is hammered into better shape.
- All of the BACnet units have been re-mapped.

@P-R-O-C-H-Y Give this branch a try.

For analog input:
1. `description` attribute must exist.
2. `engineering_units` or `application_type` (preferred) must be implemented.

For multistate input:
1. `description` attribute must exist.
2. The `state_text` array must exist OR `number_of_states`. I'm debating whether or not to allow `number_of_states` and just require `state_text`.

The only implementation detail I'm still unsure of is `state_text`: the ZCL states:

> The `PresentValue`, interpreted as an integer, serves as an index into the array.

In the ZCL, arrays are bizarrely both 0- and 1-indexed: the zero-th element is the size of the array, and the first element is the first element is the first. I think this means that `present_value == 1` is the first element of the `state_text` array, so it's one-indexed? I would need a certified device to confirm and I unfortunately don't have one at the moment.

P-R-O-C-H-Y:
@puddly I am having some issues with Analog Input. It is showing up now, but the Description and Application type is not used.
I did send more info to you on Discord.

puddly:
@P-R-O-C-H-Y This should be fixed by https://github.com/zigpy/zigpy/pull/1593. With the device diagnostics you provided, the device is correctly shown:

```diff
       {
         "info_object": {
+          "fallback_name": "Power Consumption (Watts)",
           "unique_id": "ab:cd:ef:12:39:a2:73:55-1-12-analog_input",
           "migrate_unique_ids": [],
           "platform": "sensor",
           "class_name": "AnalogInputSensor",
           "translation_key": null,
           "device_class": "power",
           "state_class": "measurement",
           "entity_category": null,
           "entity_registry_enabled_default": true,
           "enabled": true,
           "primary": false,
           "cluster_handlers": [
             {
               "class_name": "AnalogInputClusterHandler",
               "generic_id": "cluster_handler_0x000c",
               "endpoint_id": 1,
               "cluster": {
                 "id": 12,
                 "name": "AnalogInput",
                 "type": "server"
               },
               "id": "1:0x000c",
               "unique_id": "ab:cd:ef:12:39:a2:73:55:1:0x000c",
               "status": "INITIALIZED",
               "value_attribute": "present_value"
             }
           ],
           "device_ieee": [ 85, 115, 162, 57, 18, 239, 205, 171],
           "endpoint_id": 1,
           "available": true,
           "group_id": null,
           "attribute": "present_value",
           "suggested_display_precision": null,
           "divisor": 1,
           "multiplier": 1,
+          "unit": "W",
           "options": null
         },
         "state": {
           "class_name": "AnalogInputSensor",
           "available": true,
+          "state": 280.0
         }
```

The unit is correctly identified as well. If you implement the `resolution` attribute, `suggested_display_precision` will be computed too.

P-R-O-C-H-Y:
@puddly I did pull latest commits and I still see the AnalogInput as temperature instead of Power Consumption

P-R-O-C-H-Y:
@puddly Can you give a hint about the `resolution` attribute? Setting its value to 0.001 does not affect the `suggested_display_precision`. Is something not working or am I setting wrong values?

puddly:
I've removed the `MultiStateInput` entity creation for now, unfortunately. We need to re-add these entities once zigpy's database issue with arrays has been resolved. This will unblock `AnalogInput` entities.

zha/application/platforms/number/__init__.py
```diff
@@ -99,6 +95,14 @@ def on_add(self) -> None:
                 self.handle_cluster_handler_attribute_updated,
             )
         )
+        if (
+            hasattr(self._analog_output_cluster_handler, "description")
```
TheJulianJES, line 99: The `hasattr` check shouldn't be needed, as the `description` property will always be present on the `AnalogOutputClusterHandler` that's guaranteed here since #442 in (now called) `AnalogOutputNumber`.
So, only the `None` check below is needed.

zha/application/platforms/sensor/__init__.py
```diff
@@ -198,6 +202,12 @@ def on_add(self) -> None:
                 self.handle_cluster_handler_attribute_updated,
             )
         )
+        if (
+            hasattr(self._cluster_handler, "description")
+            and self._cluster_handler.description is not None
+        ):
+            self._attr_translation_key = None
+            self._attr_fallback_name: str = self._cluster_handler.description
```
TheJulianJES, line 210: Having this in the base `Sensor` class feels a bit weird. The `description` attribute will only be available on the `AnalogInputClusterHandler` used for `AnalogInputSensor`.
So, shouldn't we just move this part there (and remove the `hasattr` check as well)?

tests/data/devices/tze200-2aaelwxk-ts0225.json
```diff
@@ -1874,12 +1874,12 @@
...
+          "fallback_name": "fading_time",
           "unique_id": "ab:cd:ef:12:94:85:f7:5f-2-13",
           "migrate_unique_ids": [],
           "platform": "number",
           "class_name": "Number",
-          "translation_key": "number",
+          "translation_key": null,
           "device_class": null,
           "state_class": null,
           "entity_category": null,
```
TheJulianJES, line 1885: More of a note: These entities shouldn't be present anymore.

We've replaced these `AnalogOutput` config entities for a lot of Tuya sensors with proper quirks v2 entities.
However, they'll still be in the zigpy attribute cache, until the device is re-paired.

I thought that the quirks v1 signature/replacement shouldn't persist to the database, but I remember we've had that issue at one point, so the quirked signature might be stuck in the database, or these were saved when the v1 quirks were still present and the regeneration of diagnostics just looked at the quirked signature and not the original signature?

We might wanna add a `.removes(AnalogInput.cluster_id, endpoint_id=xx)` to a whole lot of Tuya quirks for the various fake config entities that were added in the past...
Assuming these entities are still "provided" (but non-functional) for real users.

TheJulianJES (commented):
Two small possible improvements with the `description` attribute check:

P-R-O-C-H-Y:
@puddly @TheJulianJES Can we take a look again into this and add MultiState Input/Output? Thanks

PR #199, opened 2024-09-09
Allow setting sensor translation key, device/state class, and unit simultaneously

GitforZhangXL:
fix missing attr like uint

### DIFF
--- a/zha/application/platforms/sensor/__init__.py
+++ b/zha/application/platforms/sensor/__init__.py
@@ -202,6 +202,10 @@ class Sensor(PlatformEntity):
             self._divisor = entity_metadata.divisor
         if entity_metadata.multiplier is not None:
             self._multiplier = entity_metadata.multiplier
+        if entity_metadata.translation_key is not None:
+            self._attr_translation_key = entity_metadata.translation_key
+        if entity_metadata.state_class is not None:
+            self._attr_state_class = entity_metadata.state_class.value
         if entity_metadata.device_class is not None:
             self._attr_device_class = validate_device_class(
                 SensorDeviceClass,
@@ -209,7 +213,7 @@ class Sensor(PlatformEntity):
                 Platform.SENSOR.value,
                 _LOGGER,
             )
-        if entity_metadata.device_class is None and entity_metadata.unit is not None:
+        if entity_metadata.unit is not None:
             self._attr_native_unit_of_measurement = validate_unit(
                 entity_metadata.unit
             ).value


zha/application/platforms/sensor/__init__.py
```diff
@@ -202,14 +202,18 @@ def _init_from_quirks_metadata(self, entity_metadata: ZCLSensorMetadata) -> None
...
+            self._attr_state_class = entity_metadata.state_class.value
         if entity_metadata.device_class is not None:
             self._attr_device_class = validate_device_class(
                 SensorDeviceClass,
                 entity_metadata.device_class,
                 Platform.SENSOR.value,
                 _LOGGER,
             )
-        if entity_metadata.device_class is None and entity_metadata.unit is not None:
+        if entity_metadata.unit is not None:
```
dmulcahey, line 216: the device class controls the unit when it is set. This is a HA core thing and not controlled by us.

zha/application/platforms/sensor/__init__.py
```diff
@@ -202,14 +202,18 @@ def _init_from_quirks_metadata(self, entity_metadata: ZCLSensorMetadata) -> None
             self._divisor = entity_metadata.divisor
         if entity_metadata.multiplier is not None:
             self._multiplier = entity_metadata.multiplier
+        if entity_metadata.translation_key is not None:
```
dmulcahey, line 205: I think there is a restriction on translation key too if device class is set. I'll peek at core when I have a chance to validate this.

zha/application/platforms/sensor/__init__.py
TheJulianJES: I think we still need to provide the unit to let HA know what unit we're using. Take a look at our existing sensors in ZHA:
https://github.com/zigpy/zha/blob/e1ae71cc31cd53700038881de1f4ee51586caca9/zha/application/platforms/sensor/__init__.py#L585-L593

zha/application/platforms/sensor/__init__.py
TheJulianJES: I don't think it's a restriction in HA. For some device classes (like battery), there's just a standard name, so we don't need to set the name (or translation key) again if we want to use that default.

We also set both for some sensors in ZHA:
https://github.com/zigpy/zha/blob/e1ae71cc31cd53700038881de1f4ee51586caca9/zha/application/platforms/sensor/__init__.py#L688-L698

TheJulianJES: However, the translation key stuff should likely be changed here, not in this sensor class:
https://github.com/zigpy/zha/blob/e1ae71cc31cd53700038881de1f4ee51586caca9/zha/application/platforms/__init__.py#L347-L349
I think we can just remove the device class translation key check. There are cases where we want to use a custom name (overriding the built-in/default name for the device class). (We already do for some ZHA sensors)

zha/application/platforms/sensor/__init__.py
TheJulianJES: This part will be resolved via:
- https://github.com/zigpy/zha/pull/219/

puddly:
Replaced by https://github.com/zigpy/zha/pull/219

PR #203, opened 2024-09-15
Add more Inovelli blue switch entities

tleydxdy:


dmulcahey:
Are these supported for all devices?

tleydxdy:
Seems like the Dimming mode option is only for VZM31-SN and VZM36, the rest are in all three device's parameter table.

dmulcahey:
> Seems like the Dimming mode option is only for VZM31-SN and VZM36, the rest are in all three device's parameter table.

We should restrict those to the models they apply to then

dmulcahey:
We also need to make sure the attributes are already merged in the mfg cluster in quirks before this can be merged. 

We really should migrate all of these to v2 quirks so it’s all in one spot at some point

tleydxdy:
do you mean these? https://github.com/zigpy/zha-device-handlers/blob/acecf704f2b253c9297d901eed8909dbd589ea97/zhaquirks/inovelli/__init__.py#L212 sorry I'm very new to this :)

dmulcahey:
> do you mean these? https://github.com/zigpy/zha-device-handlers/blob/acecf704f2b253c9297d901eed8909dbd589ea97/zhaquirks/inovelli/__init__.py#L212 sorry I'm very new to this :)

If they are common across all devices they go in the base class: https://github.com/zigpy/zha-device-handlers/blob/acecf704f2b253c9297d901eed8909dbd589ea97/zhaquirks/inovelli/__init__.py#L94

Otherwise they go in the appropriate extension

tleydxdy:
yes, all these attributes are already there.

zha/application/platforms/select.py
```diff
@@ -637,6 +637,25 @@ class InovelliNonNeutralOutputEntity(ZCLEnumSelectEntity):
     _attr_translation_key: str = "increased_non_neutral_output"
 
 
+class InovelliDimmingMode(types.enum1):
+    """Inovelli dimming mode selection."""
+
+    LeadingEdge = 0x00
+    TrailingEdge = 0x01
```
puddly, line 644: We should start migrating the Inovelli attributes in the quirk to the new Zigpy format, this will keep enums like this out of ZHA:

```python
class LeadingOrTrailingEdge(t.enum8):
    LeadingEdge = 0x00
    TrailingEdge = 0x01


class InovelliVZM31SNCluster(CustomCluster):
    name = "InovelliVZM31SNCluster"

    class AttributeDefs(BaseAttributeDefs):
        """Attribute definitions."""
        ...

        leading_or_trailing_edge = ZCLAttributeDef(
            id=0x001A,
            type=LeadingOrTrailingEdge,
            zcl_type=DataTypeId.bool, # It's really a bool on the device
            is_manufacturer_specific=True,
        )

        ...
```

PR #215, opened 2024-09-27
Expose detection parameters for SNZB-03P

carterbox:
This older(?) PIR occupancy device has some of the same parameters as the newer SNZB-06P. You can manually set them through the zigbee device management menu, but it would be nicer if there was a button in the main UI. The 03P model is a PIR sensor, but it still uses the ultrasonic tagged parameters to control the device. The PIR parameters all have a value of None. These parameters are settable if you have the official Sonoff hub/app.

### DIFF
--- a/zha/application/platforms/number/__init__.py
+++ b/zha/application/platforms/number/__init__.py
@@ -868,7 +868,7 @@ class SonoffThermostatLocalTempCalibration(ThermostatLocalTempCalibration):
 
 
 @CONFIG_DIAGNOSTIC_MATCH(
-    cluster_handler_names=CLUSTER_HANDLER_OCCUPANCY, models={"SNZB-06P"}
+    cluster_handler_names=CLUSTER_HANDLER_OCCUPANCY, models={"SNZB-06P", "SNZB-03P"}
 )
 class SonoffPresenceSenorTimeout(NumberConfigurationEntity):
     """Configuration of Sonoff sensor presence detection timeout."""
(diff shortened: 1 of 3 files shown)


carterbox:
I don't know how to test these changes. The recent splitting of zha from homeassistant-core has me confused. I own the device in question, but I'm not sure how to get these changes into my homeassistant instance which is running on a homeassisant green.

TheJulianJES:
@carterbox There are multiple ways. You could set up a custom `zha` component, change the manifest (to include a version) and swap out the `zha` library for a GitHub branch of yours, but that has the potential to break things.

Another (also hacky) way, although likely less error-prone, would be to go into the `homeassistant` container shell (e.g. via SSH or a Portainer addon, might need to unhide the container from settings there).
Then, you can edit the file via `vi` for example (or replace it via basic commands by just copying from `/config` for example). To edit:
1. Enter the `homeassistant` container shell either via a Portainer addon from the community or by entering this via SSH or the HA console: `login`, `docker exec -it homeassistant /bin/sh`
2. Enter this: `vi /usr/local/lib/python3.12/site-packages/zha/application/platforms/number/__init__.py`
3. Scroll down to the line you want to edit (line numbers are shown at the bottom)
4. Navigate your cursor to where you want to add your change
5. Press `i` for insert mode
6. Make the change
7. Press `ESC` to get out of insert mode
8. Type `:wq` (write + quit)
9. Press enter
10. Then, restart Home Assistant and see if your change works as expected.

Do note that upgrading Home Assistant will revert that change. Do also note that you can break Home Assistant. Have backups ready.

carterbox:
Thanks for the help. I was able to edit the files directly once I setup developer access over SSH to the homeassistant host. The SSH add-on does not allow access to the relevent folders the default way.

I've tested these changes. The timeout definitely works. The sensitivity probably works. It's a bit hard to tell the difference with PIR sensors.

TheJulianJES (approved):
Looks good. Thanks!

PR #221, opened 2024-10-02
Convert common fixtures to plain functions

puddly:
The dependency chains created by pytest fixtures unfortunately make selectively applying fixtures (or figuring out how they work at times) difficult. I've converted the following fixtures to normal functions that usually operate on a single `Gateway` object:

- `zha_device_from_file` -> `zigpy_device_from_device_data` and `zigpy_device_from_json`
- `device_joined` -> `join_zigpy_device`
- `zigpy_device_mock` -> `create_mock_zigpy_device`

Going forward, I think mock devices created within fixtures should ideally be replaced with real device mocks created from JSON dumps.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,7 @@
 """Test configuration for the ZHA component."""
 
 import asyncio
-from collections.abc import Awaitable, Callable, Generator
+from collections.abc import Callable, Generator
 from contextlib import contextmanager
 import itertools
 import logging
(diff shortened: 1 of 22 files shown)


dmulcahey (approved):

tests/conftest.py
```diff
@@ -303,7 +303,6 @@ async def __aexit__(
 async def zha_gateway(
     zha_data: ZHAData,
     zigpy_app_controller,
-    caplog,  # pylint: disable=unused-argument
```
dmulcahey: This isn’t unused… this causes all the logs to be captured for all tests

PR #222, opened 2024-10-04
Support Timestamp class Sensor

### LINKED ISSUE #220: Sensor can only return numeric datatypes
A Home Assistant sensor can return multiple datatypes, but we are limited currently to only numeric types.

See: https://github.com/zigpy/zha/blob/9945b6921268d59891b092fc84d107eb0e4d4369/zha/application/platforms/sensor/__init__.py#L267

This came up when developing a V2 quirk, HA expects a sensor of class TIMESTAMP to return a datetime object, yet this will fail since we are casting to a float.

prairiesnpr:
It does add support for the HVAC Thermostat attribute `setpoint_change_source_timestamp`, primarily so I would have an attribute to test against.

Closes: https://github.com/zigpy/zha/issues/220

### DIFF
--- a/zha/application/platforms/sensor/__init__.py
+++ b/zha/application/platforms/sensor/__init__.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from asyncio import Task
 from dataclasses import dataclass
+from datetime import date, datetime
 import enum
 import functools
 import logging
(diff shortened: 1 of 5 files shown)


zha/application/platforms/sensor/__init__.py
```diff
@@ -264,14 +270,27 @@ def handle_cluster_handler_attribute_updated(
...
             self.maybe_emit_state_changed_event()
 
-    def formatter(self, value: int | enum.IntEnum) -> int | float | str | None:
+    def formatter(
+        self, value: int | enum.IntEnum
+    ) -> date | datetime | int | float | str | None:
+        """Pass-through formatter."""
+        return getattr(
+            self,
+            NON_NUMERIC_FORMATTERS.get(self._attr_device_class, "numeric_formatter"),
```
puddly, line 279: I think it would be clearest to have a single `formatter` function for the base sensor (`converter`?) and then have each sub-type override it, having `NumericSensor`, `StringSensor`, and `DateTimeSensor` provide an implementation.

prairiesnpr:
Going with https://github.com/zigpy/zha/pull/224

PR #224, opened 2024-10-06
Add Timestamp Sensor

### LINKED ISSUE #220: Sensor can only return numeric datatypes

prairiesnpr:
Alternative to https://github.com/zigpy/zha/pull/222

It does add support for the HVAC Thermostat attribute setpoint_change_source_timestamp, primarily so I would have an attribute to test against.

Closes: https://github.com/zigpy/zha/issues/220

### DIFF
--- a/tests/test_sensor.py
+++ b/tests/test_sensor.py
@@ -2,6 +2,7 @@
 
 import asyncio
 from collections.abc import Awaitable, Callable
+from datetime import UTC, datetime
 import math
 from typing import Any, Optional
 from unittest.mock import AsyncMock, MagicMock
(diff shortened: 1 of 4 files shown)


tests/test_sensor.py
```diff
@@ -1126,6 +1146,78 @@ async def test_elec_measurement_skip_unsupported_attribute(
     assert read_attrs == supported_attributes
 
 
+class TimestampCluster(CustomCluster, ManufacturerSpecificCluster):
+    """Timestamp Quirk V2 Cluster."""
```
dmulcahey, line 1150: Drop quirk v2 from the comment

dmulcahey (approved):

PR #232, opened 2024-10-11
Fix `quirk_applied` check only working for v1 quirks

TheJulianJES:
https://github.com/zigpy/zigpy/pull/1450 must have broken this check to no longer be `True` for v2 quirks, since `CustomDeviceV2` no longer extends `CustomDevice`, only `BaseCustomDevice`.  This PR checks for the base class now.

(Noticed this whilst working on https://github.com/zigpy/zha/pull/231)

### DIFF
--- a/zha/zigbee/device.py
+++ b/zha/zigbee/device.py
@@ -207,7 +207,7 @@ class Device(LogMixin, EventBase):
         self._gateway: Gateway = _gateway
         self._zigpy_device: ZigpyDevice = zigpy_device
         self.quirk_applied: bool = isinstance(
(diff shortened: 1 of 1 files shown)


TheJulianJES:
I believe this also needs some fixes in zigpy.

dmulcahey (approved):

dmulcahey:
> I believe this also needs some fixes in zigpy.

What is needed there?

TheJulianJES:
Separate from this PR, but:
- https://github.com/zigpy/zigpy/pull/1485

PR #234, opened 2024-10-13
Allow translation key even when device class is overridden

benbancroft:
Since https://github.com/zigpy/zigpy/pull/1474 was merged, translation_key and/or fallback_name is a mandatory field for a V2 quirk entity, and the restriction of not being able to set them when device class is explictly set has been dropped.

From what I can tell, there isn't a good reason for this restriction, and the device class alone isn't a replacement for not having a translation key (such as the case when a device has two sensors of the same device class and you need a way to differentiate between them). If my understanding is flawed here, please let me know and we can drop this PR.

On the ZHA side, the check for device is still present. This pull request is for its removal.

### DIFF
--- a/zha/application/platforms/__init__.py
+++ b/zha/application/platforms/__init__.py
@@ -340,21 +340,19 @@ class PlatformEntity(BaseEntity):
         if entity_metadata.initially_disabled:
             self._attr_entity_registry_enabled_default = False
 
-        has_device_class = hasattr(entity_metadata, "device_class")
         has_attribute_name = hasattr(entity_metadata, "attribute_name")
         has_command_name = hasattr(entity_metadata, "command_name")
         has_fallback_name = hasattr(entity_metadata, "fallback_name")
 
-        if not has_device_class or entity_metadata.device_class is None:
-            if has_fallback_name:
-                self._attr_fallback_name = entity_metadata.fallback_name
+        if has_fallback_name:
+            self._attr_fallback_name = entity_metadata.fallback_name
 
-            if entity_metadata.translation_key:
-                self._attr_translation_key = entity_metadata.translation_key
-            elif has_attribute_name:
-                self._attr_translation_key = entity_metadata.attribute_name
-            elif has_command_name:
-                self._attr_translation_key = entity_metadata.command_name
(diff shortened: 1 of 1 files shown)


puddly:
I believe this should be good to go with the recent change to ZHA in Core: https://github.com/home-assistant/core/blob/af41a41046635ed319c4872920dd0ab3c67ab5d1/homeassistant/components/zha/entity.py#L58-L70

TheJulianJES:
So, https://github.com/zigpy/zigpy/pull/1474 doesn't seem to make fallback name or translation keys mandatory. 
The last commit is named "Drop device class validation and mandate specifying translation keys and fallback names", but removes the logging message added in previous commits in that PR. Was that intentional?

There are three ways for an HA entity to get its name:
1. translation key
2. fallback name
3. device class (e.g. battery)

Looking at the linked HA code, `super().name` should return either a translated name or the device class name.
So yeah, I think we can remove this validation.

The fallback name behavior might still cause some weird behavior though.
If an entity has all three of the above options set (translation key, fallback name, and device class) and if a translation is provided, it would use that, as expected.
But if a translation isn't provided in HA `strings.json`, then it would fall back to the device class, right? Since we'd never reach the fallback name condition in the ZHA entity class. If this is the case, that's probably not behavior we want. It should still fall back to the fallback name.
IMO, only if both do not exist, it should fall back to the device class.

puddly:
> If this is the case, that's probably not behavior we want. It should still fall back to the fallback name.

The code in Core is sort of a proxy for checking if the translation exists or not. We may be able to introspect by comparing with `self._device_class_name_helper(self.platform.component_translations)` on the entity but it's accessing private methods, I'm not too sure what the best path forward is without explicit support from Core.

TheJulianJES:
Yeah, it's not completely ideal. Let's just leave the HA code how it is at the moment.
When a v2 quirk and its translations are merged into HA, it'll always prefer the translations or fall back to the device class name, as expected (only after that, to the fallback name).
The "issue" described above should only happen during development, so it's probably fine.

TheJulianJES (approved):
This should be good to go then. Thanks!

PR #238, opened 2024-10-15
Bind and set up reporting for Hue contact sensor

mguaylam:
Report config for SOC001 Philips manufacturer specific cluster.
https://github.com/zigpy/zha-device-handlers/issues/3314


TheJulianJES:
You can ignore the failing test for now. It fails because a cluster with that id isn't defined anywhere.
The test is a bit aggressive maybe... Just for reference, it was added in this commit: https://github.com/home-assistant/core/commit/37707edc478eff9bd123f669effbeff9f317268a#diff-7126848f7becb03b017f77521684b3372d7a95f1859f6e9763493af52de8a780

TheJulianJES:
Uh, the test ignores custom clusters on v2 quirks at the moment..  Taking a look

TheJulianJES (approved):
Thanks!

PR #245, opened 2024-10-17
Convert dataclasses to pydantic models

dmulcahey:
First step towards adding web socket functionality. This PR converts the dataclass objects to Pydantic models so they can be validated, serialized and deserialized between the WS server and clients. 

**Note:** the target branch is a project branch and NOT dev.


puddly:
As long as unit tests pass, I think this may actually be something that can be merged straight into `dev`. Or do you wanna keep it separate?

dmulcahey:
> As long as unit tests pass, I think this may actually be something that can be merged straight into `dev`. Or do you wanna keep it separate?

separate. This is actually auto validating the objects on creation and I want to make sure everything is good before putting it in dev.

dmulcahey:
will recreate when ready

PR #251, opened 2024-10-21
Remove defaulting of `translation_key`

TheJulianJES:
Part 2 of changes related to: https://github.com/zigpy/zigpy/pull/1488

## Proposed change

`translation_key` no longer defaults to `attribute_name` or `command_name`.
There's more information in the linked zigpy PR below on why this should be the case.

## Additional information

Somewhat requires this zigpy PR to ensure `translation_key` or `device_class` is provided for a proper name:
- https://github.com/zigpy/zigpy/pull/1488

Requires for tests to pass:
- https://github.com/zigpy/zha/pull/252

I wanted to keep these changes separate from #252, as the changes are only semi-related.

### DIFF
--- a/zha/application/platforms/__init__.py
+++ b/zha/application/platforms/__init__.py
@@ -349,10 +349,6 @@ class PlatformEntity(BaseEntity):
 
         if entity_metadata.translation_key:
             self._attr_translation_key = entity_metadata.translation_key
-        elif has_attribute_name:
-            self._attr_translation_key = entity_metadata.attribute_name
-        elif has_command_name:
-            self._attr_translation_key = entity_metadata.command_name
 
         if has_attribute_name:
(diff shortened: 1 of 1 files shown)


puddly:
Will this require any changes in quirks, to manually add the defaults?

TheJulianJES:
Somewhat. When zigpy is bumped in the quirks repo, quirks missing `translation_key` will fail to validate.
At the moment, the zha-quirks version used for ZHA doesn't contain any quirks v2 entities.
The next version of zha-quirks will contain quirks v2 entities, but in the quirks repo, a `translation_key` and `fallback_name` is already present for all of those. (only some tests were missing it: https://github.com/zigpy/zha-device-handlers/pull/3449)

So, only custom quirks would need changes. But they'll throw a validation error with the next zigpy version if they're missing `fallback_name` or `translation_key`.

PR #252, opened 2024-10-21
Add `translation_key` and `fallback_name` to tests

TheJulianJES:
Part 1 of changes related to: https://github.com/zigpy/zigpy/pull/1488

## Proposed change

This adds `translation_key` and `fallback_name` to all tests where a v2 quirk entity is created.

## Additional information

With the PR below, zigpy always requires `fallback_name` name to be set. `translation_key` also needs to be set if `device_class` is not set.

This PR should be merged before a zigpy bump with the PR below to ensure tests still pass. However, this PR only makes sense if we go through with the changes from the PR below.
- https://github.com/zigpy/zigpy/pull/1488

### DIFF
--- a/tests/test_button.py
+++ b/tests/test_button.py
@@ -206,11 +206,15 @@ class FakeManufacturerCluster(CustomCluster, ManufacturerSpecificCluster):
         FakeManufacturerCluster.ServerCommandDefs.self_test.name,
         FakeManufacturerCluster.cluster_id,
         command_args=(5,),
+        translation_key="self_test",
+        fallback_name="Self test",
     )
     .write_attr_button(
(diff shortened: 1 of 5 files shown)


TheJulianJES:
https://github.com/zigpy/zigpy/pull/1488 is merged now, so we can already merge this to be ready for the zigpy bump.
The slight decrease in code coverage is the defaulting to `attribute_name`/`command_name` no longer being tested. https://app.codecov.io/gh/zigpy/zha/pull/252/indirect-changes

That code will be removed with the next PR: https://github.com/zigpy/zha/pull/251
I wanted to separate all test changes (this PR) from proper/behavioral changes.

PR #271, opened 2024-10-27
[DO NOT MERGE] Add web socket server

dmulcahey:


dmulcahey:
Closing this experimental WebSocket prototype. Its 113-file transport/client/server and entity-model rewrite is built around cluster-handler APIs that have since been removed, and the PR does not establish a supported deployment, authentication, and API contract sufficient to justify that maintenance scope.

Bringing this forward would require a new design project rather than a rebase. A future proposal should start with a concrete supported deployment and a focused transport/API contract against the current entity and cluster architecture.

PR #284, opened 2024-10-31
Consolidate device diagnostics and create a tool to regenerate them

puddly:
Combines the logic from HA Core, this library, and https://github.com/dmulcahey/zha-device-exporter/ into a single method: `Device.get_diagnostics_json()`.


puddly:
Everything works but with a small problem: clusters that modify the state (e.g. double battery percentage) modify it globally, so running the tool keeps doubling them after re-load. We'll need to restore entity state after loading device info, it seems.

---

Fixed with https://github.com/zigpy/zha/pull/284/commits/9f4fe5b197050efbca3af1debb761a3b9246724b.

tests/test_discover.py
```diff
@@ -755,9 +755,15 @@ async def test_devices_from_files(
...
-        zha_device = await join_zigpy_device(zha_gateway, zigpy_device)
+
+        # XXX: attribute updates during device initialization unfortunately triggers
+        # logic within quirks to "fix" attributes. Since these attributes are *read out*
+        # in this state, this will compound the "fix" repeatedly.
+        with mock.patch("zigpy.zcl.Cluster._update_attribute"):
+            zha_device = await join_zigpy_device(zha_gateway, zigpy_device)
 
         assert zha_device is not None
+        zha_device.get_diagnostics_json()
```
dmulcahey: what is this call doing?

zha/zigbee/device.py
```diff
@@ -1087,3 +1102,167 @@ def log(self, level: int, msg: str, *args: Any, **kwargs: Any) -> None:
...
+        info["power_source"] = self.power_source
+        info["lqi"] = self.lqi
+        info["rssi"] = self.rssi
+        info["last_seen"] = self.device._last_seen.isoformat()
+        info["available"] = self.available
+        info["device_type"] = self.device_type
+        info["active_coordinator"] = self.is_active_coordinator
+
+        node_desc = self.device.node_desc
+        info["node_descriptor"] = {
```
dmulcahey, line 1268: how come this is handled field by field here but `as_dict` is used below when processing `original_signature` ?

zha/zigbee/device.py
```diff
@@ -1087,3 +1102,167 @@ def log(self, level: int, msg: str, *args: Any, **kwargs: Any) -> None:
...
+                        ]
+
+            if "node_desc" in original_signature:
+                original_signature["node_desc"] = original_signature[
+                    "node_desc"
+                ].as_dict()
+
+            info["original_signature"] = original_signature
+
+        info["zha_lib_entities"] = {}
```
dmulcahey: maybe use defaultdict to remove lines 1258/1259

tests/common.py
```diff
@@ -313,123 +314,139 @@ def zigpy_device_from_device_data(
...
-        if "original_signature" in device_data
-        else endpoints
+    device.manufacturer = device_data["manufacturer"]
+    device.model = device_data["model"]
+    device.last_seen = datetime.fromisoformat(device_data["last_seen"])
+    device.lqi = int(device_data["lqi"]) if device_data["lqi"] is not None else None
+    device.rssi = int(device_data["rssi"]) if device_data["rssi"] is not None else None
+
+    node_desc = device_data["node_descriptor"]
+    device.node_desc = zdo_t.NodeDescriptor(
```
dmulcahey: can we collapse this by using `**node_desc`?

tests/test_discover.py
puddly: Just getting coverage up and making sure the existing JSON works.

zha/zigbee/device.py
puddly: `as_dict` doesn't convert enum member names and instead leaves them as integers. It's not very readable.

dmulcahey: Ah right. Forgot about that.

dmulcahey: Could we use the encoder to do the conversion?  Meh… probably not worth it.

puddly: We should have JSON-serializable version of all of these (device info, original signature, entity state, and topology info) once the Pydantic PRs start coming in. I think this should really clean up the code in `device.py` and move it into `models.py`

dmulcahey: #soon

puddly:
I've rebased this PR. I think after the new `unique_id` migrations and entity lifecycle cleanup, we should be able to merge this.

PR #297, opened 2024-11-19
Use an explicit entity category of `CONTROL` instead of `None`

puddly:
Having `None` be the "control" entity category always felt a little strange. This PR makes the mapping explicit in the ZHA lib.

Core diff:

```diff
diff --git a/homeassistant/components/zha/entity.py b/homeassistant/components/zha/entity.py
index 3e3d0642ca2..c26f93ecfb6 100644
--- a/homeassistant/components/zha/entity.py
+++ b/homeassistant/components/zha/entity.py
@@ -9,6 +9,7 @@ import logging
 from typing import Any
 
 from propcache import cached_property
+from zha.application.platforms import EntityCategory as ZhaEntityCategory
 from zha.mixins import LogMixin
 
 from homeassistant.const import ATTR_MANUFACTURER, ATTR_MODEL, ATTR_NAME, EntityCategory
@@ -45,7 +46,9 @@ class ZHAEntity(LogMixin, RestoreEntity, Entity):
         meta = self.entity_data.entity.info_object
         self._attr_unique_id = meta.unique_id
 
-        if meta.entity_category is not None:
+        if meta.entity_category is ZhaEntityCategory.CONTROL:
+            self._attr_entity_category = None
+        else:
             self._attr_entity_category = EntityCategory(meta.entity_category)
 
         self._attr_entity_registry_enabled_default = (
```

### DIFF
--- a/tests/data/devices/adurosmart-eria-ad-rgbw3001.json
+++ b/tests/data/devices/adurosmart-eria-ad-rgbw3001.json
@@ -1309,7 +1309,7 @@
           "translation_key": "light",
           "device_class": null,
           "state_class": null,
-          "entity_category": null,
+          "entity_category": "control",
           "entity_registry_enabled_default": true,
           "cluster_handlers": [
             {
--- a/tests/data/devices/adurosmart-eria-vms-adurolight.json
+++ b/tests/data/devices/adurosmart-eria-vms-adurolight.json
@@ -497,7 +497,7 @@
           "translation_key": null,
           "device_class": "motion",
           "state_class": null,
-          "entity_category": null,
+          "entity_category": "control",
           "entity_registry_enabled_default": true,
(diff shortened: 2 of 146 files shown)


TheJulianJES:
The `None` default in HA is is a bit weird, yeah..
I guess it's fine to have our `EntityCategory` deviate from HA's `EntityCategory` for a good reason, not sure if we need to do it here. Let's assume there would be a "HA entity API" in the future, we'd still need our own `EntityCategory` and map then if we want to keep `CONTROL` (and could not directly use the HA `EntityCategory` in ZHA).

Another thing: e.g. temperature sensors with (primary) sensor entities would also all have the `CONTROL` category?

puddly:
Ah, good point, I completely forgot about sensors. Ignore this PR 😄

PR #298, opened 2024-11-19
Compute primary entities

puddly:
Primary entities are picked with a simple weighting system that only picks a clear winner: if there are any ties, no primary entity is chosen.

"Control" entities are given a higher priority than sensors and some sensors internally are given higher priority than others:

1. For night lights with a motion sensor, the light entity is primary.
2. For motion sensors with occupancy, opening, and illuminance clusters, the motion entity is primary.
3. For combined temperature, humidity, and pressure sensors, there is no primary entity.
4. For power strips with multiple outlets, there is no primary entity.

For the most part it works for me for the vast majority of entities.

This approach does suffer from a few problems. Some devices have unnecessary entities created solely to control/be controlled by other devices.

For example, my blinds have an OnOff cluster:

image

And my water leak sensor has an `Opening` cluster:

image

I think we should provide an API within quirks v2 to *remove* entities instead of disabling clusters.

### DIFF
--- a/zha/application/platforms/binary_sensor/__init__.py
+++ b/zha/application/platforms/binary_sensor/__init__.py
@@ -167,6 +167,7 @@ class Occupancy(BinarySensor):
 
     _attribute_name = "occupancy"
     _attr_device_class: BinarySensorDeviceClass = BinarySensorDeviceClass.OCCUPANCY
+    _attr_primary_weight = 2
 
 
 @MULTI_MATCH(cluster_handler_names=CLUSTER_HANDLER_HUE_OCCUPANCY)
@@ -174,6 +175,7 @@ class HueOccupancy(Occupancy):
     """ZHA Hue occupancy."""
 
     _attr_device_class: BinarySensorDeviceClass = BinarySensorDeviceClass.OCCUPANCY
+    _attr_primary_weight = 3
 
 
 @STRICT_MATCH(cluster_handler_names=CLUSTER_HANDLER_ON_OFF)
@@ -182,6 +184,7 @@ class Opening(BinarySensor):
 
     _attribute_name = "on_off"
     _attr_device_class: BinarySensorDeviceClass = BinarySensorDeviceClass.OPENING
+    _attr_primary_weight = 1
 
 
 @MULTI_MATCH(cluster_handler_names=CLUSTER_HANDLER_BINARY_INPUT)
@@ -215,6 +218,7 @@ class IASZone(BinarySensor):
     """ZHA IAS BinarySensor."""
 
     _attribute_name = "zone_status"
+    _attr_primary_weight = 3
 
     def __init__(
         self,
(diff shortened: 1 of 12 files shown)


zha/application/platforms/binary_sensor/__init__.py
```diff
@@ -197,12 +200,17 @@ class Motion(Opening):
 
     _attr_device_class: BinarySensorDeviceClass = BinarySensorDeviceClass.MOTION
 
+    # Many devices support both "motion" and "occupancy", with "motion" being the more
+    # useful entity to have
+    _attr_primary_weight = 3
```
TheJulianJES: True, but this `Motion` class is the `OnOff` one, only for IKEA sensors and older Hue sensors.
IKEA motion sensors only have an `OnOff` entity, no occupancy or IAS one.

Hue sensors have the `OnOff` entity (doesn't actually work on SML001, SML002 by default IIRC, only on SML003 and SML004 that aren't even matched here), but also the occupancy entity.
The occupancy entity is actually the motion one, so it times out after 10 seconds.
This "motion"/OnOff one times out after a couple of minutes.

I'd argue this should be `_attr_primary_weight = 1`.

zha/application/platforms/binary_sensor/__init__.py
```diff
@@ -197,12 +200,17 @@ class Motion(Opening):
...
+
 
 @MULTI_MATCH(cluster_handler_names=CLUSTER_HANDLER_ZONE)
 class IASZone(BinarySensor):
     """ZHA IAS BinarySensor."""
 
     _attribute_name = "zone_status"
+    _attr_primary_weight = 2
```
TheJulianJES: This will be the motion entity for many sensors, like Aqara ones. It'll tie with the (fake) occupancy cluster we add.
I'd bump this to `_attr_primary_weight = 3`.

zha/application/platforms/binary_sensor/__init__.py
```diff
@@ -155,13 +155,15 @@ class Occupancy(BinarySensor):
...
     _attr_device_class: BinarySensorDeviceClass = BinarySensorDeviceClass.OCCUPANCY
+    _attr_primary_weight = 2
 
 
 @MULTI_MATCH(cluster_handler_names=CLUSTER_HANDLER_HUE_OCCUPANCY)
 class HueOccupancy(Occupancy):
     """ZHA Hue occupancy."""
 
     _attr_device_class: BinarySensorDeviceClass = BinarySensorDeviceClass.OCCUPANCY
+    _attr_primary_weight = 1
```
TheJulianJES: This is the motion entity for Hue sensors, I'd bump this to not tie with the OnOff/opening one, so: `_attr_primary_weight = 2` or maybe even `3`.
We should also consider changing the device class for this one in another PR.

TheJulianJES (commented):
This looks like a nice approach in general

puddly:
Feel free to update the weights however you see fit, I think I picked defaults that fixed most of the devices I saw in my own home and a few more from memory.

puddly:
Works with the new discovery PR:

image

PR #300, opened 2024-11-22
Correct Sonoff TRVZB local temperature calibration min/max

fgsch:
From https://github.com/Koenkk/zigbee-herdsman-converters/pull/8214 via https://github.com/zigpy/zha-device-handlers/pull/3358#issuecomment-2485862165.

### DIFF
--- a/zha/application/platforms/number/__init__.py
+++ b/zha/application/platforms/number/__init__.py
@@ -862,8 +862,8 @@ class ThermostatLocalTempCalibration(NumberConfigurationEntity):
 class SonoffThermostatLocalTempCalibration(ThermostatLocalTempCalibration):
     """Local temperature calibration for the Sonoff TRVZB."""
 
-    _attr_native_min_value: float = -7
-    _attr_native_max_value: float = 7
+    _attr_native_min_value: float = -12.8
+    _attr_native_max_value: float = 12.7
     _attr_native_step: float = 0.2
 
 


tr4nt0r:
I think the step should also be 0.1

fgsch:
@tr4nt0r, can you elaborate on why you think so? The original change uses 0.2.

tr4nt0r:
I can write to the `local_temperature_calibration` attribute directly and it takes any integer value without any problems. Also how would it otherwise be possible to set a value of +12.7

fgsch:
The max and min don't necessarily imply those are the values you can use; they correspond to the range for this type -128..127 (int8) before applying the multiplier.

TheJulianJES:
I think we could increase the step size if we wanted. I'd guess 0.2 was originally chosen, as it's easier doing 0.2°C steps when moving the slider in the HA frontend.

I'll merge this for now, as Z2M also uses step size 0.2 Feel free to open a separate PR if you'd prefer the smaller step size.
Does HA also limit the values to the step size when entered manually?

TheJulianJES (approved):
Thanks!

tr4nt0r:
> I think we could increase the step size if we wanted. I'd guess 0.2 was originally chosen, as it's easier doing 0.2°C steps when moving the slider in the HA frontend.
> 
> I'll merge this for now, as Z2M also uses step size 0.2 Feel free to open a separate PR if you'd prefer the smaller step size. Does HA also limit the values to the step size when entered manually?

@TheJulianJES if the number entity is used, Home Assistant validates the inputs and allows only the values within min/max and I think it also the step is validated, but I'm not 100% sure about that right now. Currently the entity for the offset has a step of 0.1 so it would change with the update. Having a step of 0.1 would make it easier when calculating the offsets via automation.

TheJulianJES:
> Currently the entity for the offset has a step of 0.1 

Ah, the ZCL entity indeed has a 0.1 step. I've only looked at the Sonoff one before. We should probably change this one then.

fgsch:
@TheJulianJES Ok, if a 0.1 step is the norm, I can open a PR to do this.

TheJulianJES:
See:
- https://github.com/zigpy/zha/pull/304
- https://github.com/zigpy/zha/pull/306

Issue #291, opened 2024-11-07
Quirks v2 unable to specify EntityType.DIAGNOSTIC for write_attr_button


jeverley:
I'm currently writing a quirk for the Aquara FP1E presence sensor https://[github.com/zigpy/zha-device-handlers/issues/3294](https://github.com/zigpy/zha-device-handlers/issues/3294) and I've observed that specifying `EntityType.DIAGNOSTIC` as the entity_type for `write_attr_button` entities results in the error:

```
Logger: zha.application.discovery
Source: components/zha/__init__.py:151
First occurred: 12:57:58 (1 occurrences)
Last logged: 12:57:58

Device: 54:ef:44:10:00:db:47:bb-Aqara Presence Sensor FP1E has an entity with details: {'cluster_details': (1, 64704, <ClusterType.Server: 0>), 'entity_metadata': WriteAttributeButtonMetadata(entity_platform=<EntityPlatform.BUTTON: 'button'>, entity_type=<EntityType.DIAGNOSTIC: 'diagnostic'>, cluster_id=64704, endpoint_id=1, cluster_type=<ClusterType.Server: 0>, initially_disabled=False, attribute_initialized_from_cache=True, translation_key='restart_device', fallback_name='Restart device', attribute_name='restart_device', attribute_value=0)} that does not have an entity class mapping - unable to create entity
```

This can be replicated with the following entity definition.

```
    .write_attr_button(
        OppleCluster.AttributeDefs.restart_device.name,
        0,
        OppleCluster.cluster_id,
        entity_type=EntityType.DIAGNOSTIC,
        translation_key="restart_device",
        fallback_name="Restart device",
    )
```
Investigating the issue it would appear this is due to a mapping for `EntityType.DIAGNOSTIC` currently being absent in the  [zha/application/discovery.py](https://github.com/zigpy/zha/blob/dev/zha/application/discovery.py) `QUIRKS_ENTITY_META_TO_ENTITY_CLASS`.

There are currently only `button.WriteAttributeButton` mappings for `STANDARD` and `CONFIG` (the default in [zigpy/quirks/v2/__init__.py](https://github.com/zigpy/zigpy/blob/dev/zigpy/quirks/v2/__init__.py) for `write_attr_button` entities).

```
QUIRKS_ENTITY_META_TO_ENTITY_CLASS = {
    (
        Platform.BUTTON,
        WriteAttributeButtonMetadata,
        EntityType.CONFIG,
    ): button.WriteAttributeButton,
[...]

PR #308, opened 2024-11-25
Add support for DIAGNOSTIC WriteAttributeButton entities

jeverley:
Addresses the bug https://github.com/zigpy/zha/issues/291 where v2 quirks can't create WriteAttributeButton entities with a DIAGNOSTIC EntityType.

This change adds the required mapping to QUIRKS_ENTITY_META_TO_ENTITY_CLASS in `discovery.py`


dmulcahey:
Just curious, what’s the use case here?

jeverley:
> Just curious, what’s the use case here?

Probably fairly niche, but some Aqara devices allow the user to trigger the device to restart (example below from my FP1E quirk):

image

This feels to me more like a diagnostic function (like Identify) rather than configuration.

PR #323, opened 2024-12-08
Fix `IndexError` in `issue_cluster_command` for IAS ACE arm response

Sisylam:
Fix issue on sending IasAce:arm_response which does not return a tuple but only a single state but in general to be more fail tolerante as we can't always guarantee the result from the multitude of Zigbee devices


Sisylam:
PS: see https://github.com/home-assistant/core/issues/131908

Sisylam:
I will try to add a tets to increase coverage

Issue #311, opened 2024-11-26
Rename `quirk_id` to `exposes_features` and allow multiple features in quirks


TheJulianJES:
## Idea

We should consider renaming `quirk_id` to `exposes_features` and allowing the possibility for one quirk to define more than one "exposed feature".

This issue is mostly here to draw up the plan and to make sure we consider everything affected (like cluster handler matching per quirk id).

## Additional information

Currently, a quirk (`CustomDevice`) can only have one `quirk_id`. There might be cases where we want to match multiple independent entities in ZHA, without having to add new quirks to the entities in ZHA.

For this use-case, we'd want `exposes_features` to be able to expose multiple features, like, for example, both the "IAS tamper bit" feature and a "Tuya config plugs" feature.
This would allow the (theoretical) example of a Tuya plug with an IAS tamper sensor matching both the special treatment in the cluster handler for Tuya plugs + creating the Tuya config entities and exposing the IAS tamper bit as an entity.

For new devices, this would require no changes to ZHA then, as we can just add all exposed features of a device to the quirk (instead of needing to give devices that support multiple features a new quirk id each and modify the entities in ZHA to also match those). This is how a (v1) quirk with multiple exposed features would set them:
`exposes_features={IAS_TAMPER_BIT, TUYA_PLUG_ONOFF}`

### Tamper bit entity example

This would be somewhat relevant for PRs like this:
- https://github.com/zigpy/zha/pull/246

Currently, we could already make that PR work if a device only needs a one `quirk_id`.
We could allow the tamper entity to match on a quirk id defined in zha-quirks (e.g. `IAS_TAMPER_BIT`).
For devices that support the tamper bit, we'd just need to set the `quirk_id` attribute on the `CustomDevice` class in zha-quirks.

An example of the match decorator for the tamper entity matching on an IAS tamper bit quirk id.
```python
@MULTI_MATCH(
    cluster_handler_names=CLUSTER_HANDLER_ON_OFF, quirk_ids={IAS_TAMPER_BIT}  # can be a list or a string
)
```
This would be renamed to `exposes_features` if we want to do this.

### Quirks v2

We'll also want to expose `quirk_id` / the new `exposes_features` via the quirk v2 API, so a quirk can be really compact to expose a feature for supporting devices.

### Considerations for cluster handler matching

https://github.com/home-assistant/core/pull/101709 added support for matching a cluster handler per quirk id, so the cluster id can be ignored.
[...]

PR #324, opened 2024-12-09
Add support for rms_current_ph_b/c

abmantis:
Add support for 3 phase current attributes from the ElectricalMeasurement cluster.
To avoid adding the 2 extra entities to all devices with an ElectricalMeasurement cluster, this only adds them if the attributes have a value. The downside is that this only adds them after a restart/reload. With https://github.com/zigpy/zha/issues/311 this could be improved to match on a `3phase` feature instead.

Related PR on HA Core: https://github.com/home-assistant/core/pull/132871


abmantis:
~~Drafting since `test_in_cluster_handler_config` is failing~~

zha/application/platforms/sensor/__init__.py
```diff
@@ -705,6 +717,35 @@ class ElectricalMeasurementRMSCurrent(PolledElectricalMeasurement):
...
+
+
+@MULTI_MATCH(cluster_handler_names=CLUSTER_HANDLER_ELECTRICAL_MEASUREMENT)
+class ElectricalMeasurementRMSCurrentPhC(ElectricalMeasurementRMSCurrentPhB):
+    """RMS current measurement."""
+
+    _attribute_name: str = "rms_current_ph_c"
+    _unique_id_suffix: str = "rms_current_ph_c"
+    _attr_translation_key: str = "rms_current_ph_c"
+
```
puddly: Does this one also need `_skip_creation_if_none = True`?

Just so I understand correctly:
1. For existing devices with these attributes you will have to click the "re-configure" button (and probably reload ZHA) to get them to show up.
2. If you don't do that, there won't be any disabled entities.
3. For new joins, they just work properly?

abmantis: > Does this one also need `_skip_creation_if_none = True`?


`ElectricalMeasurementRMSCurrentPhC` inherits from `ElectricalMeasurementRMSCurrentPhB`, so no need to repeat the attribute.



> Just so I understand correctly:
> 
>     1. For existing devices with these attributes you will have to click the "re-configure" button (and probably reload ZHA) to get them to show up.
> 
>     2. If you don't do that, there won't be any disabled entities.
> 
>     3. For new joins, they just work properly?

It works a bit differently: it adds the entities when the integration gets loaded if there is a value for the attribute already. So, this means that:
1. If the user already had a 3 phase device, after this update the new entities would just show up (because the ph_b and ph_c attributes already have values on the database).
2. For new joins the two new entities will not show until the user reboots HA or reloads ZHA. This is because when the device is added the attributes have no values.

We already used this approach for a few other entities like [this ](https://github.com/zigpy/zha/pull/324/files#diff-d003cdf3ae320a84576be48fc6cd8a446c5aecb7949a038d2601b72d6be35c72R1167).

Ideally, ZHA would add entities as soon as the attribute gets a value, but I don't think we support that ATM.
Also we could use [exposes_feaures](https://github.com/zigpy/zha/issues/311#issuecomment-2532109597) to add the entities for devices that expose a `3_phase` feature, instead of waiting for a value on the attr.

I am planning to eventually work on the [exposes_feaures](https://github.com/zigpy/zha/issues/311) feature, but IMO it is better to have support for the 3ph devices now, even if it is not perfect, than delaying it even more.

puddly (approved):
Looks good to me. @dmulcahey @TheJulianJES what do you think?

Only nitpick is the inheritance of phase C from phase B. I think it would be clearer to be explicit with `_skip_creation_if_none = True`.

abmantis:
> Only nitpick is the inheritance of phase C from phase B. I think it would be clearer to be explicit with `_skip_creation_if_none = True`.

That's what I [had initially](https://github.com/zigpy/zha/pull/324/commits/a0296f564c4c0490d60ed1993c0cb3c6108cdb8b) but changed it to reduce the number of duplicated attributes. I am fine either way so I'll revert back.

zha/zigbee/cluster_handlers/homeautomation.py
```diff
@@ -77,10 +77,26 @@ class MeasurementType(enum.IntFlag):
...
             config=REPORT_CONFIG_DEFAULT,
         ),
+        AttrReportConfig(
+            attr=ElectricalMeasurement.AttributeDefs.rms_current_max_ph_b.name,
+            config=REPORT_CONFIG_DEFAULT,
+        ),
+        AttrReportConfig(
+            attr=ElectricalMeasurement.AttributeDefs.rms_current_max_ph_c.name,
+            config=REPORT_CONFIG_DEFAULT,
+        ),
```
TheJulianJES, line 99: Similar to the comment [here](https://github.com/zigpy/zha/pull/348#discussion_r1911336824), this will also cause four additional attributes to be read at every startup per "smart plug", even if the device doesn't support the attributes.

I guess we can't really avoid this for now and maybe it doesn't really matter overall, but it might be worth looking into only polling attributes if they're not marked as unsupported on startup in the future? Since a couple of versions, the startup attribute polling can at least be disabled completely from the ZHA config UI.
Or expand "quirk id / exposes features" to also cover "rare" devices (even if they conform to ZCL spec) for phase b and c devices?

zha/application/platforms/sensor/__init__.py
```diff
@@ -1116,29 +1158,14 @@ class SmartEnergySummationReceived(PolledSmartEnergySummation):
...
-        unique_id: str,
-        cluster_handlers: list[ClusterHandler],
-        endpoint: Endpoint,
-        device: Device,
-        **kwargs: Any,
-    ) -> Self | None:
-        """Entity Factory.
-
-        This attribute only started to be initialized in HA 2024.2.0,
+    """ This attribute only started to be initialized in HA 2024.2.0,
```
TheJulianJES: Is this docstring properly formatted? Can't add a susggestion here, but this might be better:
```python
    """
    This attribute only started to be initialized in HA 2024.2.0,
    so the entity would be created on the first HA start after the
    upgrade for existing devices, as the initialization to see if
    an attribute is unsupported happens later in the background.
    To avoid creating unnecessary entities for existing devices,
    wait until the attribute was properly initialized once for now.
    """
    _skip_creation_if_none = True
```

zha/application/platforms/sensor/__init__.py
```diff
@@ -157,6 +157,7 @@ class Sensor(PlatformEntity):
     _attr_native_unit_of_measurement: str | None = None
     _attr_device_class: SensorDeviceClass | None = None
     _attr_state_class: SensorStateClass | None = None
+    _skip_creation_if_none: bool = False
```
TheJulianJES: I guess the name is fine, but to make the "if None" part a bit clearer, how about `_skip_creation_if_no_attr_cache` or (inverted) `_require_attr_cache_for_creation`?
They're both a bit long though, so your call on what you like best out of the three.

zha/application/platforms/sensor/__init__.py
```diff
@@ -705,6 +717,36 @@ class ElectricalMeasurementRMSCurrent(PolledElectricalMeasurement):
...
+
+    _attribute_name = "rms_current_ph_b"
+    _unique_id_suffix = "rms_current_ph_b"
+    _attr_translation_key: str = "rms_current_ph_b"
+    _skip_creation_if_none = True
+
+    @property
+    def _max_attribute_name(self) -> str:
+        """Return the max attribute name."""
+        return "rms_current_max_ph_b"
```
TheJulianJES: Only minor too, but since we override the property with a static string twice, we could have a `_attr_max_attribute_name` attribute that's `None` by default in the base `Sensor` class.

Then have the property in `Sensor` look like this:
```python
    @property
    def _max_attribute_name(self) -> str:
        """Return the max attribute name."""
        return _attr_max_attribute_name or f"{self._attribute_name}_max"
```

That would allow us to only define `_attr_max_attribute_name: str = "rms_current_max_ph_b"` (or `ph_c`) in the other classes and we don't have to override the whole property every time and have the attribute defined in a way similar to others. We only do this twice at the moment, so eh, do this if you like it.

TheJulianJES (commented):
Added some small comments.

I think we can do this generally, with the only real caveat being that we will poll four additional attributes on startup for each device with an EM cluster, even if most devices never support `ph_b` and `ph_c`.
.. It should be fine though (for now)? Users will large networks can disable the "initialize attributes" on startup option, but that's also not a great solution.
We might be able to just never poll unsupported attributes on startup again. (If we want to do this, it's for a future PR and I don't consider that to be blocking this.)

zha/zigbee/cluster_handlers/homeautomation.py
abmantis: Since there 3 phase devices are not super common, maybe the `exposes features` quirk would be fine for this.
We could also make it so that for devices that do not have the `exposes features` flag we only setup attr reporting after we get at least one value (similar to what we have with the `_skip_creation_if_none`). But I am not sure if a device will send anything before attr reporting, so that may just not work at all.

zha/application/platforms/sensor/__init__.py
abmantis: I like `_skip_creation_if_no_attr_cache`

zha/application/platforms/sensor/__init__.py
abmantis: Nice tip. Since it is going to be used by other attributes in the followup PR, its a nice cleanup.

abmantis:
> We might be able to just never poll unsupported attributes on startup again.

You mean not polling if we never got a value after the device joins?

TheJulianJES:
> You mean not polling if we never got a value after the device joins?

If we poll an attribute the first time a device joins, the device can report that attribute as "unsupported" and zigpy will add it to `cluster.unsupported_attributes`. This is persistent between restarts (saved in zigpy db), so theoretically, we could just adapt this check to only get attributes if they're not in `self.cluster.unsupported_attributes` here:
https://github.com/zigpy/zha/blob/fd6cf2f570f1778a9c0933a56ebf476caafe859d/zha/zigbee/cluster_handlers/__init__.py#L458-L460

TheJulianJES (approved):
Looks good. Thanks!

PR #325, opened 2024-12-11
Add room_temperature attribute for Sinope floor heat thermostat

decompil3d:
Hi folks - this is my first time contributing here. I've got this Sinope floor heat thermostat. The quirk for it already knows about the `room_temperature` attribute. But I want to expose that attribute as a Home Assistant sensor.

I think this is how to do it, but I'd appreciate any feedback. Also, I'm not sure how to test this - is there an easy way to try this on my HA instance to verify that it works?

Thanks for your help!

### DIFF
--- a/zha/zigbee/cluster_handlers/manufacturerspecific.py
+++ b/zha/zigbee/cluster_handlers/manufacturerspecific.py
@@ -535,8 +535,8 @@ class SinopeManufacturerClusterHandler(ClusterHandler):
 
         if self.cluster.endpoint.model == "TH1300ZB":
             self.ZCL_INIT_ATTRS["room_temperature"] = True
-            return # Don't apply switch-specific attributes below
-        
+            return  # Don't apply switch-specific attributes below
+
         self.ZCL_INIT_ATTRS = {
             "double_up_full": True,
             "on_led_color": True,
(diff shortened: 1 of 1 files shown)


TheJulianJES:
This won't create a new entity. It'll only initialize (read) the `room_temperature` attribute upon pairing.
We should be able to create a new `sensor` entity via a v2 quirk. 

The [original `TH1300ZB` v1 quirk](https://github.com/zigpy/zha-device-handlers/blob/d3d43b17294bf9e2e783c3cfb299824d22ff92c2/zhaquirks/sinope/thermostat.py#L463-L521) would have to be converted to a v2 quirk first.
Then, we can add `.sensor`. Take a look at other quirks in the repo that use [`.sensor`](https://github.com/search?q=repo%3Azigpy%2Fzha-device-handlers+.sensor%28&type=code) or [`.tuya_sensor`](https://github.com/search?q=repo%3Azigpy%2Fzha-device-handlers+.tuya_sensor%28&type=code).

In the end, it should look something like this I think:
```python
(
    QuirkBuilder(SINOPE, "TH1300ZB")
    .replaces(SinopeTechnologiesElectricalMeasurementCluster)
    .replaces(SinopeTechnologiesThermostatCluster)
    .replaces(SinopeTechnologiesManufacturerCluster)
    .sensor(
        attribute_name=SinopeTechnologiesManufacturerCluster.AttributeDefs.floor_temperature.name,
        cluster_id=SinopeTechnologiesManufacturerCluster.cluster_id,
        device_class=SensorDeviceClass.TEMPERATURE,
        state_class=SensorStateClass.MEASUREMENT,
        unit=UnitOfTemperature.CELSIUS,
        translation_key="room_temperature",
        fallback_name="Room temperature",
    )
    .add_to_registry()
)
```

Drafting this PR for now.

PR #328, opened 2024-12-15
Allow device cluster entities overwrite from v2 quirks

mrrstux:
Entities created by a v2 quirk should be allowed to overwrite any default generated ones.

Current patch harmonizes entities unique ids format across the component. It requires HA entities unique id migration code.

Fixes https://github.com/home-assistant/core/issues/128253.


zha/application/discovery.py
```diff
@@ -314,7 +314,7 @@ def discover_quirks_v2_entities(self, device: Device) -> None:
                 endpoint.async_new_entity(
                     platform=platform,
                     entity_class=entity_class,
-                    unique_id=endpoint.unique_id,
+                    unique_id=f"{endpoint.unique_id}-{cluster.cluster_id}",
```
dmulcahey: I think this will break all existing v2 entities and it will require a migration of the HA entity registry.

zha/application/platforms/__init__.py
```diff
@@ -310,15 +310,14 @@ def __init__(
...
-        else:
-            _LOGGER.debug(
-                "Not registering entity %r, unique id %r already exists: %r",
-                self,
+        if (self.PLATFORM, self.unique_id) in self._device.platform_entities:
+            _LOGGER.warning(
+                "Duplicate entity detected - unique id %r already exists: %r. Replacing with %r",
                 (self.PLATFORM, self.unique_id),
                 self._device.platform_entities[(self.PLATFORM, self.unique_id)],
+                self,
```
dmulcahey: What is this doing?

zha/application/platforms/__init__.py
```diff
@@ -310,15 +310,14 @@ def __init__(
...
             )
+        self._device.platform_entities[(self.PLATFORM, self.unique_id)] = self
```
dmulcahey: Same question as above

zha/application/discovery.py
mrrstux: Is that migration a post-install operation?

zha/application/platforms/__init__.py
mrrstux: Intention was to offer a debugging message for those cases where it is not intended or expected to encounter duplicated entities. Can be removed or demoted to debug.

zha/application/platforms/__init__.py
mrrstux: It is allowing a newer entity instance to overwrite an existing one. In my specific case, an automatic zha created local temp calibration entity would be replaced by one provided by a quirk v2. At least, in my setup, the quirk v2 entity gets created last.

zha/application/discovery.py
mrrstux: In ZHA `async_setup_entry` I can get access to all HA ZHA entities. But it doesn't look like the HA associated entity data contains the info I need to migrate the ids - I would not know the cluster id of each entity in order to migrate to `<unique id>-<cluster id>`.

Here is how the registry entry looks like for a local temperature calibration entity:

```
'RegistryEntry(entity_id='number.thermostat_trv_local_temperature_offset', unique_id='28:fa:26:01:02:0b:67:2a-1-513-local_temperature_calibration', platform='zha', previous_unique_id=None, aliases=set(), area_id=None, categories={}, capabilities={'min': -5.0, 'max': 5.0, 'step': 0.1, 'mode': 'auto'}, config_entry_id='75fe69aebaacaaed3a012a97167d2d10', created_at=datetime.datetime(1970, 1, 1, 0, 0, tzinfo=datetime.timezone.utc), device_class=None, device_id='9e9268b18b27e97adb5d7543c1df8f62', disabled_by=None, entity_category=<EntityCategory.CONFIG: 'config'>, hidden_by=None, icon=None, id='617948e1ccca4e53075e64956998a17e', has_entity_name=True, labels=set(), modified_at=datetime.datetime(2024, 12, 16, 19, 51, 43, 399371, tzinfo=datetime.timezone.utc), name=None, options={'conversation': {'should_expose': False}}, original_device_class=None, original_icon=None, original_name='Local temperature offset', supported_features=0, translation_key='local_temperature_calibration', unit_of_measurement='°C', _cache={})'.
```

Is there any other place where the migration could be done?

mrrstux: Found a way: inside BaseEntityInfo, the previous unique id can be stored and used by ZHA to migrate existing HA entities.

mrrstux: Here is the migration code: https://github.com/home-assistant/core/pull/136788.

puddly:
@mrrstux I'm going to pick this PR up to rebase it on top of `dev` and then deal with a bunch of corner cases for migration. Do you mind if I push commits straight to your branch? Or would you prefer I make a separate PR to supersede this one?

mrrstux:
@puddly Yes, by all means, please take it over!

puddly:
I think this PR is ready to go for the upcoming beta.

This will affect every ZHA entity in Home Assistant so I think the Home Assistant side of this change needs a bit of a rework and to pre-compute the migration in order to perform it pseudo-atomically.

puddly:
I've scaled back the changes in this PR to *not* perform any migrations: it only affects the way entities are created and adds a new attribute to entity info to hold migrations. The format for it is provisional and likely will change (likely to include a version prefix for performance reasons?).

PR #329, opened 2024-12-16
Configure local temperature calibration range for Bosch thermostats

mrrstux:
Applies to Bosch Zigbee thermostats models: 
- RBSH-RTH0-ZB-EU
- RBSH-TRV0-ZB-EU
- RBSH-TRV1-ZB-EU
- RBSH-RTH0-BAT-ZB-EU

Workaround until (if ever) the general solution https://github.com/zigpy/zha/pull/328 gets adopted.

Fixes https://github.com/home-assistant/core/issues/128253 for Bosch Zigbee thermostats.

Required by https://github.com/zigpy/zha-device-handlers/pull/2808


TheJulianJES (approved):
Thanks. This is good for now, as we already do the same for Sonoff TRVs.

Issue #327, opened 2024-12-12
`validate_unit` doesn't allow valid values


TheJulianJES:
First, `validate_unit` typing is `Enum`, although it's typed as `str` everywhere else, so in quirks v2 metadata, and so on.
In ZHA and HA, `_attr_native_unit_of_measurement` is also typed as `str | None`.

It should be changed to accept and return a `str` IMO.

The mapping is also not ideal, as we have units replicated in both zigpy and ZHA now. We should change this in the future, e.g. by just having it in zigpy (and maybe renaming the `quirks/v2` where it is currently to something else). Or a separate lib?

We also need to fix not accepting the units strings that are defined, like `PERCENTAGE`. We need to allow those.
Actually, I don't think we really need the validation for a string anyways. Quirks v2 entities are still reviewed and obviously need to use the constants defined here before they can be merged.
HA might also already validate the units used?

Link to method (and units file):

https://github.com/zigpy/zha/blob/c013c5bfa754605606394307d911147760aaa8bd/zha/units.py#L179-L181

cc @prairiesnpr

TheJulianJES:
Related (and PRs blocked by the validation):
- https://github.com/zigpy/zha/issues/326
- https://github.com/zigpy/zha-device-handlers/pull/3563
- https://github.com/zigpy/zha-device-handlers/pull/3575
- https://github.com/zigpy/zha-device-handlers/issues/3356
- https://github.com/zigpy/zha-device-handlers/issues/3582
- https://github.com/zigpy/zha-device-handlers/pull/3635
- https://github.com/zigpy/zha-device-handlers/pull/3656
- https://github.com/zigpy/zha-device-handlers/pull/3673
- https://github.com/zigpy/zha-device-handlers/pull/3683/
- https://github.com/zigpy/zha-device-handlers/issues/1286

TheJulianJES:
I'll have to double check, but I don't think ESPHome validates the provided `unit_of_measurement`.
HA accepts any values, even something like `steps`. Unit translations were also added for units like that.

In the worst case, HA logs a warning if the unit doesn't match the state class. The state class and the unit playing together is the important part. If there's no state class, all units should be valid.
Still, we review v2 quirks before merging them, so state class and units have to match. We don't have this validation for entities added in ZHA (and it's reviewed by the same people as v2 quirks).

So, we really should just drop this validation for units IMO. There's nothing positive we get from it.

prairiesnpr:
[...]

Issue #326, opened 2024-12-11
Sensor init fails if unit of measurement is a string


tr4nt0r:
When providing a unit in the tuya V2 quirk builder that is a string (instead of an Enum) the sensor init fails. 
I used the unit lx, which is defined in the const `LIGHT_LUX`

```
    .tuya_sensor(
        dp_id=101,
        attribute_name="measured_value",
        type=t.uint16_t,
        fallback_name="Illuminance",
        device_class=SensorDeviceClass.ILLUMINANCE,
        state_class=SensorStateClass.MEASUREMENT,
        unit=LIGHT_LUX,
    )

```
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/zha/application/gateway.py", line 335, in create_platform_entities
    platform_entity = platform_entity_class.create_platform_entity(
        *args, **kw_args
    )
  File "/usr/local/lib/python3.13/site-packages/zha/application/platforms/sensor/__init__.py", line 186, in create_platform_entity
    return cls(unique_id, cluster_handlers, endpoint, device, **kwargs)
  File "/usr/local/lib/python3.13/site-packages/zha/application/platforms/sensor/__init__.py", line 198, in __init__
    super().__init__(unique_id, cluster_handlers, endpoint, device, **kwargs)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/zha/application/platforms/__init__.py", line 297, in __init__
    self._init_from_quirks_metadata(entity_metadata)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/zha/application/platforms/sensor/__init__.py", line 220, in _init_from_quirks_metadata
    self._attr_native_unit_of_measurement = validate_unit(
                                            ~~~~~~~~~~~~~^
        entity_metadata.unit
        ^^^^^^^^^^^^^^^^^^^^
    ).value
    ^
  File "/usr/local/lib/python3.13/site-packages/zha/units.py", line 181, in validate_unit
    return UNITS_OF_MEASURE[type(external_unit).__name__](external_unit.value)
           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #331, opened 2024-12-19
Improved validate unit to handle strings and to raise exception for invalid units

cfulton:
Improved the validate_unit method so that it can handle both string and enums. 
* Created set of all supported units. Set includes current enums and consts in units.py
* Handled if unit was string or enum. 
* Checked if provided unit is in set of all supported units. Return unit if it is in set.
* raise new InvalidUnitOfMeasureException if we didn't return a valid unit. 
* Changed return to always be string and updated implementation. 

Tests created to validate scenarios of:
* Tested with valid units as enum values. 
* Tested with valid units as string. 
* Verified that return is string type.
* Tested Invalid unit scenarios to always raise InvalidUnitOfMeasureException

### DIFF
--- a/tests/test_units.py
+++ b/tests/test_units.py
@@ -3,28 +3,80 @@
 import enum
 
 import pytest
-from zigpy.quirks.v2.homeassistant import UnitOfPower as QuirksUnitOfPower
+from zigpy.quirks.v2.homeassistant import PERCENTAGE, UnitOfPower as QuirksUnitOfPower
 
-from zha.units import UnitOfPower, validate_unit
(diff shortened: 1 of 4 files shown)


cfulton:
This change should cover issues raised in:
* https://github.com/zigpy/zha/issues/326
* https://github.com/zigpy/zha/issues/327
* https://github.com/zigpy/zha-device-handlers/issues/3356
* https://github.com/zigpy/zha-device-handlers/pull/3563
* https://github.com/zigpy/zha-device-handlers/pull/3575

TheJulianJES:
We've gone with this approach for now: https://github.com/zigpy/zha/pull/352
Still, thank you very much for the PR!

PR #332, opened 2024-12-19
Removed ZHA units and unit validation

cfulton:
This is an alternative to PR https://github.com/zigpy/zha/pull/331

Breaking Change:
* Any quirks that use the ZHA units will need to be updated to use a string value of the unit or to use the homeassistant const units enums instead.

This PR:
* Removed unit definitions from ZHA
* Updated references to units from HA instead
* Removed validate units on quirks

Should resolve the following issues:
* https://github.com/zigpy/zha/issues/326
* https://github.com/zigpy/zha/issues/327
* https://github.com/zigpy/zha-device-handlers/issues/3356
* https://github.com/zigpy/zha-device-handlers/pull/3563
* https://github.com/zigpy/zha-device-handlers/pull/3575


cfulton:
[a0e2b93](https://github.com/zigpy/zha/pull/332/commits/a0e2b93f0e35323174209871ed4f57cd609b67f1)

This commit is in case we do not want to have a dependency on home assistant.

TheJulianJES (commented):
We do not want to completely remove units. The issue is that they're duplicated in both [quirks v2 in zigpy](https://github.com/zigpy/zigpy/blob/dev/zigpy/quirks/v2/homeassistant/__init__.py) and in ZHA right now.

TheJulianJES:
Noting again here for reference, we've gone with this approach for now: https://github.com/zigpy/zha/pull/352
And again, thank you very much for the PR!

PR #339, opened 2024-12-24
Add remaining 3 phase electrical attributes

abmantis:
Follow up to https://github.com/zigpy/zha/pull/324

Related HA PR: https://github.com/home-assistant/core/pull/133969


zha/zigbee/cluster_handlers/homeautomation.py
```diff
@@ -77,10 +77,26 @@ class MeasurementType(enum.IntFlag):
...
             config=REPORT_CONFIG_DEFAULT,
         ),
+        AttrReportConfig(
+            attr=ElectricalMeasurement.AttributeDefs.rms_current_max_ph_b.name,
+            config=REPORT_CONFIG_DEFAULT,
+        ),
+        AttrReportConfig(
+            attr=ElectricalMeasurement.AttributeDefs.rms_current_max_ph_c.name,
+            config=REPORT_CONFIG_DEFAULT,
+        ),
```
TheJulianJES: ~~The not so great thing about this is that we'll add another four attributes to poll per device on each startup. 
With https://github.com/zigpy/zha/pull/324, that makes it eight attributes that are polled (per device on startup).~~

That isn't ideal, so we might want to look at only polling the attributes [here](https://github.com/zigpy/zha/blob/fd6cf2f570f1778a9c0933a56ebf476caafe859d/zha/zigbee/cluster_handlers/__init__.py#L458-L460) if they're not marked as unsupported already. Unsupported attributes might change with firmware updates, but otherwise shouldn't (hopefully).

TheJulianJES: Ah never mind, this PR includes the changes from the other one. So it's only four additional attributes (added in the previous PR).

TheJulianJES: But we'd also need to set up attribute reporting for power, power factor, and voltage here, right?
That would make a lot of attributes that are read on each startup.

Like mentioned above, we might want to consider only polling the attributes if they're not marked as unsupported already OR allow setting up attribute reports without also polling them on each startup?

zha/zigbee/cluster_handlers/homeautomation.py
```diff
@@ -126,7 +170,6 @@ class MeasurementType(enum.IntFlag):
         ElectricalMeasurement.AttributeDefs.measurement_type.name: True,
         ElectricalMeasurement.AttributeDefs.power_divisor.name: True,
         ElectricalMeasurement.AttributeDefs.power_multiplier.name: True,
-        ElectricalMeasurement.AttributeDefs.power_factor.name: True,
```
abmantis, line 119: I am not sure why `power_factor` was here, but I think reporting needs to be configured for it too

zha/zigbee/cluster_handlers/homeautomation.py
abmantis: should we add support for the conditional polling before this PR gets in?

zha/zigbee/cluster_handlers/homeautomation.py
TheJulianJES: According to ZCL spec, it is not reportable.

zha/zigbee/cluster_handlers/homeautomation.py
```diff
@@ -113,6 +145,18 @@ class MeasurementType(enum.IntFlag):
...
+            config=REPORT_CONFIG_OP,
+        ),
+        AttrReportConfig(
+            attr=ElectricalMeasurement.AttributeDefs.power_factor_ph_b.name,
+            config=REPORT_CONFIG_OP,
+        ),
+        AttrReportConfig(
+            attr=ElectricalMeasurement.AttributeDefs.power_factor_ph_c.name,
+            config=REPORT_CONFIG_OP,
+        ),
```
TheJulianJES: All power factor attributes do not support attribute reporting according to ZCL spec.

zha/zigbee/cluster_handlers/homeautomation.py
```diff
@@ -101,10 +117,26 @@ class MeasurementType(enum.IntFlag):
...
             config=REPORT_CONFIG_DEFAULT,
         ),
+        AttrReportConfig(
+            attr=ElectricalMeasurement.AttributeDefs.rms_voltage_max_ph_b.name,
+            config=REPORT_CONFIG_DEFAULT,
+        ),
+        AttrReportConfig(
+            attr=ElectricalMeasurement.AttributeDefs.rms_voltage_max_ph_c.name,
+            config=REPORT_CONFIG_DEFAULT,
+        ),
```
TheJulianJES: The `RMSVoltageMax` attributes also don't support attribute reporting per spec.

zha/zigbee/cluster_handlers/homeautomation.py
```diff
@@ -65,10 +65,26 @@ class MeasurementType(enum.IntFlag):
...
             config=REPORT_CONFIG_DEFAULT,
         ),
+        AttrReportConfig(
+            attr=ElectricalMeasurement.AttributeDefs.active_power_max_ph_b.name,
+            config=REPORT_CONFIG_DEFAULT,
+        ),
+        AttrReportConfig(
+            attr=ElectricalMeasurement.AttributeDefs.active_power_max_ph_c.name,
+            config=REPORT_CONFIG_DEFAULT,
+        ),
```
TheJulianJES: The `ActivePowerMax` attributes also don't support attribute reporting per spec.

zha/application/platforms/sensor/__init__.py
```diff
@@ -752,6 +774,27 @@ class ElectricalMeasurementRMSVoltage(PolledElectricalMeasurement):
...
+
+    _attribute_name = "rms_voltage_ph_b"
+    _unique_id_suffix = "rms_voltage_ph_b"
+    _attr_translation_key: str = "rms_voltage_ph_b"
+    _skip_creation_if_no_attr_cache = True
+    _attr_max_attribute_name = "rms_voltage_max_ph_b"
+
+
+@MULTI_MATCH(cluster_handler_names=CLUSTER_HANDLER_ELECTRICAL_MEASUREMENT)
+class ElectricalMeasurementRMSVoltagePhC(ElectricalMeasurementRMSVoltagePhB):
```
TheJulianJES: Let's remove the inheritance from `ElectricalMeasurementRMSVoltagePhB` and change it to `ElectricalMeasurementRMSVoltage`, similar to the other PR.
`_skip_creation_if_no_attr_cache = True` has to be added then.

zha/application/platforms/sensor/__init__.py
```diff
@@ -777,6 +820,27 @@ class ElectricalMeasurementPowerFactor(PolledElectricalMeasurement):
...
+
+    _attribute_name = "power_factor_ph_b"
+    _unique_id_suffix = "power_factor_ph_b"
+    _attr_translation_key: str = "power_factor_ph_b"
+    _skip_creation_if_no_attr_cache = True
+    _attr_max_attribute_name = "power_factor_max_ph_b"
+
+
+@MULTI_MATCH(cluster_handler_names=CLUSTER_HANDLER_ELECTRICAL_MEASUREMENT)
+class ElectricalMeasurementPowerFactorPhC(ElectricalMeasurementPowerFactorPhB):
```
TheJulianJES: Let's remove the inheritance from `ElectricalMeasurementPowerFactorPhB` and change it to `ElectricalMeasurementPowerFactor`, similar to the other PR.
`_skip_creation_if_no_attr_cache = True` has to be added then.

zha/application/platforms/sensor/__init__.py
```diff
@@ -694,6 +694,28 @@ class PolledElectricalMeasurement(ElectricalMeasurement):
...
+    _attribute_name = "active_power_ph_b"
+    _unique_id_suffix = "active_power_ph_b"
+    _attr_translation_key: str = "active_power_ph_b"
+    _use_custom_polling = False  # Poll indirectly by ElectricalMeasurementSensor
+    _skip_creation_if_no_attr_cache = True
+    _attr_max_attribute_name = "active_power_max_ph_b"
+
+
+@MULTI_MATCH(cluster_handler_names=CLUSTER_HANDLER_ELECTRICAL_MEASUREMENT)
+class ElectricalMeasurementRMSActivePowerPhC(ElectricalMeasurementRMSActivePowerPhB):
```
TheJulianJES: Let's remove the inheritance from `ElectricalMeasurementRMSActivePowerPhB` and change it to `PolledElectricalMeasurement`, similar to the other PR.
`_skip_creation_if_no_attr_cache = True` and `_use_custom_polling = False` have to be added then.

zha/zigbee/cluster_handlers/homeautomation.py
TheJulianJES: The other PR was merged without it too, so this one is also fine, but we should try and target to try implementing that before the next HA beta (/ZHA release).

zha/zigbee/cluster_handlers/homeautomation.py
TheJulianJES: Hmm, apparently we also set up attribute reporting for some attributes that don't apparently don't even support it before this PR...

zha/application/platforms/sensor/__init__.py
abmantis: Sorry, I intended to change that and forgot to do it

zha/zigbee/cluster_handlers/homeautomation.py
abmantis: Removed reporting for all the ones that are unsupported

TheJulianJES: So the "good thing" is that we already [poll all supported attributes at runtime](https://github.com/zigpy/zha/blob/a833f43c0c3507c62b4331ae102e924e86bbb96c/zha/zigbee/cluster_handlers/homeautomation.py#L132-L149), so that's why `power_factor` and others work in the first place.

But for the ones that change over time, where attribute reporting can't be set up (according to spec), we should just add them to the `ZCL_INIT_ATTRS` dict with `cache=True`, as we don't need to poll them when starting up ZHA (the `unsupported_attributes` check isn't done there), but we do need to poll them at runtime, which is done through the cluster handler, so there is one big attribute read for multiple attributes. 

But we should still move some attributes (where attribute reporting isn't possible) from `REPORT_CONFIG` to `ZCL_INIT_ATTRS`.
(EDIT: Wrote this before your comment above 😄)

zha/zigbee/cluster_handlers/homeautomation.py
```diff
@@ -85,48 +88,42 @@ class MeasurementType(enum.IntFlag):
...
-            attr=ElectricalMeasurement.AttributeDefs.ac_frequency_max.name,
-            config=REPORT_CONFIG_DEFAULT,
+            attr=ElectricalMeasurement.AttributeDefs.ac_frequency.name,
+            config=REPORT_CONFIG_OP,
         ),
     )
     ZCL_INIT_ATTRS = {
         ElectricalMeasurement.AttributeDefs.ac_current_divisor.name: True,
         ElectricalMeasurement.AttributeDefs.ac_current_multiplier.name: True,
+        ElectricalMeasurement.AttributeDefs.ac_frequency_divisor.name: True,
```
abmantis, line 137: sorted this dict alphabetically to make it easier to look things up. let me know if that is not OK and I'll revert to the previous order

zha/zigbee/cluster_handlers/homeautomation.py
```diff
@@ -85,48 +88,42 @@ class MeasurementType(enum.IntFlag):
...
         ElectricalMeasurement.AttributeDefs.ac_voltage_divisor.name: True,
         ElectricalMeasurement.AttributeDefs.ac_voltage_multiplier.name: True,
-        ElectricalMeasurement.AttributeDefs.ac_frequency_divisor.name: True,
-        ElectricalMeasurement.AttributeDefs.ac_frequency_multiplier.name: True,
+        ElectricalMeasurement.AttributeDefs.active_power_max.name: False,
         ElectricalMeasurement.AttributeDefs.measurement_type.name: True,
         ElectricalMeasurement.AttributeDefs.power_divisor.name: True,
+        ElectricalMeasurement.AttributeDefs.power_factor.name: False,
+        ElectricalMeasurement.AttributeDefs.power_factor_ph_b.name: False,
+        ElectricalMeasurement.AttributeDefs.power_factor_ph_c.name: False,
```
TheJulianJES: Per [this comment](https://github.com/zigpy/zha/pull/339#discussion_r1917450016), we should be able to use `True` for the cache here.
They are [polled using the EM cluster at runtime](https://github.com/zigpy/zha/blob/a833f43c0c3507c62b4331ae102e924e86bbb96c/zha/zigbee/cluster_handlers/homeautomation.py#L132-L149), which checks for `unsupported_attributes`, so we can avoid the extra polling during during each startup, which doesn't check for `unsupported_attributes` (yet).
Even if we eventually add it, we can avoid polling this at startup, since we'll poll it after a few seconds/minute anyway.

EDIT: The polling part is not completely correct: https://github.com/zigpy/zha/pull/339#discussion_r1918884910

zha/zigbee/cluster_handlers/homeautomation.py
abmantis: It seems that we only poll attributes that for which we set up reporting, because we only [poll the ones in `REPORT_CONFIG`](https://github.com/zigpy/zha/blob/a833f43c0c3507c62b4331ae102e924e86bbb96c/zha/zigbee/cluster_handlers/homeautomation.py#L139), so if we move them to `ZCL_INIT_ATTRS` they will not be polled, right?

Also, why do we poll the ones that have reporting set up?

TheJulianJES: Ah, you're right. I confused the startup polling doing both and this one only doing the reporting attributes.

We poll the attribute because many smart plugs (mostly Tuya) do not respect/allow setting up attribute reporting on most/all EM cluster attributes.

I do wonder a bit how some of this would work at all then, or at least how it works at the moment (with `power_factor` and so on), but maybe we've just not had any devices that support it? I wanna check how Z2M deals with some of these devices later. It has a polling configuration per plug/device at least.

abmantis: Either no one complained about `power_factor` not getting updated (since there maybe not be many devices with it), or the devices report the value even without reporting config :shrug: 

I reverted the attributes back to REPORT_CONFIG for now.

TheJulianJES:
Instead of trying to set up attribute reporting for attributes that aren't reportable according to spec, just so the EM cluster will poll them with the current implementation, we could also add another `ZCL_POLLING_ATTRS` list with just the attributes that should be polled and update `async_update` in the EM cluster handler.

Then, we'd have all attributes that need to be initialized in `ZCL_INIT_ATTRS` with `True`. They will be polled later on anyways (in batches of 5), so no need to force poll them individually at startup.
Only attributes that support attribute reporting would be in `REPORT_CONFIG`.
And all attributes that aren't reportable but which can still update over time would go in `ZCL_POLLING_ATTRS`. 

`async_update` should then be updated to poll the attributes from `ZCL_POLLING_ATTRS`, instead of `REPORT_CONFIG`.
For clarity, I'd also rather duplicate some of the attribute names between `ZCL_POLLING_ATTRS` and `REPORT_CONFIG` (instead of `async_update` combining them and polling those all).

---

Another thing, do we really need to poll the `_max` attributes all the time? Do we even display/use them anywhere?

abmantis:
> For clarity, I'd also rather duplicate some of the attribute names between `ZCL_POLLING_ATTRS` and `REPORT_CONFIG` (instead of `async_update` combining them and polling those all).

In that case, won't all the attributes in `REPORT_CONFIG` also be in `ZCL_POLLING_ATTRS` then, due to the issue you mention of Tuya devices not reporting them?

> Another thing, do we really need to poll the `_max` attributes all the time? Do we even display/use them anywhere?

To be honest I don't know. Since they were already in `REPORT_CONFIG`, I assumed they were needed and did the same for phases b/c.

TheJulianJES:
> In that case, won't all the attributes in REPORT_CONFIG also be in ZCL_POLLING_ATTRS then, due to the issue you mention of Tuya devices not reporting them?

Yeah, probably all attributes that are in `REPORT_CONFIG` should also go in `ZCL_POLLING_ATTRS` for now.
(But not all attributes that are in `ZCL_POLLING_ATTRS` should go in  `REPORT_CONFIG`.)
Other than separating it for clarity, I could see that the polling attributes change over time, since we might only need to poll some attributes (could peek into Z2M for that). Currently, we just poll all reportable attrs.

If it looks really stupid to duplicate most/all attributes, we can also combine them in `async_update` for now. Personally, I'd rather have this be "clear" and easily changeable, but if it's bad/overwhelming, we can do it differently.

abmantis:
Opened a PR with the suggested changes: https://github.com/zigpy/zha/pull/354
I'll mark this one as draft until that gets in.

zha/zigbee/cluster_handlers/homeautomation.py
```diff
@@ -116,7 +143,6 @@ class MeasurementType(enum.IntFlag):
         ElectricalMeasurement.AttributeDefs.measurement_type.name: True,
         ElectricalMeasurement.AttributeDefs.power_divisor.name: True,
         ElectricalMeasurement.AttributeDefs.power_multiplier.name: True,
-        ElectricalMeasurement.AttributeDefs.power_factor.name: True,
```
TheJulianJES, line 119: Ok, I think this is almost good to go now, except for one thing:
All attributes that are exclusively in `ZCL_POLLING_ATTRS` like `power_factor` (and I guess also the `_max` attributes) need to be added to `ZCL_INIT_ATTRS` too, with `cache=True`, so they're initialized during pairing. Otherwise, the entities will not be created until ZHA is restarted.

(Attributes that are in `ATTR_REPORTING` should *not* be duplicated into `ZCL_INIT_ATTRS` though.)

There are some other PRs I want to get in first, so I'll look into slightly changing this behavior in the future, but to have less conflicts with the (IKEA) divisor/multiplier reporting PR, let's just add the three `power_factor` and I guess also the ten `_max` attributes to `ZCL_INIT_ATTRS`.

TheJulianJES (approved):
thanks!

PR #341, opened 2024-12-28
Fix missing state_class for quirk v2 sensors

abmantis:


zha/application/platforms/sensor/__init__.py
```diff
@@ -201,6 +201,21 @@ def __init__(
...
+        try:
+            return SensorStateClass(state_class_value.value)
+        except ValueError as ex:
+            _LOGGER.warning(
+                "Quirks provided an invalid state class: %s: %s",
+                state_class_value,
+                ex,
+            )
+            return None
+
```
TheJulianJES, line 218: Even though this is only used for the sensor platform, I feel like we could still move this into the [helpers module](https://github.com/zigpy/zha/blob/dev/zha/application/platforms/helpers.py) (as a function), due to it being very similar to the device class validation and to avoid cluttering the sensor class more. wdyt?

tests/test_sensor.py
```diff
@@ -1292,6 +1320,26 @@ async def test_last_feeding_size_sensor_v2(zha_gateway: Gateway) -> None:
...
+        zha_device, platform=Platform.SENSOR, qualifier="energy_delivered"
+    )
+    energy_invalid_state_class = get_entity(
+        zha_device, platform=Platform.SENSOR, qualifier="energy_invalid_state_class"
+    )
+
+    assert power_entity.state_class == SensorStateClass.MEASUREMENT
+    assert energy_entity.state_class == SensorStateClass.TOTAL
+    assert energy_delivered_entity.state_class == SensorStateClass.TOTAL_INCREASING
+    assert energy_invalid_state_class.state_class is None
```
TheJulianJES, line 1342: For device class validation, we test this inside [`test_discovery.py`](https://github.com/zigpy/zha/blob/2d1501309bc98ee5598953e4c0d9d0dd838da344/tests/test_discover.py#L665-L705) and check for the error text.
It's probably fine to leave this in the sensor tests here though. We could add an additional check for asserting that the "Quirks provided an invalid state class [...]" message is logged.

zha/application/platforms/sensor/__init__.py
abmantis: I've placed it in the `helpers` module initially, but moved it to `sensor` because this imports `SensorStateClass` (which is in the `sensor` module) and resulted in a circular import.

tests/test_sensor.py
abmantis: I've added the log check.
Since this is for sensors only, I think it fits `test_sensor.py` better, but I am ok with moving it to `test_discovery.py` if you prefer.

zha/application/platforms/sensor/__init__.py
TheJulianJES: Ok. Let's leave it like this for now then.

TheJulianJES (approved):
Thanks!

PR #350, opened 2025-01-09
Bind Legrand cable outlet cluster

piitaya:
This PR creates a cluster handler for the Legrand cable outlet cluster, thus causing ZHA to bind the cluster and support attribute reporting.

Quirks PR:
- https://github.com/zigpy/zha-device-handlers/pull/3031


zha/zigbee/cluster_handlers/manufacturerspecific.py
```diff
@@ -577,3 +579,15 @@ def matches(cls, cluster: zigpy.zcl.Cluster, endpoint: Endpoint) -> bool:
...
+@registries.CLUSTER_HANDLER_ONLY_CLUSTERS.register(LEGRAND_MANUFACTURER_CLUSTER)
+@registries.CLUSTER_HANDLER_REGISTRY.register(LEGRAND_MANUFACTURER_CLUSTER)
+class LegrandManufacturerClusterHandler(ClusterHandler):
+    """Legrand cable outlet cluster handler."""
+
+
+@registries.CLUSTER_HANDLER_ONLY_CLUSTERS.register(LEGRAND_CABLE_OUTLET_CLUSTER)
+@registries.CLUSTER_HANDLER_REGISTRY.register(LEGRAND_CABLE_OUTLET_CLUSTER)
+class LegrandCableOutletClusterHandler(ClusterHandler):
+    """Legrand cable outlet cluster handler."""
```
TheJulianJES: Z2M only binds the second cluster `0xFC40` [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/4b7136d2c7593bc93af193e4504325424bbfe468/src/devices/legrand.ts#L568).
The (here first) `0xFC01`/`LegrandManufacturerClusterHandler` cluster is also present on a [few other devices](https://github.com/search?q=repo%3Azigpy%2Fzha-device-handlers+0xFC01&type=code), where it would be bound during pairing now.* It's likely not an issue, but maybe we don't even need to bind that other cluster? (Z2M also doesn't)

I'll go ahead and remove the `LegrandManufacturerClusterHandler` from this PR, and only keep `LegrandCableOutletClusterHandler`. That should be fine, but we can adjust later (ideally when the quirks v2 "claiming cluster handler" support is ready and merged).

---

Explanation for what this code does exactly, as it's a bit obfuscated:
Registering the cluster in `CLUSTER_HANDLER_ONLY_CLUSTERS` will cause ZHA to claim the cluster handler, thus later also binding the cluster.

We'll need to add a way to allow this via quirks v2 in the future, even if we do not need to set up attribute reporting explicitly. It could already be done via `apply_custom_configuration` in quirks, but that wouldn't show in the ZHA reconfiguration UI then.
*: This would also allow us to only bind clusters with a certain ID on some devices, but not on all. So, we could bind both Legrand clusters just via quirks v2 then (and remove this code in the future).

TheJulianJES (approved):

TheJulianJES:
Thanks! Slightly adjusted the PR per this review comment: https://github.com/zigpy/zha/pull/350#discussion_r1930990146.

PR #351, opened 2025-01-09
Add a generic Centralite thermostat quirk

gmacon:
This adds a "quirk" (if that's the right name) for the Centralite 3156105 thermostat to correctly report the HVAC Action. This thermostat's running_state includes heat even when the heat is not running, you have to look at running_mode to determine that. I noticed that this was the same as the Sinope `_rm_rs_action` override, but I just copied it.

For the time being, this is "open loop" development, since I haven't figured out how to get my modified code onto my Home Assistant appliance to try it, but I figure early feedback can't hurt.

### DIFF
--- a/zha/application/platforms/climate/__init__.py
+++ b/zha/application/platforms/climate/__init__.py
@@ -600,6 +600,38 @@ class CentralitePearl(ZenWithinThermostat):
     """Centralite Pearl Thermostat implementation."""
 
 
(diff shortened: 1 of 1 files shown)


gmacon:
I've changed my mind about the best way to address this. I'm now going to add a quirk to zha-device-handlers. I've written a quirk and tested it via the custom quirks mechanism.

PR #352, opened 2025-01-16
Remove unnecessary unit validation

### LINKED ISSUE #327: `validate_unit` doesn't allow valid values

TheJulianJES:
## Proposed change

This PR removes the broken unit validation for v2 quirks and actually allows valid units/strings.

## Additional information

Fixes https://github.com/zigpy/zha/issues/327 and unblocks a lot of v2 quirks.

Supersedes the alternative solutions:
- https://github.com/zigpy/zha/pull/332 (<- we do not want to remove all constants)
- https://github.com/zigpy/zha/pull/331 (<- IMO a bit too complicated for what we actually want and also doesn't easily allow "custom units" if they're ever needed in quirks in the future)

### Background on why this change is fine

The issue linked above (https://github.com/zigpy/zha/issues/327) contains some more background info. To summarize (and add more):

The ZHA quirks v2 implementation previously didn't allow individually defined units like `PERCENTAGE` or `LIGHT_LUX`, as these are not a `StrEnum`, but just a `str`. Now, we do allow all strings.

The ESPHome integration in Home Assistant also doesn't validate units of measurements, but does validate device classes and state classes, which we also do and where the validation actually matters. ESPHome code for reference:
[homeassistant/components/esphome/sensor.py#L78-L82](https://github.com/home-assistant/core/blob/79ee2e954bd716f78a8a3190f78ade304d3170f0/homeassistant/components/esphome/sensor.py#L78-L82)

Unlike ESPHome (where a user can use any arbitrary unit of measurement), we review the units used when quirks are merged into the quirks repo, so we already make sure the constants defined in zigpy/ZHA are used (if possible).
Still, it would even be possible to define a custom unit (str), as some official integrations in HA also do, but we haven't needed that for any v2 quirks yet.
This is very similar to how both the HA Core and ZHA repos treat (reviewing) entities/units.

And IIRC, the concerns with the initial quirks v2 implementation were mostly regarding the validation of device classes and state classes, not the unit of measurement.

### Future

In the future, we could also consider dropping the units from ZHA and just using the "quirks v2" units from zigpy (and maybe rename them in zigpy, ideally with backwards compatibility to keep custom quirks working..?).

The units are identical to HA and it feels a bit redundant to have them in both ZHA and zigpy. An alternative that was also discussed was a "HA API" package that we can share between ZHA/zigpy/quirks, but we likely want to avoid having another package.


prairiesnpr (approved):
Looks good, I much prefer this approach.

PR #354, opened 2025-01-19
Separate polling attributes in electrical measurement cluster

abmantis:
As suggested in https://github.com/zigpy/zha/pull/339#issuecomment-2599403130 this adds
an `ZCL_POLLING_ATTRS` to define attributes that should be polled
separatelly from the ones that can have reporting config.

This also moves some attributes that do no support reporting config out
of `REPORT_CONFIG`.


TheJulianJES (commented):
Let's change `ZCL_POLLING_ATTRS` to just be a list of the attribute names, similar to `ZCL_INIT_ATTRS`. (`async_update` needs to be adjusted too then.)

abmantis:
> Let's change `ZCL_POLLING_ATTRS` to just be a list of the attribute names, similar to `ZCL_INIT_ATTRS`. (`async_update` needs to be adjusted too then.)

Yeah, that was a stupid Sunday night copy/paste I did there :upside_down_face:

zha/zigbee/cluster_handlers/homeautomation.py
```diff
@@ -85,35 +80,30 @@ class MeasurementType(enum.IntFlag):
...
         AttrReportConfig(
             attr=ElectricalMeasurement.AttributeDefs.ac_frequency.name,
             config=REPORT_CONFIG_OP,
         ),
-        AttrReportConfig(
-            attr=ElectricalMeasurement.AttributeDefs.ac_frequency_max.name,
-            config=REPORT_CONFIG_DEFAULT,
-        ),
     )
+    ZCL_POLLING_ATTRS = {
```
TheJulianJES: Actually, we might want `ZCL_POLLING_ATTRS` to be a `list` instead of a `set`, so that the attributes are always polled in the same (expected) order. So, just `ZCL_POLLING_ATTRS = [...]`.

Since we don't poll the divider/multiplier here at the moment, this might not matter, but I think it would be better anyways and keep the same order from before when a (sorted) dict was used.

(Since we're not searching for anything in `ZCL_POLLING_ATTRS`, a list should be fine performance-wise. Might even save a bit on memory if we use a list. Though it's not like that really matters here..)

TheJulianJES (approved):
Thanks. This looks good.

PR #355, opened 2025-01-22
Provide effects on supported Philips Hue lights

BetaRavener:
## Description
Adds additional effects to supported Philips Hue lights. To control the effects, commands are sent to a exposed cluster.
Implements HA / UI portion of https://github.com/zigpy/zha-device-handlers/issues/2517, which contains detailed discussion.

The effect is turned on by command 0x22 that accepts brightness and the effect ID.
To turn off, command 0x20 is used that will preserve most recent color and brightness while stopping further effect execution.

## Testing
Tested with Philips Hue Go. See video below:
https://github.com/user-attachments/assets/9d6236f6-4d8f-40ee-8525-82e674d4ca85

## Prerequisite
PR implementing quirk for the Philips cluster: https://github.com/zigpy/zha-device-handlers/issues/2517

### DIFF
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -72,6 +72,7 @@ from zha.zigbee.cluster_handlers.const import (
     CLUSTER_HANDLER_LEVEL,
     CLUSTER_HANDLER_LEVEL_CHANGED,
     CLUSTER_HANDLER_ON_OFF,
+    CLUSTER_HANDLER_PHILIPS_HUE_LIGHT,
 )
 from zha.zigbee.cluster_handlers.general import LevelChangeEvent
 
@@ -996,7 +997,7 @@ class HueLight(Light):
     aux_cluster_handlers={
         CLUSTER_HANDLER_COLOR,
         CLUSTER_HANDLER_LEVEL,
-        "philips_hue_cluster",
+        CLUSTER_HANDLER_PHILIPS_HUE_LIGHT,
     },
     manufacturers={"Philips", "Signify Netherlands B.V."},
 )
@@ -1018,14 +1019,16 @@ class HueEffectLight(HueLight):
     ) -> None:
         """Initialize the ZHA light."""
         super().__init__(unique_id, cluster_handlers, endpoint, device, **kwargs)
-        self._hue_cluster = self.cluster_handlers.get("philips_hue_cluster")
+        self._hue_light_cluster = self.cluster_handlers.get(
+            CLUSTER_HANDLER_PHILIPS_HUE_LIGHT
+        )
         self._effect_list.extend(self.HUE_EFFECTS.keys())
 
     async def async_turn_on(self, **kwargs: Any) -> None:
         """Turn the entity on."""
-        # If only change of brightness is requested, the effect doesn't have to be interupted
+        # If only change of brightness is requested, the effect doesn't have to be interrupted
         if kwargs.get(ATTR_BRIGHTNESS) is not None and all(
-            attr == ATTR_BRIGHTNESS or kwargs.get(attr) is None for attr in kwargs
+            kwargs.get(attr) is None for attr in (kwargs.keys() - {ATTR_BRIGHTNESS})
         ):
             effect = self._effect
         else:
@@ -1038,7 +1041,7 @@ class HueEffectLight(HueLight):
 
         if effect in self.HUE_EFFECTS:
             effect_id = self.HUE_EFFECTS[effect]
-            await self._hue_cluster.multicolor(
+            await self._hue_light_cluster.multicolor(
                 data=bytearray([0x22, 0x00, self._brightness, effect_id])
             )
             self._effect = effect
@@ -1047,7 +1050,9 @@ class HueEffectLight(HueLight):
         ) and self._effect in self.HUE_EFFECTS:
             # Only stop effect if it was started by us
             # Following command will stop the effect while preserving brightness
-            await self._hue_cluster.multicolor(data=bytearray([0x20, 0x00, 0x00, 0x00]))
+            await self._hue_light_cluster.multicolor(
+                data=bytearray([0x20, 0x00, 0x00, 0x00])
+            )
             self._effect = EFFECT_OFF
         else:
             # Don't react on unknown effects, for example 'colorloop'
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -993,6 +993,69 @@ class HueLight(Light):
 
 @STRICT_MATCH(
     cluster_handler_names=CLUSTER_HANDLER_ON_OFF,
+    aux_cluster_handlers={
+        CLUSTER_HANDLER_COLOR,
+        CLUSTER_HANDLER_LEVEL,
+        "philips_hue_cluster",
+    },
+    manufacturers={"Philips", "Signify Netherlands B.V."},
+)
+class HueEffectLight(HueLight):
+    # Supported effects and their ID used in commands
+    HUE_EFFECTS = {"candle": 1, "fireplace": 2, "prism": 3}
+
+    """Representation of a HUE light with effects."""
+
+    def __init__(
+        self,
+        unique_id: str,
+        cluster_handlers: list[ClusterHandler],
+        endpoint: Endpoint,
+        device: Device,
+        **kwargs,
+    ) -> None:
+        """Initialize the ZHA light."""
+        super().__init__(unique_id, cluster_handlers, endpoint, device, **kwargs)
+        self._hue_cluster = self.cluster_handlers.get("philips_hue_cluster")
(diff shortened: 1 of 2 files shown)


BetaRavener:
Test asserts are failing, but seem to be correct - tests will need to be changed. All lights that provide cluster 0xFC03 are failing, while those missing it (e.g. `signify-netherlands-b-v-rwl022.json`) are not.

TheJulianJES:
You can delete the JSON test files for now then. We'll eventually regenerate and upload them again.
In the future, we might have a tool to re-generate these tests automatically (without needing to grab them from a running instance).

BetaRavener:
@TheJulianJES I pushed changes to match the last renaming from quirk PR. 

I haven't resolved the failing tests yet as I lost track which JSONs were problematic, but I realized I also don't know why they were failing in the first place. Given that I had old cluster name, even if the quirk PR got already picked up during tests, my class shouldn't have matched any device yet. 

I've added `CLUSTER_HANDLER_PHILIPS_HUE_LIGHT` to `aux_cluster_handlers` to give `HueEffectLight` just small extra weight over `HueLight` class thinking that `strict_match` would require the new cluster to be present. But looking at `strict_matched`, is it true the aux clusters are not considered for matching? Should I then move the `CLUSTER_HANDLER_PHILIPS_HUE_LIGHT` to `cluster_handler_names` instead?

jtbandes:
Hi @TheJulianJES, just wondering if you were able to provide any feedback on the above?

jtbandes:
@puddly hi there, just a friendly ping - there are several folks in the HA community who would love to see this merged (see https://github.com/zigpy/zha-device-handlers/issues/2517)  Could you help us figure out what the next steps are?

TheJulianJES:
The "light turn on" call in the class does a lot. I think the current approach for the Hue class might be fine, but I hope to have a look soon to see if it can/should be done differently. Right now, we have to make sure the effect on/off behavior exactly matches the one in the base light class. That might not be ideal.

One other thing I can think of right now: Can we somehow poll the current Hue effect in `async_update`?
Otherwise, I'll try to play around with this soon-ish.

jtbandes:
@BetaRavener let us know if you are able to continue driving this forward, otherwise I am sure someone else can try to pick it up :)

Orange-GT3:
Just giving this a nudge, @BetaRavener or at @TheJulianJES.

BetaRavener:
I could finish this PR but unfortunately I still don't have an answer regarding the clusters, if I should keep it in aux list or move to the "main" list as I mentioned above.

The rest from Julian sounded like possible but non-blocking improvements.

The single question about polling current effect - I honestly don't know. I only found people successfully snooping and reverse engineering the set effect command, I didn't come across one to query current effect. It would also have to be part of the manufacturer specific cluster if it exists.

Orange-GT3:
> I could finish this PR but unfortunately I still don't have an answer regarding the clusters, if I should keep it in aux list or move to the "main" list as I mentioned above.
> 
> The rest from Julian sounded like possible but non-blocking improvements.
> 
> The single question about polling current effect - I honestly don't know. I only found people successfully snooping and reverse engineering the set effect command, I didn't come across one to query current effect. It would also have to be part of the manufacturer specific cluster if it exists.

Possibly a redundant question, but have you seen this: [Zigbee format for Philips Hue manufacturer-specific light updates](https://github.com/chrivers/bifrost/blob/master/doc/hue-zigbee-format.md)?

BetaRavener:
Yes, I think it's also mentioned on the original issue. But this doesn't allow query as I said, let's say somebody uses different app or buttons on the light to change the effect, the HA will get out of sync because we can't get the new state.

Also, the more advanced configuration features are not helpful as there's no interface to build these up and include in effect list. So if you want more reddish, slower, candle effect, you'll have to still send command to cluster yourself. This is aimed at non-technical users who (like me earlier) just click around, and then having an effect readibly available in dropdown might be already good enough for them.

Orange-GT3:
OK, I just thought I'd check.

I understand what you are trying to achieve with your interface and fully support it. Other than testing, though, I can't contribute to coding as it's all totally beyond me.

I'd love to see a fully fledged Philips Hue Integration but I realise that would be an absolute tonne of work but maybe if parts get done here and there, it would be possible for the HA team to pull them together into a single integration.

Orange-GT3:
> But this doesn't allow query as I said, let's say somebody uses different app or buttons on the light to change the effect, the HA will get out of sync because we can't get the new state.

Correct me if I'm wrong, but it's possible to read cluster values so couldn't the state be checked by reading the cluster and then decoding the return value according the Frame Format detailed at the above link?

BetaRavener:
I'm bit out of the loop but yes you're right you can read values from cluster. What we are missing (I think) is from where, i.e. the address of the attribute in cluster that would provide us this value is unknown.

I don't think there's a return value other than command success / failure.

jtbandes:
@TheJulianJES could you help answer [this question from above](https://github.com/zigpy/zha/pull/355#issuecomment-2654340237)?

> I've added `CLUSTER_HANDLER_PHILIPS_HUE_LIGHT` to `aux_cluster_handlers` to give `HueEffectLight` just small extra weight over `HueLight` class thinking that `strict_match` would require the new cluster to be present. But looking at `strict_matched`, is it true the aux clusters are not considered for matching? Should I then move the `CLUSTER_HANDLER_PHILIPS_HUE_LIGHT` to `cluster_handler_names` instead?

Arney:
Just wondering, would this PR allow me to turn on the candle effect on a Hue Go from the UI?

Orange-GT3:
> Just wondering, would this PR allow me to turn on the candle effect on a Hue Go from the UI?

Maybe, yes, or it might be another PR that @BetaRavener was working on. I forget now.

BetaRavener:
Yes, @Arney, tbis PR implements the UI portion and allows switching effects through dropbox. You can see this in the linked video from testing on the original PR post. 

However @TheJulianJES [said on other PR here](https://github.com/zigpy/zha-device-handlers/pull/3664#issuecomment-3591649981) that this PR is basically blocked by some other, unknown work that needs to be done in HA first. 

I understand that this approach might be bit rough and has limitations, but it would give people easy access to feature their lights support *now*,  rather than some unknown time in future. Ubfortunatelly, I don't see this PR getting merged anytime soon if ever as it didn't really get any attention in last 11 months..

Orange-GT3:
> Yes, @Arney, tbis PR implements the UI portion and allows switching effects through dropbox. You can see this in the linked video from testing on the original PR post.
> 
> However @TheJulianJES [said on other PR here](https://github.com/zigpy/zha-device-handlers/pull/3664#issuecomment-3591649981) that this PR is basically blocked by some other, unknown work that needs to be done in HA first.
> 
> I understand that this approach might be bit rough and has limitations, but it would give people easy access to feature their lights support _now_, rather than some unknown time in future. Ubfortunatelly, I don't see this PR getting merged anytime soon if ever as it didn't really get any attention in last 11 months..

@BetaRavener - I'm running HAOS on a Raspberry Pi, is there any way that I can hack your code into my installation to do some testing with some more devices?

I have a Festavia Light String, a Solo Lightstrip and a Lightstrp Plus as well as Akari downlights and Discover floodlights.

BetaRavener:
HAOS should be bit simpler - I have a docker container to which you first have to "log in" via docker exec, and then make sure the changes persist. HAOS being in the host filesystem, you should be able to just find the files I have edited in this PR and merge in the content. Just maybe back up the original file before to be able to easily revert once done with testing. And restart HA after every change, it needs to reload the Zigbee integration.

Orange-GT3:
> HAOS should be bit simpler - I have a docker container to which you first have to "log in" via docker exec, and then make sure the changes persist. HAOS being in the host filesystem, you should be able to just find the files I have edited in this PR and merge in the content. Just maybe back up the original file before to be able to easily revert once done with testing. And restart HA after every change, it needs to reload the Zigbee integration.

It turns out these files are buried within a Docker container on an HAOS install too. Getting into that and editing anything is beyond my skillset :(.

I'd love to be able to test this on my devices though.

Orange-GT3:
@TheJulianJES - please can this PR be merged as an interim until the other work (whatever that is) happens? I think it would make a great many Hue users very happy.

Orange-GT3:
@BetaRavener - is your home brew version still allowing you to set effects? ZHAToolkit won't even work for me any more and all the endpoints relevant seem to have disappeared from my lights.

Orange-GT3:
@TheJulianJES - any chance of some movement in this area please?

Using ZHAToolkit to start effects is no longer working for some unknown reason.

puddly:
@Orange-GT3 We're working on an API to make these types of things possible in the next couple of releases: https://github.com/zigpy/zha/pull/762

Orange-GT3:
> @Orange-GT3 We're working on an API to make these types of things possible in the next couple of releases: #762

Thanks @puddly. I've read that and the various linked PRs/epics/etc and the future for ZHA looks great!

PR #358, opened 2025-01-24
Fix issues with `read_attributes` and `attribute_updated`

TheJulianJES:
## Proposed change

This PR converts the `get_attributes` partialmethod to an actual coroutine that calls `_get_attributes` with `raise_exceptions=False` (fix: https://github.com/zigpy/zha/commit/4fda764b7686db5ebc6b98c5dd5a944637af0388).

This is needed, as `partialmethod` doesn't correctly work with coroutines nor with mypy type checking.
~~This was seen in [tests](https://github.com/zigpy/zha/actions/runs/12940028032/job/36093455183): "Warning: coroutine 'Cluster.read_attributes' was never awaited"~~
EDIT: The test warnings might be unrelated and still seem to happen. This should be looked into.

After doing this, mypy was now able to correctly do type checking and threw some other errors:
- the switch platform tried to poll a `None` attribute (`_inverter_attribute_name` is `None` by default) (fix: https://github.com/zigpy/zha/commit/2baa937b6117f3d53fcbbc1d3a234740ad27b91a)
- the cover platform called `attribute_updated` with some types being (potentially) incompatible (fix in: https://github.com/zigpy/zha/commit/c042eec3983f7fafc7a3dadaf434501e712a62c2)

This led me to look into `attribute_updated`. This is called by zigpy in `_update_attribute`.
So, when we read/poll an attribute, zigpy automatically calls that method. For some reason, we also called it in ZHA:

### Incorrectly calling `attribute_updated`

We also called that method incorrectly with the "attribute name" being provided as the "value" and the actual "value" being provided as the "timestamp".
It seems like this never manifested into a real issue, because we never actually access the event data from `CLUSTER_HANDLER_ATTRIBUTE_UPDATED` that's fired in `ClusterHandler.attribute_updated`, instead relying on directly reading zigpy attribute cache in the various platforms.

This is why https://github.com/zigpy/zha/commit/c042eec3983f7fafc7a3dadaf434501e712a62c2 allows us to remove manually calling `attribute_updated` completely, zigpy already does it for us.

### Cover platform

The cover platform also tried to do some inversion in `WindowCoveringClusterHandler.async_update`, but due to us never actually accessing the inverted value from the event in the [cover platform here](https://github.com/zigpy/zha/blob/7292a0cc638d294438a8e75080af95f595d32e2a/zha/application/platforms/cover/__init__.py#L252-L271) (that was incorrectly provided as the `timestamp` value), those values were never used anywhere. That's why we can completely remove that code too.

The values are [properly inverted in the cluster handler properties](https://github.com/zigpy/zha/blob/7292a0cc638d294438a8e75080af95f595d32e2a/zha/zigbee/cluster_handlers/closures.py#L213-L233). This is what's actually used.

## Additional information

If needed, this PR can also be split up a bit. We can remove the `attribute_updated` calls in a first PR and only then replace the `get_attributes` partialmethod. Doing this the other way around would cause some mypy errors.

This PR wasn't tested manually yet, but tests still pass. Still, we should probably try to see if plugs are still being polled correctly and that covers didn't break. I doubt they did though.

### DIFF
--- a/zha/zigbee/cluster_handlers/smartenergy.py
+++ b/zha/zigbee/cluster_handlers/smartenergy.py
@@ -320,14 +320,7 @@ class MeteringClusterHandler(ClusterHandler):
             for a in self.REPORT_CONFIG
             if a["attr"] not in self.cluster.unsupported_attributes
         ]
-        result = await self.get_attributes(attrs, from_cache=False, only_cache=False)
-        if result:
-            for attr, value in result.items():
-                self.attribute_updated(
-                    self.cluster.find_attribute(attr).id,
-                    attr,
-                    value,
-                )
+        await self.get_attributes(attrs, from_cache=False, only_cache=False)
 
     @staticmethod
     def get_formatting(formatting: int) -> str:
--- a/zha/application/platforms/switch.py
+++ b/zha/application/platforms/switch.py
@@ -322,13 +322,13 @@ class ConfigurableAttributeSwitch(PlatformEntity):
     async def async_update(self) -> None:
         """Attempt to retrieve the state of the entity."""
         self.debug("Polling current state")
+
+        polling_attrs = [self._attribute_name]
+        if self._inverter_attribute_name:
+            polling_attrs.append(self._inverter_attribute_name)
+
         results = await self._cluster_handler.get_attributes(
-            [
-                self._attribute_name,
-                self._inverter_attribute_name,
-            ],
-            from_cache=False,
-            only_cache=False,
+            polling_attrs, from_cache=False, only_cache=False
         )
 
         self.debug("read values=%s", results)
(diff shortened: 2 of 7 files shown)


zha/zigbee/cluster_handlers/smartenergy.py
```diff
@@ -320,14 +320,7 @@ async def async_update(self) -> None:
...
         ]
-        result = await self.get_attributes(attrs, from_cache=False, only_cache=False)
-        if result:
-            for attr, value in result.items():
-                self.attribute_updated(
-                    self.cluster.find_attribute(attr).id,
-                    attr,
-                    value,
-                )
+        await self.get_attributes(attrs, from_cache=False, only_cache=False)
```
TheJulianJES, line 323: Coverage for this (SmartEnergy Metering sensor polling) added here:
- https://github.com/zigpy/zha/pull/359

puddly (approved):

PR #359, opened 2025-01-24
Add test checking SmartEnergy Metering polling

TheJulianJES:
## Proposed change

This adds test coverage for polling (`async_update`) in the SmartEnergy Metering cluster handler, similar to how it's done for the ElectricalMeasurement cluster.

There are two separate variants possible for this test:
1. basically copying the entire EM sensor test and changing it for Metering: https://github.com/zigpy/zha/commit/c2dca75041183a6345b298ebe6c44fccfa3939e0
2. using `parametrize` to combine both tests: [current diff of this PR](https://github.com/zigpy/zha/pull/359/files) (diff to first commit: https://github.com/zigpy/zha/commit/0108f9580ebdad5c0824c7a02ec6dceab24f2d4d)

Let me know what you prefer. The first variant with the two separate tests might be a bit easier to understand(?)

## Additional information

This was noticed whilst working on https://github.com/zigpy/zha/pull/358.

### DIFF
--- a/tests/test_sensor.py
+++ b/tests/test_sensor.py
@@ -78,6 +78,37 @@ def elec_measurement_zigpy_device_mock(
     return zigpy_device
 
 
+def metering_zigpy_device_mock(
+    zha_gateway: Gateway,
+) -> ZigpyDevice:
+    """Metering zigpy device."""
+
+    zigpy_device = create_mock_zigpy_device(
+        zha_gateway,
+        {
+            1: {
+                SIG_EP_INPUT: [
+                    general.Basic.cluster_id,
+                    smartenergy.Metering.cluster_id,
+                ],
+                SIG_EP_OUTPUT: [],
+                SIG_EP_TYPE: zigpy.profiles.zha.DeviceType.SIMPLE_SENSOR,
+                SIG_EP_PROFILE: zigpy.profiles.zha.PROFILE_ID,
+            }
+        },
+        model="TS011F",  # polling is enabled for this model
+    )
+    zigpy_device.node_desc.mac_capability_flags |= 0b_0000_0100
+    zigpy_device.endpoints[1].smartenergy_metering.PLUGGED_ATTR_READS = {
+        "unit_of_measure": 0,  # kWh
+        "multiplier": 1,
+        "divisor": 1000,
+        "summation_formatting": 0b0_0100_011,  # read from plug
+        "metering_device_type": 0,  # electric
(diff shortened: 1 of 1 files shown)


puddly:
I think two separate tests is easier to read.

Hedda:
Does this mean that devices that only use the Zigbee Smart Energy (Zigbee SE) profile and standard protocol could be supported?

And if so, which versions of Zigbee Smart Energy can be supported? ...FYI; CSA very recently ratified Zigbee Smart Energy 1.4a

https://csa-iot.org/newsroom/smart-energy-1-4a-for-zigbee-based-energy-management-technology/

https://zigbeealliance.org/wp-content/uploads/2021/10/docs-07-5356-21-zse-zigbee-smart-energy-profile-specification.pdf

https://zigbeealliance.org/wp-content/uploads/2019/11/docs-07-5356-19-0zse-zigbee-smart-energy-profile-specification.pdf

Asking since ZHA integration documentation state under limitations that the ZSE (“Zigbee Smart Energy”) profile is not supported (but unclear if that limitation is in the zha or zigpy libraries and the type of Zigbee networks they supports or if the limitation is in the Zigbee Coordinator and its Zigbee stack):

https://www.home-assistant.io/integrations/zha#limitations

_"ZHA does not currently support devices that can only use the ZSE (“Zigbee Smart Energy”) profile, that is however due to the “Zigbee SE” specification not being part of the standard Zigbee 3.0 specification and thus not implemented in most of the Zigbee protocol stacks that are commonly available Zigbee Coordinator radio adapters and modules."_

From my understanding it sounds as if you may need a dedicated Zigbee SmartEnergy network for such devices though?

And on top of that many Zigbee Smart Energy devices uses Sub-1GHz frequency instead so could need a Zigbee Coordinator with multiprotocol and multi-band support, such as the CC1352 (CC1352P or CC1352P) from Texas Instruments, or alternativly the nRF52832 from Nordic Semiconductor (which uses the ZBOSS stack). 

Anyway, the point is that both Zigbee Smart Energy and Sub-1GHz are missing links in zha and zigpy-based solutions.

PS: I believe that Develco makes some Zigbee Smart Energy (Zigbee SE) meter products that uses Sub-1GHz and based on Nordic Semi's nRF52 SoCs?

https://www.develcoproducts.com/software/edge-services/wireless-protocols/zigbee/zigbee-smart-energy/

https://www.develcoproducts.com/products/

Issue #305, opened 2024-11-25
Allow quirks v2 sensors/platforms to provide a lambda for parsing state


TheJulianJES:
We should consider providing a way that lets v2 quirks provide a lambda to allow ZHA to parse ZCL attribute state to a different HA state with the parsing being in the v2 quirk.

This should be implemented for at least binary sensors and normal/analog sensors.

We may also want to consider the other platforms.
Switch currently has a [special `force_inverted`, `invert_attribute_name` implementation](https://github.com/zigpy/zha/blob/d7e6544118c01629a01a2dba55fae6e18c5cd1c0/zha/application/platforms/switch.py#L273-L290).

TheJulianJES:
Some helpers in zha-quirks tests would also be needed, so we can actually test the "attribute converters" in zha-quirks.

puddly:
This may help clean up all of the `_attribute_updated` overrides too. These end up in the zigpy attribute cache and are persisted to disk in a modified form.

TheJulianJES:
This is implemented for binary sensors and (analog) sensors with https://github.com/zigpy/zha/pull/360 and https://github.com/zigpy/zigpy/pull/1540.
Closing this for now. If this is also needed for other platforms, please create a new isuse.

PR #360, opened 2025-01-26
Implement quirks v2 `attribute_converter`

TheJulianJES:
## Proposed change

This adds an `attribute_converter` to the following quirk v2 platforms:
- binary sensor
- sensor

This will allow us to parse a ZCL attribute value in quirks, before passing it to the HA state.
This is mostly needed for binary sensors.

## Additional information

Corresponding zigpy PR:
- https://github.com/zigpy/zigpy/pull/1540

Closes:
- https://github.com/zigpy/zha/issues/305

### DIFF
--- a/tests/test_binary_sensor.py
+++ b/tests/test_binary_sensor.py
@@ -26,7 +26,12 @@ from tests.common import (
 from zha.application import Platform
 from zha.application.gateway import Gateway
 from zha.application.platforms import PlatformEntity
-from zha.application.platforms.binary_sensor import Accelerometer, IASZone, Occupancy
+from zha.application.platforms.binary_sensor import (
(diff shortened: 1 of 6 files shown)


TheJulianJES:
This should be ready to review too now. No official quirks will use it with HA 2025.2.0 yet, but it'll allow development for new ones to be easier. 

There are some changes in `test_switch` that I can move to another PR if wanted. To follow patterns in other tests, it's changed to use a test-specific `DeviceRegistry` instead of the global one. 
**EDIT: Removed them** from this PR now. They're not related anymore, since the `attriubte_converter` for switches was removed. We should still make the changes in another PR though (and likely also to some other tests).

Issue #303, opened 2024-11-25
Add additional unique ID suffix to v2 quirks


TheJulianJES:
We should provide an optional kwarg for a unique ID "suffix suffix" to this:
https://github.com/zigpy/zha/blob/d7e6544118c01629a01a2dba55fae6e18c5cd1c0/zha/application/platforms/__init__.py#L353-L356

Currently, v2 quirks can only add one entity (per platform) per attribute id. Otherwise, they silently fail to create at runtime due to duplicate IDs. This is an issue with buttons (writing different values) or (binary) sensors (parsing values differently).

We should validate quirks v2 metadata (in zha-quirks, zigpy, or zha?) to make sure there's an explicit warning about this.
If this happens, the conflicting quirk v2 entities should have a `uid_suffix` kwarg added. This would be an additional suffix to the attribute UID suffix already in ZHA.

TheJulianJES:
This was implemented with https://github.com/zigpy/zha/pull/362 and https://github.com/zigpy/zigpy/pull/1541.
Note that this implementation will replace the existing/default UID suffix of the attribute/command name.

PR #362, opened 2025-01-26
Implement quirks v2 `unique_id_suffix`

TheJulianJES:
## Proposed change

This implements the quirks v2 `unique_id_suffix` that's appended to the existing `unique_id` of IEEE+endpoint+cluster (or something like that). So, it just replaces the default ZHA `unique_id_suffix` (which is the attribute name for v2 quirk entities by default)

This is needed when there are multiple quirks v2 entities for a specific attribute.
A test is also added that checks this for "write attr buttons" (writing the same attribute with different values).

## Additional information

Corresponding zigpy PR:
- https://github.com/zigpy/zigpy/pull/1541

Closes:
- https://github.com/zigpy/zha/issues/303

### DIFF
--- a/zha/application/platforms/__init__.py
+++ b/zha/application/platforms/__init__.py
@@ -350,15 +350,12 @@ class PlatformEntity(BaseEntity):
         if entity_metadata.translation_key:
             self._attr_translation_key = entity_metadata.translation_key
 
-        if has_attribute_name:
+        if unique_id_suffix := entity_metadata.unique_id_suffix:
+            self._unique_id_suffix = unique_id_suffix
+        elif has_attribute_name:
             self._unique_id_suffix = entity_metadata.attribute_name
         elif has_command_name:
             self._unique_id_suffix = entity_metadata.command_name
-
-        if entity_metadata.unique_id_suffix:
-            self._unique_id_suffix = (
-                f"{self._unique_id_suffix}-{entity_metadata.unique_id_suffix}"
-            )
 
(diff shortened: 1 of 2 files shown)


zha/application/platforms/__init__.py
```diff
@@ -355,6 +355,11 @@ def _init_from_quirks_metadata(self, entity_metadata: EntityMetadata) -> None:
         elif has_command_name:
             self._unique_id_suffix = entity_metadata.command_name
 
+        if entity_metadata.unique_id_suffix:
+            self._unique_id_suffix = (
+                f"{self._unique_id_suffix}-{entity_metadata.unique_id_suffix}"
+            )
```
TheJulianJES: You mean that instead of appending the quirks v2 provided `unique_id_suffix` to the existing `unique_id_suffix` (created in the lines above), that we just replace it?
I guess that could work. The "attribute_name + extra suffix" is likely a bit unnecessary.

It would still be a suffix (to the original UID of IEEE +endpoint+cluster), so I'd keep the name the same probably and just change the implementation here. I don't think we should or even need to allow changing the whole UID (or the first part).

Changing just this suffix part to not append to the attr name should allow quirks v2 to take over ZHA/"v1" entities.

TheJulianJES:
Changed the implementation to override the default attribute name `unique_id_suffix` for quirks v2 entities now, instead of appending it to the default.
This should allow us to replace ZHA/v1 entities from v2 quirks.

(Tests will fail until zigpy PR is merged + bump)

puddly (approved):

PR #254, opened 2024-10-21
Support configuration of entity reporting for V2 quirks

benbancroft:
# Proposed change

This change hooks up the V2 Quirk `report_config` option to the cluster handler created for V2 quirks. It also 'claims' the cluster, which is required for reporting and binding hooks to work upon configuration of the cluster and previously wasn't being done for V2 Quirk clusters.

Screenshot 2024-10-21 111956

# Additional information

Please refer to https://github.com/zigpy/zigpy/pull/1490 for more information on the motivation behind this patch. This patch will need to be merged after that has been merged and zigpy dependency has been bumped.

PR #363, opened 2025-01-27
Implement quirks v2 reporting config, optimize discovery

TheJulianJES:
## Proposed change

This implements quirks v2 `reporting_config` in ZHA. The quirks v2 processing for `ZCL_INIT_ATTRS` is also optimized to not copy `ZCL_INIT_ATTRS` for every entity metadata object.

A test is also added that actually checks the existing `ZCL_INIT_ATTRS` code and new code: 
- that it doesn't modify the original `ZCL_INIT_ATTRS` class variable, affecting other cluster handler instances
- that `ZCL_INIT_ATTRS` are correctly set (and doesn't contain attributes also in `REPORTING_CONFIG`)
- that `REPORT_CONFIG` is correctly set
- that the cluster handler is claimed when `reporting_config` is set, thus binding it during ZHA device configuration

## Additional information

Supersedes:
- https://github.com/zigpy/zha/pull/254

### DIFF
--- a/zha/application/discovery.py
+++ b/zha/application/discovery.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from collections import Counter
+from dataclasses import astuple
 import logging
 from typing import TYPE_CHECKING, cast
 
@@ -51,6 +52,7 @@ from zha.application.registries import (
 
 # importing cluster handlers updates registries
 from zha.zigbee.cluster_handlers import (  # noqa: F401 pylint: disable=unused-import
+    AttrReportConfig,
     ClusterHandler,
     closures,
     general,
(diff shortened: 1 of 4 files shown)


zha/zigbee/cluster_handlers/manufacturerspecific.py
```diff
@@ -120,7 +120,7 @@ def __init__(self, cluster: zigpy.zcl.Cluster, endpoint: Endpoint) -> None:
 class OppleRemoteClusterHandler(ClusterHandler):
     """Opple cluster handler."""
 
-    REPORT_CONFIG = ()
+    REPORT_CONFIG: tuple[AttrReportConfig, ...] = ()
```
TheJulianJES, line 123: In a future PR, we can remove empty `REPORT_CONFIG`s for all `ClusterHandler`s, as this is already the default.
Otherwise, we can just specify the type everywhere. It was only needed here for now, as mypy otherwise complained about an impossible equality check in the added test.

zha/application/discovery.py
```diff
@@ -298,18 +300,37 @@ def discover_quirks_v2_entities(self, device: Device) -> None:
...
+                        # copy existing ZCL_INIT_ATTRS into instance variable once,
+                        # so we don't modify other instances of the same cluster handler
+                        if zha_const.ZCL_INIT_ATTRS not in cluster_handler.__dict__:
+                            cluster_handler.ZCL_INIT_ATTRS = (
+                                cluster_handler.ZCL_INIT_ATTRS.copy()
+                            )
+                        # add the attribute to the guaranteed instance variable
+                        cluster_handler.ZCL_INIT_ATTRS[attr_name] = (
+                            entity_metadata.attribute_initialized_from_cache
+                        )
```
TheJulianJES, line 333: If there are too many comments here, we can remove some:
```suggestion
                if attr_name := getattr(entity_metadata, "attribute_name", None):
                    if rep_conf := getattr(entity_metadata, "reporting_config", None):
                        # if attr is already in REPORT_CONFIG, remove to override
                        cluster_handler.REPORT_CONFIG = tuple(
                            filter(
                                lambda cfg: cfg["attr"] != attr_name,
                                cluster_handler.REPORT_CONFIG,
                            )
                        )
                        cluster_handler.REPORT_CONFIG += (
                            AttrReportConfig(attr=attr_name, config=astuple(rep_conf)),
                        )
                        # claim the cluster handler, so ZHA configures and binds it
                        endpoint.claim_cluster_handlers([cluster_handler])

                    elif attr_name not in cluster_handler.ZCL_INIT_ATTRS:
                        # copy existing ZCL_INIT_ATTRS into instance variable once,
                        # so we don't modify other instances of the same cluster handler
                        if zha_const.ZCL_INIT_ATTRS not in cluster_handler.__dict__:
                            cluster_handler.ZCL_INIT_ATTRS = (
                                cluster_handler.ZCL_INIT_ATTRS.copy()
                            )
                        cluster_handler.ZCL_INIT_ATTRS[attr_name] = (
                            entity_metadata.attribute_initialized_from_cache
                        )
```

TheJulianJES:
This PR is ready IMO. It's also well-tested to verify that the existing `ZCL_INIT_ATTRS` ch class variable isn't screwed up.
If there are too many comments, we can [remove some](https://github.com/zigpy/zha/pull/363#discussion_r1931191867). Others are crucial though if you're looking at this for the first time.

PR #365, opened 2025-01-28
[RFC] Explicit entity object discovery, lifetimes, and registration

puddly:
`PlatformEntity` objects, among others, rely on implicit "registration" inside of `__init__` by mutating a global dictionary that ZHA flushes when creating new devices. This is a bit difficult to work with, in my opinion, and is somewhat preventing https://github.com/zigpy/zigpy/pull/1535 from being implemented.

This PR:
1. Construct entity objects *early*.
2. Implement an `on_add` interface to mirror `on_remove` and allow for entity object creation to not rely on any sort of global state.
3. Allow entities to explicitly indicate if they are not supported by the current device via `Entity.is_supported` (this is mostly a copy/paste of the `create_platform_entity` logic).
4. Funnel all device entity discovery explicitly via `gateway.py`.

### DIFF
--- a/zha/application/gateway.py
+++ b/zha/application/gateway.py
@@ -242,10 +242,6 @@ class Gateway(AsyncUtilMixin, EventBase):
 
     async def _async_initialize(self) -> None:
         """Initialize controller and connect radio."""
-        discovery.DEVICE_PROBE.initialize(self)
-        discovery.ENDPOINT_PROBE.initialize(self)
-        discovery.GROUP_PROBE.initialize(self)
-
         self.shutting_down = False
 
         app_controller_cls, app_config = self.get_application_controller_data()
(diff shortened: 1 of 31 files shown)


puddly:
This PR is "done" architecturally, for the most part.

Main changes:
- **All entities must compute capability bitmasks and other feature flags in `Entity.recompute_capabilities()`.**
   This is required because we effectively perform multiple passes when setting up entities: an initial pass to discover entities and claim cluster handlers, a second pass to initialize the cluster handlers and possibly populate the attribute cache, and a third pass to actually validate whether or not the discovered entities are applicable to the device. In the future, it would be nice to consolidate this functionality entirely into the `Entity` subclass.
- **Entities indicate device applicability via `Entity._is_supported()`.**
   For the same reason as above, since entity objects are created at runtime, we need a way to signal applicability to a device independent of creation. I've also added a new `_attr_is_always_supported` to allow entities that do not abide by these rules to be easily flagged (quirks v2, group, RSSI/LQI, and device counters).
- **All entity objects' listeners, callbacks, and tasks must be registered in `Entity.on_add()` and deregistered with `Entity.on_remove()`.**
   For simplicity, there is also a `Entity._on_remove_callbacks` list that will be iterated over by the default `Entity.on_remove()` callback.

That about covers everything. This is a very, very rough first pass PR and is in not in a state for review beyond architecture changes.

zha/application/gateway.py
```diff
@@ -314,7 +310,7 @@ def load_devices(self) -> None:
                 delta_msg,
                 zha_device.consider_unavailable_time,
             )
-        self.create_platform_entities()
+            self._maybe_create_device_entities(zha_device)
```
dmulcahey: should we drop "device" from the name here? this is going to create all entities and we have the concept of "device specific" entities w/ the counters and such. I think the name could be a bit confusing.

PR #370, opened 2025-01-29
Add Aqara H1 with neutral features

krakonos1602:
This PR implements handling of Do not disturb mode, Decoupled mode, Power outage memory and Reversed LEDs switches present in Aqara H1 with neutral ("lumi.switch.n1aeu1", "lumi.switch.n2aeu1") wall switches.

### DIFF
--- a/zha/application/platforms/select.py
+++ b/zha/application/platforms/select.py
@@ -15,6 +15,7 @@ from zhaquirks.quirk_ids import (
     TUYA_PLUG_ONOFF,
 )
 from zhaquirks.xiaomi.aqara.magnet_ac01 import OppleCluster as MagnetAC01OppleCluster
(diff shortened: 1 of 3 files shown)


TheJulianJES (commented):
We likely wanna implement these as quirks v2 entities (in the zha-device-handlers repo).
The quirk might need to be updated from a v1 to a v2 quirk first.

krakonos1602:
> We likely wanna implement these as quirks v2 entities (in the zha-device-handlers repo). The quirk might need to be updated from a v1 to a v2 quirk first.

Sorry, I have no idea of v1 vs v2 quirks and entities. I just needed to make these switches work.

TheJulianJES:
Let's leave this open for now, so we can track it.
But you should already be able to change the values via the "Manage Zigbee device" -> Clusters menu, even without these changes.

PR #376, opened 2025-02-07
Improve cover entity handling

jeverley:
Aims to address the issue where a cover gets stuck in opening/closing state, also corrects some other bugs I've encountered whilst refactoring the code.

## Changes
 - Implements movement timeout and previous sate comparison to determine device instigated changes
   - The timeout is initially calculated based on expected travel distance because some devices only report new position after stopping
   - A device position update will reduce the timeout to the default (5 seconds)
   - Tilt has a lower dynamic timeout range due to shorter physical travel distance on the tilt axis
 - Removed target position entity attributes (we don't need to restore these)
 - An open/moving tilt state takes precedence over a static lift state
 - Utilise CoverState class in const instead of string literals for state (follows approach used in core)

## Fixes
 - Fix issue where the cover gets stuck in opening/closing state
   - This was because previous targets were never being cleared
   - Devices could also fail to reach the target position due to physical factors
 - Fix the _attr_device_class logic for Rollershade devices
   - This was resulting in the default icon _mdi:window-open_ rather than _mdi:roller-shade_ being used for the entity

## Addresses issues
- https://github.com/home-assistant/core/issues/98933
   - https://github.com/home-assistant/core/issues/137327
   - https://github.com/zigpy/zha-device-handlers/pull/3686#issuecomment-2578907234
   - https://github.com/home-assistant/core/issues/93836
   - https://github.com/zigpy/zha-device-handlers/issues/1397
- https://github.com/zigpy/zha/issues/378


jeverley:
I've finished making amendments to this.

puddly:
Thanks!

I've merged `dev` into this PR in a separate branch (https://github.com/zigpy/zha/compare/dev...puddly:zha:cover-platform-entity-status), since as of yesterday we now more explicitly recompute entity capabilities [in a dedicated method](https://github.com/puddly/zha/blob/007c33951294b36045b1075a12c52f5caf3bad87/zha/application/platforms/cover/__init__.py#L116-L138) and have [explicit lifecycle hooks](https://github.com/puddly/zha/blob/007c33951294b36045b1075a12c52f5caf3bad87/zha/application/platforms/cover/__init__.py#L140-L148) for attaching/detaching listeners.

I have a ThirdReality blind and am noticing nothing unusual when testing this PR out. The only thing is the cover state sometimes jittering "opening -> open -> opening -> open" in the last second, but this is still an improvement over the previous functionality where the cover sometimes got "stuck" if an attribute report was missed.

jeverley:
> I have a ThirdReality blind and am noticing nothing unusual when testing this PR out. The only thing is the cover state sometimes jittering "opening -> open -> opening -> open" in the last second, but this is still an improvement over the previous functionality where the cover sometimes got "stuck" if an attribute report was missed.

Thanks @puddly!

Thinking about it, what you describe might be a side effect of the tolerance logic.

i.e.
- Go to 95% command
- The device reports a new position of 94% (95 - tolerance).
  - The handler would conclude it is open (close enough) and clear the target
- The device then immediately reports 95% (the actual finishing point)
  - The handler interprets this as device instigated movement (because the target had been cleared), starting a new 5 second timeout

Setting the tolerance to 0 would allow us to test this hypothesis.

If that's the root cause I suspect we could remove the tolerance because we now have the 5 second timeout after no position update.
Previously it waited forever if the device stopped early/over ran, getting stuck in opening/closing.

_Edited for clarity._

dmulcahey:
> I have a ThirdReality blind and am noticing nothing unusual when testing this PR out. The only thing is the cover state sometimes jittering "opening -> open -> opening -> open" in the last second, but this is still an improvement over the previous functionality where the cover sometimes got "stuck" if an attribute report was missed.

I wonder if we can implement something like the light transitioning logic we have…. Maybe it isn’t worth the effort / complexity.

jeverley:
> > I have a ThirdReality blind and am noticing nothing unusual when testing this PR out. The only thing is the cover state sometimes jittering "opening -> open -> opening -> open" in the last second, but this is still an improvement over the previous functionality where the cover sometimes got "stuck" if an attribute report was missed.
> 
> I wonder if we can implement something like the light transitioning logic we have…. Maybe it isn’t worth the effort / complexity.

I think I might have fixed this particular issue with https://github.com/zigpy/zha/pull/376/commits/517822eb478695cb3c5d9774d9e238cd07d17c56, but I'll take a look at the methods in the light cluster handler this morning and see if there's some learnings we could apply.

jeverley:
I'll be pushing an update to this tomorrow with split methods for handling tilt/lift transition state timeout once I've finished testing it locally.

Trying to handle them both in a single timer creates more problems than it solves with devices supporting both axis.

jeverley:
Hi @puddly, @dmulcahey https://github.com/zigpy/zha/pull/376/commits/dad925c241bec5689fcc2dea982e5e8c524c7685 incorporates the separate timers for lift/tilt movement transitions.

This ensures we correctly handle concurrent movement of the two axis and their transition states.

I've added a test to explicitly validate this scenario performs as expected.

jeverley:
> Thanks!
> 
> I've merged `dev` into this PR in a separate branch ([dev...puddly:zha:cover-platform-entity-status](https://github.com/zigpy/zha/compare/dev...puddly:zha:cover-platform-entity-status)), since as of yesterday we now more explicitly recompute entity capabilities [in a dedicated method](https://github.com/puddly/zha/blob/007c33951294b36045b1075a12c52f5caf3bad87/zha/application/platforms/cover/__init__.py#L116-L138) and have [explicit lifecycle hooks](https://github.com/puddly/zha/blob/007c33951294b36045b1075a12c52f5caf3bad87/zha/application/platforms/cover/__init__.py#L140-L148) for attaching/detaching listeners.
> 
> I have a ThirdReality blind and am noticing nothing unusual when testing this PR out. The only thing is the cover state sometimes jittering "opening -> open -> opening -> open" in the last second, but this is still an improvement over the previous functionality where the cover sometimes got "stuck" if an attribute report was missed.

Hi @puddly, following the merging of https://github.com/zigpy/zha/pull/365 I've implemented the remaining `recompute_capabilities` changes in this PR.

puddly:
Thanks! This PR looks good to me. We will get it into the next Home Assistant beta release.

puddly (approved):

jeverley:
> Thanks! This PR looks good to me. We will get it into the next Home Assistant beta release.

Sorry @puddly, realised I missed the change to ensure supported_features property is no longer cached, sorted now.

TheJulianJES:
This causes test failures in Home Assistant cover tests:
```python
FAILED tests/components/zha/test_cover.py::test_cover - AssertionError: assert 'open' == <CoverState.CLOSED: 'closed'>
FAILED tests/components/zha/test_cover.py::test_cover_failures - AssertionError: assert 'open' == <CoverState.CLOSED: 'closed'>

Results (65.48s (0:01:05)):
     302 passed
       2 failed
         - tests/components/zha/test_cover.py:63 test_cover
         - tests/components/zha/test_cover.py:320 test_cover_failures
       4 skipped
```

We're blocked with a release right now.

TheJulianJES:
~~I'll revert this for now to get a release ready with all other changes.
We might still be able to land this in the beta period if the HA tests are fixed (or behavior in this PR), but I'm not sure.~~

jeverley:
Thanks @TheJulianJES, I'll review the core tests today and raise a PR for them.

TheJulianJES:
Thanks @jeverley! Messaged you on Discord with a short question if you have some time.
But I see that the HA test changes might be easy, so I'll see if I can make them really quick (and not have to revert).

ant1maussion:
I've installed 2025.4.3, but I have still my covers in OPEN state when totally closed.

Here is my devices signature:
```json
{
  "node_descriptor": {
    "logical_type": 1,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4747,
    "maximum_buffer_size": 82,
    "maximum_incoming_transfer_size": 500,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 500,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0202",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0102",
        "0x1000"
      ],
      "output_clusters": [
        "0x0003",
        "0x0019",
        "0x0102"
      ]
    },
    "242": {
      "profile_id": "0xa1e0",
      "device_type": "0x0066",
      "input_clusters": [
        "0x0021"
      ],
      "output_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "NodOn",
  "model": "SIN-4-RS-20",
  "class": "zigpy.quirks.v2.CustomDeviceV2"
}
```

TheJulianJES:
Please open a new issue and attach the diagnostics file from one of the affected covers whilst it is showing the wrong state.

PR #395, opened 2025-03-13
Update script/setup to detect current environment

abmantis:
Check if we are already in a venv before trying to create and activate one. This allows using the script to update the requirements in an existing venv (instead of having to manually delete it first).
This also uses `pip` to install `uv`, instead of using the install script.

### DIFF
--- a/script/setup
+++ b/script/setup
@@ -14,9 +14,19 @@ if [ ! -f "$SETTINGS_FILE" ]; then
     cp "$SETTINGS_TEMPLATE_FILE" "$SETTINGS_FILE"
 fi
 
-curl -LsSf https://astral.sh/uv/install.sh | sh
-uv venv venv
-. venv/bin/activate
+if [ ! -n "$VIRTUAL_ENV" ]; then
+  if [ -x "$(command -v uv)" ]; then
+    uv venv venv
+  else
+    python3 -m venv venv
+  fi
+  source venv/bin/activate
+fi
+
+if ! [ -x "$(command -v uv)" ]; then
+  python3 -m pip install uv
+fi
+
 uv pip install -U pip setuptools pre-commit
 uv pip install -r requirements_test.txt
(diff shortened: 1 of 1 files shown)


puddly (approved):

TheJulianJES (commented):
Runs way faster now. The uv script was really slowing it down a lot.

This and the quirks PR seem fine (https://github.com/zigpy/zha-device-handlers/pull/3958).
Only difference is the order of `uv pip install -e .` and `pre-commit install`.
Do we wanna move it in one PR to have the same order across both projects? HA basically does the following:
1. `uv pip install -r requirements_test.txt`
2. (our `uv pip install -U pip setuptools pre-commit`)
3. `pre-commit install`
4. `uv pip install -e .`

For ZHA, we install pre-commit last. It doesn't really matter tbh, so we can also merge this as-is and adjust the quirks PR to swap the two commands, so both scripts are more consistent.

abmantis:
> Runs way faster now. The uv script was really slowing it down a lot.
> 
> This and the quirks PR seem fine ([zigpy/zha-device-handlers#3958](https://github.com/zigpy/zha-device-handlers/pull/3958)). Only difference is the order of `uv pip install -e .` and `pre-commit install`. Do we wanna move it in one PR to have the same order across both projects? HA basically does the following:
> 
>     1. `uv pip install -r requirements_test.txt`
> 
>     2. (our `uv pip install -U pip setuptools pre-commit`)
> 
>     3. `pre-commit install`
> 
>     4. `uv pip install -e .`
> 
> 
> For ZHA, we install pre-commit last. It doesn't really matter tbh, so we can also merge this as-is and adjust the quirks PR to swap the two commands, so both scripts are more consistent.

Changed the one in zha-device-handlers to keep all `uv pip` commands together.

TheJulianJES (approved):
Thanks!

PR #408, opened 2025-03-31
Replace decimals with suggested_display_precision + remove rounding

P-R-O-C-H-Y:


zha/application/platforms/sensor/__init__.py
```diff
@@ -658,7 +654,7 @@ def formatter(self, value: int) -> int | float:
 
         value = float(value * multiplier) / divisor
         if value < 100 and divisor > 1:
-            return round(value, self._decimals)
+            return round(value, self._attr_suggested_display_precision)
```
puddly, line 674: For the `ElectricalMeasurement` sensor, we effectively set:
- `_attr_suggested_display_precision = 1` if `divisor > 1 and value < 100`
- `_attr_suggested_display_precision = 0` otherwise

Home Assistant caches `suggested_display_precision` so I don't think we can dynamically change it.

Should we globally set the decimals to 1?

CC @TheJulianJES @dmulcahey

P-R-O-C-H-Y: The default `_attr_suggested_display_precision` is 1, so keeping the round here as is now will keep it working the same as it was before.

puddly:
I've updated this PR a bit to be compatible with `dev`. Requires a Core patch:

```diff
diff --git a/homeassistant/components/zha/sensor.py b/homeassistant/components/zha/sensor.py
index a8383857e57..7980a3b53ee 100644
--- a/homeassistant/components/zha/sensor.py
+++ b/homeassistant/components/zha/sensor.py
@@ -114,6 +114,10 @@ class Sensor(ZHAEntity, SensorEntity):
         if hasattr(entity.info_object, "unit") and entity.info_object.unit is not None:
             self._attr_native_unit_of_measurement = entity.info_object.unit
 
+        self._attr_suggested_display_precision = (
+            entity.info_object.suggested_display_precision
+        )
+
         if (
             hasattr(entity, "entity_description")
             and entity.entity_description is not None
```

PR #410, opened 2025-04-07
Verify cover supported features per window covering type

jeverley:
## What's changed
- The `current_position` and `current_tilt_position` state attributes now report `None` if their features are not supported by the device, resolves [#142352](https://github.com/home-assistant/core/issues/142352)
  - This is required because some devices are reporting a position of 0 rather than null
- Revise `supported_features` logic so it _excludes_ tilt only window covering types when enabling lift features
  - This means that lift features will still be enabled as a fallback if the `window_covering_type` attribute value is undefined by the device
  - Previously lift features were enabled for all devices, they should not be enabled for `WCT.Shutter` or `WCT.Tilt_blind_tilt_only` types, see _Table 7-41. Window Covering Type_ from the [Cluster Library Specification 07-5123 Revision 8](https://zigbeealliance.org/wp-content/uploads/2021/10/07-5123-08-Zigbee-Cluster-Library.pdf)
image

- Updates to cover platform tests
  - Check lift only devices don't have tilt features enabled and that state attribute `current_tilt_position` is always None
  - Check tilt only devices don't have lift features enabled and that state attribute `current_position` is always None

## Fixes issue
- Some devices report 0 (open) rather than null values for non-functional tilt/lift position attributes
https://github.com/home-assistant/core/issues/142352

## Related
- https://github.com/zigpy/zha/pull/376

FYI @TheJulianJES, @puddly, @dmulcahey

### DIFF
--- a/zha/application/platforms/cover/__init__.py
+++ b/zha/application/platforms/cover/__init__.py
@@ -51,10 +51,10 @@ _LOGGER = logging.getLogger(__name__)
 
 MULTI_MATCH = functools.partial(PLATFORM_ENTITIES.multipass_match, Platform.COVER)
 
-# Upper limit for dynamic timeout
+# Timeout for device transition state following a position attribute update
 DEFAULT_MOVEMENT_TIMEOUT: float = 5
 
-# Timeout for device transition state following a position attribute update
+# Upper limit for dynamic timeout
 LIFT_MOVEMENT_TIMEOUT_RANGE: float = 300
 TILT_MOVEMENT_TIMEOUT_RANGE: float = 30
 
--- a/zha/application/platforms/cover/__init__.py
+++ b/zha/application/platforms/cover/__init__.py
@@ -90,6 +90,7 @@ class Cover(PlatformEntity):
                 )
             )
         self._supported_features: CoverEntityFeature = CoverEntityFeature(0)
+        self.recompute_capabilities()
 
         self._target_lift_position: int | None = None
         self._target_tilt_position: int | None = None
@@ -106,7 +107,6 @@ class Cover(PlatformEntity):
         self._tilt_transition_timer: asyncio.TimerHandle | None = None
 
         self._state: CoverState | None = CoverState.OPEN
-        self.recompute_capabilities()
         self._determine_cover_state(refresh=True)
 
     def recompute_capabilities(self) -> None:
--- a/zha/application/platforms/cover/__init__.py
+++ b/zha/application/platforms/cover/__init__.py
@@ -49,6 +49,10 @@ if TYPE_CHECKING:
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


zha/application/platforms/cover/__init__.py
```diff
@@ -669,7 +676,7 @@ def state(self) -> dict[str, Any]:
     @functools.cached_property
     def supported_features(self) -> CoverEntityFeature:
         """Return supported features."""
-        return self._attr_supported_features
```
puddly: For consistency with the other platforms and attributes, let's keep the `_attr` prefix for now. These are meant to mirror those in Home Assistant: https://github.com/home-assistant/core/blob/3aae280de50a92cdfcf5c04b088fe9178fba0b2e/homeassistant/components/cover/__init__.py#L240

zha/application/platforms/cover/__init__.py
```diff
@@ -106,32 +104,37 @@ def __init__(
...
-            self._attr_supported_features |= CoverEntityFeature.SET_TILT_POSITION
-            self._attr_supported_features |= CoverEntityFeature.OPEN_TILT
-            self._attr_supported_features |= CoverEntityFeature.CLOSE_TILT
-            self._attr_supported_features |= CoverEntityFeature.STOP_TILT
+            self._supported_features |= (
+                CoverEntityFeature.OPEN_TILT
+                | CoverEntityFeature.CLOSE_TILT
+                | CoverEntityFeature.STOP_TILT
+                | CoverEntityFeature.SET_TILT_POSITION
+            )
```
puddly, line 198: Just for future reference, this is the ZCL spec table that explains the tilt and lift support of each window covering type:

image

jeverley: I've now added this to the PR description for better visibility.

zha/application/platforms/cover/__init__.py
jeverley: Hey @puddly, I've reverted the naming to include the `_attr` prefix, I've also adjusted the `recompute_capabilities` method so that it only updates the final value after evaluating all criteria.

zha/application/platforms/cover/__init__.py
```diff
@@ -123,6 +116,10 @@ async def async_stop_cover(self, **kwargs: Any) -> None:
 class Cover(BaseCover):
     """Representation of a ZHA cover."""
 
+    DEFAULT_MOVEMENT_TIMEOUT: float = 5
+    LIFT_MOVEMENT_TIMEOUT_RANGE: float = 300
+    TILT_MOVEMENT_TIMEOUT_RANGE: float = 30
```
TheJulianJES: This doesn't really need to be in the class, right?
Seems a bit weird to have constants (in caps) as class attributes.

jeverley: Fair enough, thanks for the amendments @puddly

PR #412, opened 2025-04-09
Improve cover state restoration from the HA entity cache

jeverley:
## What's changed
This PR handles edge cases in restoring state after ZHA has been offline.

The current behaviour results in the cover stuck in a restored `OPENING`/`CLOSING` state until the device reports a position update (could be up to 5 mins with some devices).
We also currently restore open/closed state regardless of the device's actual position.

### The below changes have been made to handle state restoration correctly
- When restoring `CLOSING`/`OPENING` states
  - The function checks if the cover is already `CLOSED`/completely `OPEN` (i.e. 100%)
  - The function schedules a call-back to determine the final state after the default timeout period
- We do not restore `CLOSED`/`OPEN` states, instead the state is determined from current position during init

### Other improvements
- The lift/tilt transition and state restore TimerHandles are cancelled upon entity removal

### Updates to cover tests
- Expand state restoration test to handle the different possible scenarios
- Use a common mock device create function (like the ZHA light platform) to reduce code repetition
- Add tests for cancelling TimerHandles on removal of the cover entity


puddly:
I've rebased this PR on top of `dev`, since there was a conflict with the first cover platform PR now being merged.

jeverley:
> I've rebased this PR on top of `dev`, since there was a conflict with the first cover platform PR now being merged.

Thanks @puddly, this one should be ready for review now too.

zha/application/platforms/cover/__init__.py
```diff
@@ -213,12 +213,30 @@ def restore_external_state_attributes(
...
+            and self.current_cover_tilt_position in (100, None)
+        ):
+            return
+
         self._state = state
-        # Target positions have been removed
+        self._loop.call_later(
+            DEFAULT_MOVEMENT_TIMEOUT,
+            functools.partial(self._determine_cover_state, refresh=True),
+        )
```
puddly, line 242: To make sure we don't have lingering tasks and such, add `self._on_remove_callbacks.append(handle.cancel)` to make sure the handle is cleaned up in case the entity is removed before this timer runs.

jeverley: Will fix shortly, assume we should also do this in our `_start_lift_transition` and `_start_tilt_transition` functions?

Edit: wondering if it might be better to append those to `_tracked_handles`, since they are cancelled and overwritten to None by `_clear_lift_transition`/`_clear_tilt_transition`.

puddly: Yeah, this entity should probably override `async def on_remove` to cancel `self._lift_transition_timer` and `self._tilt_transition_timer`. Those are recreated more than once and would be a memory leak if added to `self._on_remove_callbacks` each time.

jeverley: I opted to add the transition clear methods as `_on_remove_callbacks` to avoid us needing to implement a custom `on_remove` method.

PR #413, opened 2025-04-10
Ignore endpoint profiles other than ZHA or ZLL

puddly:
I believe our current behavior of creating entities for any endpoint may not be correct. Develco/Frient, for example, has a "special" endpoint on their devices that has a custom profile but does not conform to the ZCL and results in useless entities being created.

According to the (limited) diagnostics database, no devices lose functionality with this change.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -17,6 +17,7 @@ import zigpy.config
 import zigpy.device
 import zigpy.group
 import zigpy.profiles
+from zigpy.profiles.zha import PROFILE_ID as ZHA_PROFILE_ID
 import zigpy.types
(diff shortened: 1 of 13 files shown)


puddly:
This PR will cause breaking changes for Frient devices:

- The current "identify" button will be deleted and re-created (under a different endpoint), causing an orphaned entity.
- The broken switch entity created for every Frient device will be gone.

I don't think people automate the "Identify" button above so I think this breaking change doesn't need to be handled. Doing so would require fully implementing unique ID migration and providing a way quirks v2 entities to specify metadata to do *just* migrations.

PR #414, opened 2025-04-10
Ignore "non-value" attribute values for sensors

puddly:
Zigpy currently provides a way to determine if a value is a "non-value" for a type. This is easily supported for numeric types but not so much for more complex ones, like arrays and such. This PR will globally fix invalid sensor readings showing up in the frontend.

https://github.com/home-assistant/core/issues/96894

### DIFF
--- a/tests/test_sensor.py
+++ b/tests/test_sensor.py
@@ -34,12 +34,13 @@ from tests.common import (
     get_entity,
     join_zigpy_device,
     send_attributes_report,
+    zigpy_device_from_json,
 )
 from zha.application import Platform
 from zha.application.const import ZCL_INIT_ATTRS, ZHA_CLUSTER_HANDLER_READS_PER_REQ
 from zha.application.gateway import Gateway
 from zha.application.platforms import PlatformEntity, sensor
-from zha.application.platforms.sensor import DanfossSoftwareErrorCode
+from zha.application.platforms.sensor import DanfossSoftwareErrorCode, Temperature
 from zha.application.platforms.sensor.const import SensorDeviceClass, SensorStateClass
 from zha.units import PERCENTAGE, UnitOfEnergy, UnitOfPressure, UnitOfVolume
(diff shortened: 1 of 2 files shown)


oddlama:
Since this is a draft, can I use this to patch my local instance and test it, or do you consider it work in progress?

puddly:
Sure! I haven't written unit tests for it but have tested it at runtime with a device.

oddlama:
I just tested this with my own custom zigbee device, and I can confirm that this fixes the issue completely! Thank you very much @puddly ! And the sensor correctly reports as "Unknown" while we send the unknown value :D

This is how the mock sensor values look like (it cycles between two modes, first stepdown from 20° -> -20°, then an unknown value for the same time):

image

puddly:
Thanks for the confirmation!

tests/test_sensor.py
```diff
@@ -1748,3 +1749,34 @@ async def test_quirks_sensor_attr_converter(zha_gateway: Gateway) -> None:
 
     await send_attributes_report(zha_gateway, cluster, {"present_value": 0})
     assert entity.state["state"] == 100.0
+
+
+async def test_ignore_non_value(zha_gateway: Gateway) -> None:
+    """Test ZHA quirks v2 sensor with attribute_converter."""
```
TheJulianJES: docstring still needs to be updated

zha/application/platforms/sensor/__init__.py
```diff
@@ -173,6 +174,13 @@ def __init__(
     ) -> None:
         """Init this sensor."""
         self._cluster_handler: ClusterHandler = cluster_handlers[0]
+        self._attr_def: foundation.ZCLAttributeDef | None = None
+
+        if self._attribute_name is not None:
```
TheJulianJES, line 179: If you wanted to, could use `if (attr := self._attribute_name) is not None:` to just access `attr` for the name below, but this is obviously also fine.

TheJulianJES (approved):

PR #416, opened 2025-04-12
Fix `current_cover_tilt_position` for shades, add base class

TheJulianJES:
## Proposed change

This PR fixes an issue where the `current_cover_tilt_position` property was removed for the `Shade` implementation, based on the `LevelControl` cluster (in https://github.com/zigpy/zha/pull/376).

The Home Assistant integration does not differentiate between our `Cover` and `Shade` implementation, so tries to access the `current_cover_tilt_position` attribute/property regardless: https://github.com/home-assistant/core/blob/d23c9f715e1f7772ec8ae176f3db4b9afa7db95d/homeassistant/components/zha/cover.py#L122-L125

We need to make sure we provide it in all our cover implementations, so this PR adds it back, returning `None` by default, as a `Shade` cannot support tilt.

A new abstract `BaseCover` class is also added to make sure all relevant methods are implemented by the `Cover` and `Shade` implementations.

## Additional information

Should address:
- https://github.com/home-assistant/core/issues/142815
- https://github.com/home-assistant/core/issues/142601

We should likely add tests for this in HA in the future. I'm guessing `Shade` is not covered at all.

### DIFF
--- a/tests/test_cover.py
+++ b/tests/test_cover.py
@@ -795,6 +795,7 @@ async def test_shade(
     # coverage (these are always None for now)
     assert entity.is_opening is None
     assert entity.is_closing is None
+    assert entity.current_cover_tilt_position is None
 
     # test that the state has changed from unavailable to off
     await send_attributes_report(
--- a/zha/application/platforms/cover/__init__.py
+++ b/zha/application/platforms/cover/__init__.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from abc import ABC, abstractmethod
 import asyncio
 from collections import deque
 import functools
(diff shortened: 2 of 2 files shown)


puddly:
Could we derive both of them from a single abstract base class that implements the proper API?

TheJulianJES:
The added line is uncovered, but it looks like all ZHA tests generally use the `state` property to access various state of the entities. But we do not include the tilt percentage in the state for shades, as it's always `None`, so we'll need to access that property directly in tests, should be fine though.
-> Added in [`2195961` (#416)](https://github.com/zigpy/zha/pull/416/commits/2195961ed8fa63689282c1ecd8490d1550d5b352)

Compared to accessing the state via `state` in ours tests, HA directly accesses the different properties.

TheJulianJES:
> Could we derive both of them from a single abstract base class that implements the proper API?

Probably.

TheJulianJES:
The implementations don't share a lot, so the commit is really only the shared API used by HA and setting `PLATFORM` + `_attr_primary_weight`.
Also, it seems like ruff doesn't enforce whether to have `pass` or just the docstring in empty methods. What do we prefer for that? (edit: I guess nothing, so Codecov doesn't complain)

`async_open_cover_tilt, async_close_cover_tilt`, `async_set_cover_tilt_position`, and `async_stop_cover_tilt` are methods only used by `Cover` from HA. I've not added these to the base class, as we'd also need to implement them for `Shade` otherwise, and HA should only call them if the `supported_features` contain the corresponding tilt features.


(On an unrelated note, we should not really set `_attr_translation_key` if a device class is already present, as the generic device class names should be fine for our purpose and we don't really need another one yet.)

PR #427, opened 2025-04-23
Keep the device firmware version in sync

puddly:
At some point during the library migration, devices stopped showing their installed firmware in the frontend. This PR restores this behavior in the library using entities and events, allowing for a relatively simple re-implementation of the feature in Core.

image

At some point, I would like to split the exposed firmware version into two values: a numeric value for comparison and a string value for display. We could then retain the generic `0xabcd1234` format for generic devices but then allow quirks to create firmware version formatters.


zha/zigbee/device.py
```diff
@@ -869,16 +887,32 @@ async def async_initialize(self, from_cache: bool = False) -> None:
         # At this point we can compute a primary entity
         self._compute_primary_entity()
 
+        # Sync the device's firmware version with the first platform entity
+        for (platform, _unique_id), entity in self.platform_entities.items():
+            if platform != Platform.UPDATE:
+                continue
```
TheJulianJES, line 938: Seems a bit weird to go from entity level back up to device level for this info. We create a small (unnecessary?) dependence here.

Alternative might be for the `device.firmware_version` to return the cached `current_file_version` attribute (like the update entity currently does) and also have the device subscribe to `current_file_version` attribute changes (instead of the update entity doing that).
Then, you would have the update entity use `device.firmware_version` and have it subscribe to the `DeviceUpdatedEvent` events.

TheJulianJES (commented):
There's also a merge conflict now.

zha/zigbee/device.py
puddly: Good point. I was thinking of this from the opposite direction: the update entity can be overriden per-device and then provide a formatted firmware version (without the ZHA device needing to be patched). It would avoid having the device scan through clusters to determine the correct firmware version.

TheJulianJES: > the update entity can be overriden per-device and then provide a formatted firmware version (without the ZHA device needing to be patched)

Does the base ZHA device need to know the "friendly version" though? For HA, we generally use the update entity for that, right? Except for the little firmware version string in the device card/registry...

But yeah, I'm fine with this as well, so feel free to merge as-is if you want.

puddly:
Requires the following Core change:

```diff
diff --git a/homeassistant/components/zha/helpers.py b/homeassistant/components/zha/helpers.py
index 084e1c882ac..f5b44eb8fc4 100644
--- a/homeassistant/components/zha/helpers.py
+++ b/homeassistant/components/zha/helpers.py
@@ -74,7 +74,12 @@ from zha.event import EventBase
 from zha.exceptions import ZHAException
 from zha.mixins import LogMixin
 from zha.zigbee.cluster_handlers import ClusterBindEvent, ClusterConfigureReportingEvent
-from zha.zigbee.device import ClusterHandlerConfigurationComplete, Device, ZHAEvent
+from zha.zigbee.device import (
+    ClusterHandlerConfigurationComplete,
+    Device,
+    DeviceFirmwareInfoUpdatedEvent,
+    ZHAEvent,
+)
 from zha.zigbee.group import Group, GroupInfo, GroupMember
 from zigpy.config import (
     CONF_DATABASE,
@@ -843,8 +848,23 @@ class ZHAGatewayProxy(EventBase):
                 name=zha_device.name,
                 manufacturer=zha_device.manufacturer,
                 model=zha_device.model,
+                sw_version=zha_device.firmware_version,
             )
             zha_device_proxy.device_id = device_registry_device.id
+
+            def update_sw_version(event: DeviceFirmwareInfoUpdatedEvent) -> None:
+                """Update software version in device registry."""
+                device_registry.async_update_device(
+                    device_registry_device.id,
+                    sw_version=event.new_firmware_version,
+                )
+
+            self._unsubs.append(
+                zha_device.on_event(
+                    DeviceFirmwareInfoUpdatedEvent.event_type, update_sw_version
+                )
+            )
+
         return zha_device_proxy
 
     def _async_get_or_create_group_proxy(self, group_info: GroupInfo) -> ZHAGroupProxy:
```

Issue #424, opened 2025-04-19
Cover stay in OPEN state for NodOn SIN-4-RS-20


ant1maussion:
[zha-01J7ESJY790MB5RK9MXVS8ER54-NodOn SIN-4-RS-20-a99b44b9d7e2af58df8639f59042ac83.json](https://github.com/user-attachments/files/19822202/zha-01J7ESJY790MB5RK9MXVS8ER54-NodOn.SIN-4-RS-20-a99b44b9d7e2af58df8639f59042ac83.json)

Since HA version 2025.4, my covers have their state sticked in OPEN state whatever the real cover position.
I'm using NodOn SIN-4-RS-20 module.

Here is the signature:
```json
{
  "node_descriptor": {
    "logical_type": 1,
    "complex_descriptor_available": 0,
    "user_descriptor_available": 0,
    "reserved": 0,
    "aps_flags": 0,
    "frequency_band": 8,
    "mac_capability_flags": 142,
    "manufacturer_code": 4747,
    "maximum_buffer_size": 82,
    "maximum_incoming_transfer_size": 500,
    "server_mask": 11264,
    "maximum_outgoing_transfer_size": 500,
    "descriptor_capability_field": 0
  },
  "endpoints": {
    "1": {
      "profile_id": "0x0104",
      "device_type": "0x0202",
      "input_clusters": [
        "0x0000",
        "0x0003",
        "0x0004",
        "0x0005",
        "0x0102",
        "0x1000"
      ],
      "output_clusters": [
        "0x0003",
        "0x0019",
        "0x0102"
      ]
    },
    "242": {
      "profile_id": "0xa1e0",
      "device_type": "0x0066",
      "input_clusters": [
        "0x0021"
      ],
      "output_clusters": [
        "0x0021"
      ]
    }
  },
  "manufacturer": "NodOn",
  "model": "SIN-4-RS-20",
  "class": "zigpy.quirks.v2.CustomDeviceV2"
}
```

TheJulianJES:
Please download and attach the device diagnostics file from the device page while the cover is in the unexpected state.

TheJulianJES:
Also, did you open and close the blinds to possibly recompute the state?

ant1maussion:
> Also, did you open and close the blinds to possibly recompute the state?

Yes already tried

jeverley:
[...]

PR #428, opened 2025-04-26
Revert 'open' tilt state precedence for cover devices supporting lift & tilt

jeverley:
## Context
Some Zigbee Window Covering devices such as the Nodon SIN-4-RS-20 and Vimar Window_Cov_v1.0 do not contain any motor mechanisms themselves, instead they act as relays controlling an existing window motor, and are connectable to the existing physical switch.

The devices report a type of `tilt_blind_tilt_and_lift`, they do genuinely support both types of control when used with venetian blinds, however they can also be used in a lift only configuration with other cover types (effectively the tilt cluster commands go unused here).

This type enables both lift and tilt HA `supported_features` which enables their associated UI controls.
It also means that both lift and tilt position are used in ZHA to derive the cover state, using the logic introduced in https://github.com/zigpy/zha/pull/376.

### The problem
By default these devices report a tilt position of 0 (open), for users _not_ using tilt it means the entity is left stuck in an `OPEN` state.

In Zigbee you cannot override the `window_cover_type` because cluster attribute is read/report only, as a result some users have been using configuration.yaml overrides like below to hide the tilt UI controls, see https://github.com/home-assistant/core/issues/143080 and https://community.home-assistant.io/t/disable-tilt/639816.

```
homeassistant:
  customize:
    cover.your_cover:
      supported_features: 15
      device_class: blind
```
**Overriding the devices'  `window_cover_type` using a quirk in `zha-device-handlers` would solve the issue for lift only users, but would disable the tilt functionality for users that are using them with venetian blinds.**

## Options considered
1. Merge a quirk into `zha-device-handlers` that overrides the `window_covering_type` for these devices
    - **pros**: requires no ZHA change
    - **cons**: disables tilt functionality for all users of these devices
2. Ask the users of these devices that don't use tilt to employ a local quirk that overrides the `window_covering_type`
    - **pros**: requires no ZHA change
    - **cons**: no clear place to document this workaround, not user friendly
3. Revert the logic that currently overrides `CLOSED` lift when tilt is `OPEN`
    - **pros**: solves the issue of the cover being stuck `OPEN` when not using tilt
    - **cons**: the cover sate is less accurate for users using both lift _and_ tilt
4. Implement a ZHA config entity for Window Covers that allows users to choose lift_and_tilt/lift/tilt - _this entity would only be exposed for this particular window_covering_type_
    - **pros**: most flexible and works with future devices, handles non ZCL concepts (state/UI features) within ZHA
    - **cons**: more complex, the `device_class` and `supported_features` are currently only evaluated during ZHA entity init (though core HA does support updates to these)
5. Implement quirks for these devices that add a local config option that overrides the behaviour.
    - **pros**: requires no ZHA change
    - **cons**: requires local cluster attributes and logic for exposing the functionality in quirks, not appropriate as this isn't an issue specific to a single device

## What's changed
In this interim PR I've implemented **Option 3**:

This removes the line in function `_determine_cover_state` which gives precence `OPEN` tilt over `CLOSED` lift.
Instead only the lift state is used to drive the static (`OPEN`/`CLOSED`) state.

Whilst it does come with the downside of reduced state accuracy for users using both lift _and_ tilt, it addresses the immediate issue of users having their device stuck in `OPEN` state (and follows the logic prior to https://github.com/zigpy/zha/pull/376 / 2025.4).

**If we don't want to revert the 'correct' behaviour I can put together a quick quirk (option 2) for the users in the issue threads to use locally.**

**I'm currently progressing a proof of concept option 4, but this will take more time to implement.**

### Addresses issues
- https://github.com/zigpy/zha/issues/424
- https://github.com/home-assistant/core/issues/142820
- https://github.com/home-assistant/core/issues/143080


puddly:
I think for the time being, restoring the old "broken" behavior is the simplest approach. A few seconds delay for state changes is a minor annoyance compared to it being stuck in an incorrect state.

PR #440, opened 2025-04-30
Fix printing log message for `None` profile id

TheJulianJES:
## Proposed change

This fixes an issue where we error out on if `ep.profile_id` is `None` when trying to print a log message whilst formatting the int into a hex representation.
This PR adds another message if the `profile_id` is `None` (which it can be according to typing and apparently is in some HA Core tests).

Or are there better ideas?

## Additional information

This is what should be needed to get some HA Core tests passing. Apparently we do have some `None` `profile_id` in tests there.
I think I've also seen some devices doing (or quirks?) doing this, so we should account for this.

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -961,18 +961,5 @@ async def test_endpoint_none_profile(
     zigpy_dev.endpoints[3].profile_id = None
     zha_device = await join_zigpy_device(zha_gateway, zigpy_dev)
 
-    assert zha_device.async_get_clusters() == {
-        3: {
-            CLUSTER_TYPE_IN: {
-                general.Basic.cluster_id: zigpy_dev.endpoints[3].in_clusters[
-                    general.Basic.cluster_id
-                ],
-                general.OnOff.cluster_id: zigpy_dev.endpoints[3].in_clusters[
-                    general.OnOff.cluster_id
-                ],
-            },
-            CLUSTER_TYPE_OUT: {},
-        }
-    }
     assert zha_device.async_get_std_clusters() == {}
     assert "Skipping endpoint, profile is None" in caplog.text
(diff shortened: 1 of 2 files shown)


puddly:
I have a branch of Core ready that fixes these logging issues. It realistically should never be `None`, we sort of just allowed it to be.

Which devices/quirks did you see doing this?

TheJulianJES:
I think some Aqara remotes. There's one quirk here:
https://github.com/zigpy/zha-device-handlers/blob/c2f64fd0a34fbac677bdac7b1a5fe069c69dd0a5/zhaquirks/xiaomi/aqara/opple_remote.py#L191-L194

I'm not actually sure if a completely empty endpoint will also mean `None` profile id though.
So that might not be an issue?

puddly:
Hmmmm. I think the quirk may actually end up matching *any* endpoint with an ID of 3? I'm not sure what the empty `replacement` endpoints will do. We supposedly have [a variant of this device](https://github.com/zigpy/zha/blob/695475a927451524475bdcef90a25177727e2345/tests/data/devices/lumi-lumi-remote-b286opcn01.json#L2037) in the database and its endpoints are all populated properly in the Zigpy database.

Here's the Core branch, by the way: https://github.com/home-assistant/core/compare/dev...puddly:core:puddly/zha-deps-bump-20250423. All tests pass for me with it.

TheJulianJES:
Yeah, ok. We can just go ahead with that for now then.

The device `lumi.remote.b286opcn01` has multiple different signatures. You might only get the empty endpoints when pairing with bellows versions (or TI coordinator firmware) that does not send Xiaomi manufacturing codes during pairing. Or it was just "random" with this device. Not sure anymore.

I'm pretty sure the device signature in ZHA showed the empty `{}` endpoints though. We just added it 1:1 and it luckily worked. I can't find the old issues right now. I think there are some comments with these signatures in *some* issue though.

EDIT: I did find a really old issue with "null" `profile_id` btw: <https://github.com/zigpy/zha-device-handlers/pull/690>
Also: https://github.com/zigpy/zha-device-handlers/issues/940 and https://github.com/zigpy/zha-device-handlers/issues/2258.

TheJulianJES:
Added a basic test now. The new log message is indeed the only place `None` `profile_id` fails.

PR #442, opened 2025-04-30
Clean up `number` platform

puddly:
Should resolve https://github.com/zigpy/zha-device-handlers/pull/3821#issuecomment-2842596655.

- I've reorganized the platform code to use a `BaseNumber` entity, with subclasses implementing `native_value` and `async_set_native_value`.
- All entity metadata is set in `recompute_capabilities()`.
- `_init_from_quirks_metadata` is implemented on `NumberConfigurationEntity`. Should we move it up in the hierarchy?
- We use `fallback_name` instead of `description` for standardization.

### DIFF
--- a/zha/application/platforms/number/__init__.py
+++ b/zha/application/platforms/number/__init__.py
@@ -51,7 +51,6 @@ class NumberEntityInfo(BaseEntityInfo):
     native_max_value: float
     native_min_value: float
     native_step: float | None
-    native_value: float | None
     native_unit_of_measurement: str | None
 
 
@@ -83,7 +82,6 @@ class BaseNumber(PlatformEntity):
             native_max_value=self.native_max_value,
             native_min_value=self.native_min_value,
             native_step=self.native_step,
-            native_value=self.native_value,
             native_unit_of_measurement=self.native_unit_of_measurement,
         )
 
@@ -154,7 +152,11 @@ class Number(BaseNumber):
         self._attr_native_unit_of_measurement = UNITS.get(
             analog_output.engineering_units
         )
-        self._attr_icon = ICONS.get(analog_output.application_type >> 16)
+
+        if analog_output.application_type is not None:
+            self._attr_icon = ICONS.get(analog_output.application_type >> 16)
+        else:
+            self._attr_icon = None
 
         if analog_output.description is not None:
             self._attr_fallback_name = analog_output.description
--- a/zha/application/platforms/number/__init__.py
+++ b/zha/application/platforms/number/__init__.py
@@ -123,7 +123,7 @@ class BaseNumber(PlatformEntity):
 
 
 @STRICT_MATCH(cluster_handler_names=CLUSTER_HANDLER_ANALOG_OUTPUT)
-class Number(BaseNumber):
+class AnalogOutputNumber(BaseNumber):
     """Representation of a ZHA Number entity."""
 
     _attr_translation_key: str = "number"
--- a/zha/application/platforms/number/__init__.py
+++ b/zha/application/platforms/number/__init__.py
@@ -47,31 +47,88 @@ CONFIG_DIAGNOSTIC_MATCH = functools.partial(
 class NumberEntityInfo(BaseEntityInfo):
     """Number entity info."""
 
-    engineering_units: int
-    application_type: int
-    min_value: float | None
-    max_value: float | None
-    step: float | None
+    mode: NumberMode
+    native_max_value: float
+    native_min_value: float
+    native_step: float | None
+    native_value: float | None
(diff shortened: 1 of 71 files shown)


puddly:
For Core, this should get rid of the `name` override by setting `fallback_name`:

```diff
diff --git a/homeassistant/components/zha/number.py b/homeassistant/components/zha/number.py
index 567e2a5b37a..401ffc39651 100644
--- a/homeassistant/components/zha/number.py
+++ b/homeassistant/components/zha/number.py
@@ -46,17 +46,6 @@ async def async_setup_entry(
 class ZhaNumber(ZHAEntity, RestoreNumber):
     """Representation of a ZHA Number entity."""
 
-    @property
-    def name(self) -> str | UndefinedType | None:
-        """Return the name of the number entity."""
-        if (description := self.entity_data.entity.description) is None:
-            return super().name
-
-        # The name of this entity is reported by the device itself.
-        # For backwards compatibility, we keep the same format as before. This
-        # should probably be changed in the future to omit the prefix.
-        return f"{super().name} {description}"
-
     @property
     def native_value(self) -> float | None:
         """Return the current value."""
```

zha/application/platforms/number/__init__.py
```diff
@@ -266,17 +268,6 @@ def _init_from_quirks_metadata(self, entity_metadata: NumberMetadata) -> None:
...
-            **super().info_object.__dict__,
-            min_value=self._attr_native_min_value,
-            max_value=self._attr_native_max_value,
-            step=self._attr_native_step,
-            multiplier=self._attr_multiplier,
-        )
-
     @property
     def state(self) -> dict[str, Any]:
         """Return the state of the entity."""
```
TheJulianJES, line 277: This can now be removed from `NumberConfigurationEntity`, as it's already implemented in `BaseNumber`.

zha/application/platforms/number/__init__.py
```diff
@@ -119,59 +96,92 @@ def state(self) -> dict[str, Any]:
         response["state"] = self.native_value
         return response
```
TheJulianJES, line 93: Kinda related: Should we include the min/max value attributes (and possibly others) in `state`?
`state` is only really important for `maybe_emit_state_changed_event`. If only the min/max values were to change, the method would prevent an update from going through, unless `native_value` is also changed at the same time.
This would matter if we were to "dynamically" call  `recompute_capabilities` (i.e. after sw updates) in the future.

For the analog input entity, if `min_present_value`/`max_present_value` ZCL attributes were updated at runtime before these changes (whilst `native_value` is also changed due to what I mentioned above), the update of the min/max values would populate to HA as well.
Looking at this now, it would only happen after `recompute_capabilities` is called, but that's not done for a simple attribute change, so we always prevent min/max changes for analog input entities going to HA now, I think.

In reality, I doubt any devices/quirks exposing an `AnalogInput` cluster did this before, so it might not be an issue.

The same would also apply for  the entities that use`ZCLHeatSetpointLimitEntity` (i.e. `MaxHeatSetpointLimit` and `MinHeatSetpointLimit`), as they overrode the methods HA called for min/max values, when HA state is written. Now it's also only done in `recompute_capabilities`, so not possible "at runtime" (after the device is initialized).
Doubt that matters as well though.

TheJulianJES (approved):
Seems good, but note the comment below about changing the discovery for v2 quirks to `NumberConfigurationEntity`.

TheJulianJES:
> `_init_from_quirks_metadata` is implemented on `NumberConfigurationEntity`. Should we move it up in the hierarchy?

To `BaseNumber`? Not really, as we should never need to initialize the AnalogInput only `Number` class with quirks v2 metadata.

But we should update the discovery for quirks v2 to always initialize `NumberConfigurationEntity`.
Since we'll never need to initialize the `Number` class for quirks v2 (as that's now only for `AnalogInput` entities). https://github.com/zigpy/zha/blob/4afffe14dcdd315de531c2635615cc15852f57c0/zha/application/discovery.py#L192-L205 

Maybe we should also rename `Number` class to `AnalogInputNumber`?

Actually, we should be able to completely remove the `STANDARD`/`CONFIG`/`DIAGNOSTIC` type from the discovery schema IMO.

zha/application/platforms/number/__init__.py
puddly: Yeah, that's true. I think we maybe could emit a new event when capabilities change. I'd like to eventually figure out
`info_object` in preparation for the client/server split to allow devices to emit both types of updates in a more type-checked manner.

puddly:
Good point. I've implemented the above change and will make a follow-up to remove the unnecessary entity type handling in the mapping.

PR #443, opened 2025-05-02
Remove `EntityType` from quirks v2 entity metadata handling

puddly:
I don't see any regressions with existing devices so I think this change is good to go. We had only a few entity types without all three `EntityType` mappings already in place.

### DIFF
--- a/tests/test_discover.py
+++ b/tests/test_discover.py
@@ -614,6 +614,15 @@ async def test_quirks_v2_entity_discovery_errors(
     zigpy_device = _get_test_device(
         zha_gateway, "Ikea of Sweden3", "TRADFRI remote control3"
     )
+
+    # Inject unknown quirks v2 entity metadata
+    class UnknownEntityMetadata:
+        entity_platform = Platform.UPDATE
+
+    zigpy_device._exposes_metadata[
+        (1, zigpy.zcl.clusters.general.OnOff.cluster_id, ClusterType.Server)
+    ].append(UnknownEntityMetadata())
+
     zha_device = await join_zigpy_device(zha_gateway, zigpy_device)
 
     assert (
@@ -732,9 +741,7 @@ async def test_quirks_v2_metadata_bad_device_classes(
     zigpy.quirks._DEVICE_REGISTRY.remove(zigpy_device)
 
 
-async def test_quirks_v2_fallback_name(
-    zha_gateway: Gateway,  # pylint: disable=unused-argument
-) -> None:
+async def test_quirks_v2_fallback_name(zha_gateway: Gateway) -> None:
     """Test quirks v2 fallback name."""
 
     zigpy_device = _get_test_device(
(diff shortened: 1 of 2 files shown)


tests/test_discover.py
```diff
@@ -732,17 +741,15 @@ async def test_quirks_v2_metadata_bad_device_classes(
...
+async def test_quirks_v2_fallback_name(zha_gateway: Gateway) -> None:
     """Test quirks v2 fallback name."""
 
     zigpy_device = _get_test_device(
         zha_gateway,
         "Ikea of Sweden6",
         "TRADFRI remote control6",
         augment_method=lambda builder: builder.sensor(
-            attribute_name=zigpy.zcl.clusters.general.OnOff.AttributeDefs.off_wait_time.name,
+            attribute_name=zigpy.zcl.clusters.general.OnOff.AttributeDefs.global_scene_control.name,
```
TheJulianJES, line 752: Why did this need to change? Conflicts with something in `_get_test_device`?

TheJulianJES (approved):

tests/test_discover.py
puddly: I'm actually not sure why it passed *originally*, there's already a sensor with this name present within ZHA.

TheJulianJES: Oh, I see. In `_get_test_device`, yeah.
I guess because the unique ids were different due to us using the analog input/output(?) class before and that didn't properly use the attribute name for the unique id.

Now, they conflict (like they should have before).

EDIT: Eh, on second thought, not sure that makes sense, since this is about sensors.
But it's something along those lines I bet, since there's also a number for that attribute and two sensors originally and only the number behavior changed.

Oh, but it was already changed to `NumberConfigurationEntity` before this PR. Hmm, I dunno.
Change seems good though.

puddly: This is caused by ZHA not creating `ZCLSensorMetadata` entities with `CONFIG` entity type. An earlier quirk in the unit test now started creating the same entity as was being tested here, which created an unintentional collision. The fix in this PR is valid but the explanation I provided above was not. This can be considered resolved.

TheJulianJES: To also explain it here:

This `sensor` entity was never actually created, since there was no mapping, as it uses `EntityType.CONFIG`:
https://github.com/zigpy/zha/blob/cadf77d62bd4fea85027bf425174222cb5edc5a1/tests/test_discover.py#L564-L570

The sensor for this test used the same attribute name, thus the same unique id.
Before this PR, it did not conflict with the other sensor entity, since the former one was never created. Now, both are created and they conflict.
https://github.com/zigpy/zha/blob/cadf77d62bd4fea85027bf425174222cb5edc5a1/tests/test_discover.py#L744-L749

So, the proper fix (as done in this PR) is to just use a different attribute name in our tests, which changes the unique id.
An alternative would be to just add a unique id suffix in the quirks v2 builder method, but since these are just tests, we can switch to another attribute name entirely.

PR #444, opened 2025-05-02
Remove redundant `"number"` translation key and `description`

TheJulianJES:
## Proposed change

This removes redundant calls to set the fallback name to the ZCL description attribute in `on_add` of the sensor and number platforms, as this is already done in `recompute_capabilities`.
Calling this in `recompute_capabilities` also makes sure that the device is initialized (i.e. the attributes read once).

Furthermore, the default `"number"` translation key is also removed for the number platform to align with the sensor platform. Home Assistant will still default to the "Number" (US-EN) translation, same as before, if neither translation key and nor fallback name are provided by ZHA.

This allows for a cleaner implementation of setting the fallback name to the description name, without needing to unset the translation key or potentially needing to re-set it to `number` if the `description` attribute no longer has a value.
(Assuming `recompute_capabilities` can dynamically be called later at runtime, e.g. after a firmware update.)

## Additional information

Addresses comments in https://github.com/zigpy/zha/pull/197#pullrequestreview-2812755262

---

<details><summary>Old PR description of initial version (no longer relevant)</summary>
<p>

## Proposed change

This removes the redundant `hasattr(ch, "description")` check for `AnalogOutputNumber` and `AnalogInputSensor`.
For the latter, getting the ZCL `description` attribute is also moved into the `on_add` method of `AnalogInputSensor`, as it's not needed in the base `Sensor` class.

The typing for `_attr_translation_key` in `AnalogOutputNumber` is also fixed to include the option of it being `None` if the `description` attribute has a value.

I can also remove the walrus operator again if that's preferred. It's often used in HA code though.

### Important

Also, we set the fallback name to the description in `recompute_capabilities` for all three platforms supporting the `description` attribute (binary sensor, sensor, and number), as well as initially setting it in `on_add`.
For the number and analog sensor platforms, we also unset the translation key in `on_add` only.

However, `on_add` is called before the cluster handlers are initialized (so attributes read once).
If a device is freshly paired, even if a `description` attribute is present, the translation key would take priority over the fallback name, since `on_add` is executed before the attribute is read.
After a restart of HA, this behavior will change, as the attribute is already cached then.

It also seems a bit redundant to have this logic both in `on_add` and partly also in `recompute_capabilities`.
So, can we just remove it from `on_add` and include unsetting the translation key if `description` is present in `recompute_capabilities`? 

</p>
</details>


puddly:
I totally agree with the `on_add` / `recompute_capabilities` problem, it definitely should be in the latter.

puddly (approved):

TheJulianJES:
Want me to (re-)move it in a subsequent PR or here?

TheJulianJES:
I'll put it in here. It's mostly just code removal.

TheJulianJES:
Ok, slightly bigger change but IMO for the better:

We've previously had a "number" translation key when no fallback name or other translation key was specified.
That translation key just refers to the number title translation key in HA:
 - [homeassistant/components/zha/strings.json#L681-L684](https://github.com/home-assistant/core/blob/4450f919c3f2ad7c69a71ca153fd92f74c581d95/homeassistant/components/zha/strings.json#L681-L684)
 - to here: [homeassistant/components/number/strings.json#L2](https://github.com/home-assistant/core/blob/4450f919c3f2ad7c69a71ca153fd92f74c581d95/homeassistant/components/number/strings.json#L2)

It gets a bit weird with setting the fallback name and unsetting the translation key (or the other way around, depending on if `description` is available) for the number platform in `recompute_capabilities` then. Especially under the assumption we may dynamically call that method later and the ZCL attribute can change, causing both translation key and fallback name to potentially change again.

(Although I did just realize this would require HA changes to work as well, as we persist the fallback name to `_attr_name` there once it was available once: [homeassistant/components/zha/entity.py#L73](https://github.com/home-assistant/core/blob/4450f919c3f2ad7c69a71ca153fd92f74c581d95/homeassistant/components/zha/entity.py#L73))

For the sensor platform, we simply don't provide a translation key by default. So, if there's no translation key value in HA and also no fallback name, it defaults to the `_` device class, which points to the title (so exact same translation key):
- now for numbers: [homeassistant/components/number/strings.json#L13](https://github.com/home-assistant/core/blob/4450f919c3f2ad7c69a71ca153fd92f74c581d95/homeassistant/components/number/strings.json#L13)
- and for sensors: [homeassistant/components/sensor/strings.json#L120](https://github.com/home-assistant/core/blob/4450f919c3f2ad7c69a71ca153fd92f74c581d95/homeassistant/components/sensor/strings.json#L120)
- called from here: [homeassistant/helpers/entity.py#L612-L615](https://github.com/home-assistant/core/blob/4450f919c3f2ad7c69a71ca153fd92f74c581d95/homeassistant/helpers/entity.py#L612-L615)
- initially called from `name()` here: [homeassistant/helpers/entity.py#L740-L741](https://github.com/home-assistant/core/blob/4450f919c3f2ad7c69a71ca153fd92f74c581d95/homeassistant/helpers/entity.py#L740-L741)

Basically, the number platform is a bit more like the sensor platform now. There should be no changes on the HA side due to us removing the default `"number"` translation key, as it'll still default to "Number" for non-primary number entities with no specified translation key and no fallback name.

So, the fallback name is now only set to the description attribute in `AnalogInputSensor.recompute_capabilities()`, `AnalogOutputNumber.recompute_capabilities()` and `BinaryInputWithDescription.recompute_capabilities` in the same fashion. No special treatment with unsetting the translation key for `number` anymore.

Also updated the PR description. Let me know what you think.
Can split up the `description` removal from the number translation key removal into a separate PR if you prefer.

PR #455, opened 2025-05-25
Fix device diagnostics not regenerating with changed quirks

TheJulianJES:
## Proposed change

This fixes an issue where regenerating device diagnostics or importing legacy device diagnostics might not work if `cluster_details` still contains clusters that are no longer present on the `Device` object.

## Additional information

This can be caused when a quirk no longer exposes a certain cluster.
Noticed here: https://github.com/zigpy/zha-device-handlers/pull/3232#issuecomment-2908002330

Draft until this is ready, as I've also had the same issue not just when regenerating diagnostics, but also when importing legacy diagnostics, though I just removed the clusters from the JSON for now.
I'll come back to this later.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -387,9 +387,18 @@ def zigpy_device_from_device_data(
 
         for cluster_type in ("in_clusters", "out_clusters"):
             for cluster in ep[cluster_type]:
-                real_cluster = getattr(endpoint, cluster_type)[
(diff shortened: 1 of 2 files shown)


TheJulianJES:
Indeed. A more extensive fix was included here:
- https://github.com/zigpy/zha/pull/557

Closing...

PR #458, opened 2025-05-28
Add total_active_power attribute entity

abmantis:
`total_active_power` is currently used by some quirks, but there was no entity for it. This PR adds it.

It is defined with `_skip_creation_if_no_attr_cache = True`, so it will only be created if there is a value for it.


TheJulianJES:
> It is defined with `_skip_creation_if_no_attr_cache = True`, so it will only be created if there is a value for it.

Thinking about this and https://github.com/zigpy/zha-device-handlers/pull/4095, all Aqara plugs (that do power measurement) will still have the attribute present in cache for existing devices. We may need to clean this up in the affected quirks by calling `update_attribute(id, None)` (in the constructor or something like that).

I'm also not sure if other power measurement plugs (just single phase) also update both attributes. We'd have two sensors for all of them then. Hmm..
I should really get to https://github.com/zigpy/zha/issues/311, so we might be able to make this opt-in entirely?
Or just use a custom sensor for the Tuya quirks? But if we add the `total_active_power` sensor later, it's also not very ideal.

abmantis:
@TheJulianJES I've updated https://github.com/zigpy/zha-device-handlers/pull/4095 to reset the attribute so it shouldn't show up.

> I'm also not sure if other power measurement plugs (just single phase) also update both attributes. We'd have two sensors for all of them then. Hmm..

Mine do not, but I don't have much variance of brands. I wouldn't expect that to be common, but who knows :sweat_smile: 

> I should really get to https://github.com/zigpy/zha/issues/311, so we might be able to make this opt-in entirely?

#311 would be a nice improvement, but like the phase_b/c entities, I think it is fine to have it like this for now.

> Or just use a custom sensor for the Tuya quirks? But if we add the total_active_power sensor later, it's also not very ideal.

Since we can use a standard attribute for this, I am not a fan of having a custom Tuya sensor. But I don't have strong feelings so let me know if you prefer that.

TheJulianJES (approved):
Looks good. Thanks!

(Note: a new ZHA release with this PR should only be made when a quirks bump containing https://github.com/zigpy/zha-device-handlers/pull/4095 is merged. I'll merge this PR then.)

PR #460, opened 2025-05-29
Restore `latest_version` for update entity

abmantis:
If a version update was previously available but the user skipped it, setting the `latest_version` to be the same as the `installed_version` on startup will clear the `skipped_version` flag (see https://github.com/home-assistant/core/blob/dev/homeassistant/components/update/__init__.py#L459-L459 ).
After a while, when the new version is fetched and `latest_version` is set to the correct version, the skipped update will show up again.

Example:
- `installed_version`: 1.
- `latest_version`: 2, user skips the update.

After restart:
- `installed_version` is 1.
- `latest_version` is 1.
- `skipped_version` is gets cleared.

### DIFF
--- a/tests/data/devices/adurosmart-eria-ad-rgbw3001.json
+++ b/tests/data/devices/adurosmart-eria-ad-rgbw3001.json
@@ -1731,7 +1731,7 @@
           "installed_version": "0x00000001",
           "in_progress": false,
           "update_percentage": null,
-          "latest_version": "0x00000001",
+          "latest_version": null,
           "release_summary": null,
           "release_notes": null,
           "release_url": null
--- a/tests/data/devices/aqara-lumi-motion-ac01.json
+++ b/tests/data/devices/aqara-lumi-motion-ac01.json
@@ -1162,7 +1162,7 @@
           "installed_version": "0x00000035",
           "in_progress": false,
           "update_percentage": null,
-          "latest_version": "0x00000035",
+          "latest_version": null,
           "release_summary": null,
           "release_notes": null,
           "release_url": null
--- a/tests/data/devices/aug-winkhaus-gmbh-co-kg-fm-v-zb.json
+++ b/tests/data/devices/aug-winkhaus-gmbh-co-kg-fm-v-zb.json
@@ -1241,7 +1241,7 @@
           "installed_version": "0x00000011",
           "in_progress": false,
           "update_percentage": null,
-          "latest_version": "0x00000011",
+          "latest_version": null,
           "release_summary": null,
           "release_notes": null,
           "release_url": null
--- a/tests/data/devices/bosch-rbsh-mms-zb-eu.json
+++ b/tests/data/devices/bosch-rbsh-mms-zb-eu.json
@@ -3960,7 +3960,7 @@
           "installed_version": "0x11136760",
           "in_progress": false,
           "update_percentage": null,
-          "latest_version": "0x11136760",
+          "latest_version": null,
           "release_summary": null,
           "release_notes": null,
           "release_url": null
--- a/tests/data/devices/centralite-3405-l.json
+++ b/tests/data/devices/centralite-3405-l.json
@@ -1714,7 +1714,7 @@
           "installed_version": "0x10025310",
           "in_progress": false,
           "update_percentage": null,
-          "latest_version": "0x10025310",
+          "latest_version": null,
           "release_summary": null,
           "release_notes": null,
           "release_url": null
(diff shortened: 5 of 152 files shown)


TheJulianJES (approved):
This seems good.

I've noticed that after restarting HA, the initial state is always "Up-to-date". Then searching for updates switches a bunch of update entities from "Up-to-date" to "Unknown", when no provider is available for them. This should fix this as well.

On second thought, this might now cause update entities which have providers available switch from "Unknown" to "Up-to-date" when searching for updates, after restarting HA. This might not be that big of a deal, as always unskipping skipped firmware versions though.

I wonder if we should restore the update latest version after a HA restart (until it is overridden when searching for updates again)? Maybe we should check how other integrations deal with this.

TheJulianJES:
Yeah, Matter seems to restore of update related information in their [update entity code](https://github.com/home-assistant/core/blob/dev/homeassistant/components/matter/update.py) (`MatterUpdateExtraStoredData`).

I guess we could do that too, though I'd have to check if our current implementation easily allows for that.
Other than that, I think this is still a better bugfix than doing nothing (to fix the "skip update" feature).

abmantis:
Added restore of the latest_version.
https://github.com/home-assistant/core/pull/146400 is required for the attribute restore to work (but this PR should be merged first).

TheJulianJES:
Did we confirm that restoring just the latest version also allows the user to install the update immediately after restarting HA?
Like, is all the update data already available in zigpy then? Since we don't send an "image notify" for quite some time.

TheJulianJES:
The initial version of this PR (just setting `latest_version` to `None`) is definitely save, but I don't think this new implementation, with just restoring the latest version, will work correctly. 
Like, we don't save and restore the changelog and other stuff, so that will be missing.

Matter stores and restores a bunch more info:
https://github.com/home-assistant-libs/python-matter-server/blob/908c07cb8b0f831085c581498a41433720c034ec/matter_server/common/models.py#L228-L244

abmantis:
I've tested that the restore worked and the Update entity shows correctly: it will show as having an update right after reboot.
I have not tested triggering the update :/

abmantis:
@TheJulianJES is correct. Just tried it and I get `Failed to perform the action update/install. Version 604241926 is not available`. I'll look into adding the missing stuff.

puddly:
Ah, I see... I wasn't aware this would restore *pending* updates as well.

- We maybe should keep track of the information zigpy uses to match device OTA requests to OTA images: the entire `OTA:QueryNextImageCommand`: https://github.com/zigpy/zigpy/blob/ea966ca250276b9662aba256673647ad41268642/zigpy/zcl/clusters/general.py#L2007-L2023. With that, we have everything we need to "simulate" the device checking in. This would unfortunately result in us hitting the OTA provider upon integration reload to restore entity state, which may not be good.
- We then would need some sort of on-disk OTA cache for zigpy, to avoid having to download "pending" OTA images on startup.

This is definitely something we need to do but I'm not sure if we have time before the .0 release tomorrow to do this 😓

abmantis:
> Ah, I see... I wasn't aware this would restore _pending_ updates as well.
> 
>     * We maybe should keep track of the information zigpy uses to match device OTA requests to OTA images: the entire `OTA:QueryNextImageCommand`: https://github.com/zigpy/zigpy/blob/ea966ca250276b9662aba256673647ad41268642/zigpy/zcl/clusters/general.py#L2007-L2023. With that, we have everything we need to "simulate" the device checking in. This would unfortunately result in us hitting the OTA provider upon integration reload to restore entity state, which may not be good.
> 
>     * We then would need some sort of on-disk OTA cache for zigpy, to avoid having to download "pending" OTA images on startup.
> 
> 
> This is definitely something we need to do but I'm not sure if we have time before the .0 release tomorrow to do this 😓

The initial version of this PR was only fixing the wrong `latest_version` being set at startup, and that is good even without the restore. The restore part was added to the PR later.

Nevertheless, even if this goes out to the release as-is it is fine, because the code in HA core calling `restore_external_state_attributes` is not merged, so it is not being called. 

Today's a national holiday and I'll be out, but may be able to pick it up later tonight.

puddly:
> Today's a national holiday and I'll be out, but may be able to pick it up later tonight.

No worries, I'll test things on my home network right now and make sure there are no surprises.

TheJulianJES:
> I'm not sure if we have time before the .0 release tomorrow to do this 😓

Yeah, let's not do this for 2025.6. The current implementation with latest version being set to current version on startup isn't great, but nobody else has complained a lot.
We could do a minor change and set it to `None` for this release to fix the issue where ignored updates are unignored, but otherwise we shouldn't do anything for this release here. 

Ideally, zigpy itself could cache update info?

abmantis:
Opened a revert of the state restore code so we can still release the fix for the ignored-updates-not-working bug: https://github.com/zigpy/zha/pull/462

PR #461, opened 2025-06-10
Allow external radio types

puddly:
This will need a corresponding PR to Core (https://github.com/puddly/core/tree/puddly/zha-external-radio-libs) but will allow for pre-alpha radio libraries (Ziggurat, ZBOSS, BLZ, and ESPZB) to be used in Home Assistant.

You first need to modify your radio library's `pyproject.toml` file to register it with zigpy:

```toml
[project.entry-points."zigpy.radio"]
your_radio_type = "your_radio_module.zigbee.application:ControllerApplication"
```

And then just add a description to the `ControllerApplication` subclass:

```python
class ControllerApplication(zigpy.application.ControllerApplication):
    DISPLAY_NAME = "Friendly Name"
    DESCRIPTION = "Very short description string"

     def __init__(self, ...
```

Unfortunately, HA Core provides no simple way to install custom dependencies other than a custom component. I'll make a custom component later (if one doesn't already exist) so you can add something like this to `configuration.yaml`:
```yaml
# A custom component, unfortunately, to install dependencies
zha_custom_radio_libraries:
  - "git+https://github.com/zigpy/zigpy-ziggurat"
```

That's it. The Python module will self-register and show up in ZHA config flows, etc.

---

This is fundamentally a development tool so expect breakage if you rely on it 😄.


puddly:
I've slightly expanded the scope of this PR to clean up and entirely remove the `RadioType` enum. Instead, we can have structured information within the `Gateway` object accessible via an API call.

fangzheli:
Thanks @puddly for working on this — external radio type support would be a big deal for zigpy-blz.

We currently maintain a [custom ZHA component](https://github.com/bouffalolab/haos_custom_zha_blz) to patch BLZ radio support into ZHA, which requires tracking every HA Core release. This PR would let us drop that entirely and ship as a standalone `zigpy-blz` package (now [on PyPI](https://pypi.org/project/zigpy-blz/)).

We tested this PR locally with a Third Reality BL702 dongle — the Gateway successfully resolves and connects to zigpy-blz through `external_radio_libraries`. Happy to help with further testing as this moves forward.

Hedda:
Any status updates on this support to allow external radios? 

For reference also see this related PR for zigpy -> https://github.com/zigpy/zigpy/pull/1836

* https://github.com/zigpy/zigpy/pull/1836

puddly:
@Hedda I appreciate your enthusiasm but as you can see, this PR was updated just a few days ago. If there's any update, it will be here.

PR #462, opened 2025-06-10
Revert restoring update entity `latest_version`

abmantis:
Follow up to https://github.com/zigpy/zha/pull/460#issuecomment-2959795803 since restoring only the version is not enough to make the update work after a restart.


TheJulianJES (approved):
Thanks!

This (so setting `latest_version` to `None`) seems like the best short-term fix for ignored updated being unskipped.

PR #467, opened 2025-06-17
Add BinaryOutputCluster switch support

oddlama:
This PR adds support for BinaryOutputCluster as a switch. Currently, binary output clusters are correctly set up but not exposed into home assistant at all. Created switches look like this:

image

### DIFF
--- a/zha/application/platforms/switch.py
+++ b/zha/application/platforms/switch.py
@@ -145,6 +145,8 @@ class Switch(PlatformEntity, BaseSwitch):
 class BinaryOutputSwitch(PlatformEntity, BaseSwitch):
     """BinaryOutputCluster switch."""
 
+    _attr_translation_key = "switch"
+
     def __init__(
         self,
         cluster_handlers: list[ClusterHandler],
--- a/zha/application/platforms/switch.py
+++ b/zha/application/platforms/switch.py
@@ -11,7 +11,7 @@ from typing import TYPE_CHECKING, Any, cast
 from zhaquirks.quirk_ids import DANFOSS_ALLY_THERMOSTAT, TUYA_PLUG_ONOFF
 from zigpy.quirks.v2 import SwitchMetadata
 from zigpy.zcl.clusters.closures import ConfigStatus, WindowCovering, WindowCoveringMode
-from zigpy.zcl.clusters.general import OnOff, BinaryOutput
+from zigpy.zcl.clusters.general import OnOff
 from zigpy.zcl.foundation import Status
 
 from zha.application import Platform
@@ -37,7 +37,7 @@ from zha.zigbee.cluster_handlers.general import OnOffClusterHandler
 from zha.zigbee.group import Group
 
 if TYPE_CHECKING:
-    from zha.zigbee.cluster_handlers import ClusterHandler, BinaryOutputClusterHandler
+    from zha.zigbee.cluster_handlers import BinaryOutputClusterHandler, ClusterHandler
     from zha.zigbee.device import Device
     from zha.zigbee.endpoint import Endpoint
 
@@ -157,9 +157,9 @@ class BinaryOutputSwitch(PlatformEntity):
     ) -> None:
         """Initialize the switch."""
         super().__init__(cluster_handlers, endpoint, device, **kwargs)
-        self._binary_output_cluster_handler: BinaryOutputClusterHandler = self.cluster_handlers[
-            CLUSTER_HANDLER_BINARY_OUTPUT
-        ]
+        self._binary_output_cluster_handler: BinaryOutputClusterHandler = (
+            self.cluster_handlers[CLUSTER_HANDLER_BINARY_OUTPUT]
+        )
 
     def _is_supported(self) -> bool:
         if self._cluster_handler.description is None:
--- a/zha/application/platforms/switch.py
+++ b/zha/application/platforms/switch.py
@@ -145,8 +145,6 @@ class Switch(PlatformEntity, BaseSwitch):
 class BinaryOutputSwitch(PlatformEntity, BaseSwitch):
     """BinaryOutputCluster switch."""
 
-    _attr_translation_key = "switch"
-
     def __init__(
         self,
         cluster_handlers: list[ClusterHandler],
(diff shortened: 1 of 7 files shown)


puddly:
Thanks for making a PR for this feature!

For something that'll be used by DIY devices, I think it would be best to gate this behind the presence of a `description` attribute like we do for the `BinaryInput` cluster, like so:

```python
    def _is_supported(self) -> bool:
        if self._cluster_handler.description is None:
            return False

        return super()._is_supported()
```

This should get all of the failing entity unit tests passing as well.

oddlama:
> Thanks for making a PR for this feature!
> 
> For something that'll be used by DIY devices, I think it would be best to gate this behind the presence of a `description` attribute like we do for the `BinaryInput` cluster, like so:

Good point! I've added your suggested code.

zha/application/platforms/switch.py
```diff
@@ -140,6 +141,79 @@ def handle_cluster_handler_attribute_updated(
...
+        **kwargs: Any,
+    ) -> None:
+        """Initialize the switch."""
+        super().__init__(cluster_handlers, endpoint, device, **kwargs)
+        self._binary_output_cluster_handler: BinaryOutputClusterHandler = (
+            self.cluster_handlers[CLUSTER_HANDLER_BINARY_OUTPUT]
+        )
+
+    def _is_supported(self) -> bool:
+        if self._cluster_handler.description is None:
```
TheJulianJES: This should be `self._binary_output_cluster_handler`

zha/application/platforms/switch.py
```diff
@@ -140,6 +141,72 @@ def handle_cluster_handler_attribute_updated(
             self.maybe_emit_state_changed_event()
 
 
+@STRICT_MATCH(cluster_handler_names=CLUSTER_HANDLER_BINARY_OUTPUT)
+class BinaryOutputSwitch(PlatformEntity, BaseSwitch):
+    """BinaryOutputCluster switch."""
+
+    _attr_translation_key = "switch"
```
TheJulianJES: You're probably still working on this, but we should remove the translation key, as HA will automatically default to "Switch" [here](https://github.com/home-assistant/core/blob/42aaa888a1c1cacade3d97f82bd042756d916557/homeassistant/components/switch/strings.json#L21) (to "title" in line 2) if no custom name and no translation key are given.
This is generally preferred over duplicating translation keys. Do note that due to `description`, we'll likely never see this anyways.

We'll need to add an "exemption" in the test here: https://github.com/zigpy/zha/blob/b4f1386b5f38d7749f5b51489e2ffe3a49a2189b/tests/test_registries.py#L577-L583

TheJulianJES (commented):
~~Also noticed that the `async_update` implementation (for polling with `homeassistant.update_entity` service call) seems to be missing. Already the case for the `OnOff` variant. Only the `ConfigurableAttributeSwitch` has it (for config entities).

Just a reminder to add this when we also add it for `OnOff`. It's not that crucial, but we should still implement it in the future.~~

EDIT: Never mind, it's enough to implement `async_update` in the cluster handler and we magically call it from `PlatformEntity` if implemented: https://github.com/zigpy/zha/blob/b4f1386b5f38d7749f5b51489e2ffe3a49a2189b/zha/application/platforms/__init__.py#L461-L468

zha/application/platforms/switch.py
TheJulianJES: Actually, I'll have to double check if what I said with the "Switch" default is even true here.
In either case, the fallback name behavior isn't optimal and we'll need to adjust this (slightly?) at some point. Here, it shouldn't be a fallback name, but a (forced) name.

If we were to also set the translation key here though, we'd never see the fallback name, as we abort before reaching that part: https://github.com/home-assistant/core/blob/42aaa888a1c1cacade3d97f82bd042756d916557/homeassistant/components/zha/entity.py#L66-L74

Since we require a `description` to even create the entity, the current implementation is still good. We absolutely cannot set a translation key here, as we want the `description` (thus fallback name) visible.
But yeah, I'll have to take another look at this soon. It's a bit complicated.

zha/application/platforms/switch.py
```diff
@@ -140,6 +141,70 @@ def handle_cluster_handler_attribute_updated(
...
+                self.handle_cluster_handler_attribute_updated,
+            )
+        )
+
+    @property
+    def is_on(self) -> bool:
+        """Return if the switch is on."""
+        if self._binary_output_cluster_handler.present_value is None:
+            return False
+        return bool(self._binary_output_cluster_handler.present_value)
```
TheJulianJES, line 187: Now that we use `bool(.present_value)`, we should be able to remove the `present_value is None` check, since that will return `False`. That also fixes the coverage issue 😄

puddly: True. I was keeping it around for future review because I'm not sure if I agree with this behavior in general, it seems like `unavailable` would be a better state?

TheJulianJES: Yeah, I also thought about that. It makes sense for sensors and so on (and I think we unfortunately have this same behavior for binary sensors), but I'm not sure how this behaves with switches. Can you still turn on/off a switch if it's unavailable in HA?

Like, if it's an issue that `present_value` or `on_off` or whatever just wasn't read properly during pairing OR if there's some buggy device that just [makes zigpy] return `None` until the value was written once, do we really want to block doing that? (Assuming unavailable switch in HA means we can't turn on or off).

TheJulianJES (commented):
Since `async_update` is apparently already supported, since we magically call that method in the cluster handler from the base `PlatformEntity` [here](https://github.com/zigpy/zha/blob/b4f1386b5f38d7749f5b51489e2ffe3a49a2189b/zha/application/platforms/__init__.py#L461-L468), mind also copying the [`async_update` part from one of the OnOff switch tests](https://github.com/zigpy/zha/blob/b4f1386b5f38d7749f5b51489e2ffe3a49a2189b/tests/test_switch.py#L210-L220) to get coverage for that in the cluster handler (if you're not already working on that 😄)?

Otherwise, this looks good now.

TheJulianJES (approved):
nice!

PR #474, opened 2025-07-01
Remove empty attributes and duplicate keys in diagnostics JSON

puddly:
To reduce the amount of unnecessary clutter, attributes that are neither unsupported nor have a value will not be included in the JSON.


tests/common.py
```diff
@@ -397,11 +397,11 @@ def zigpy_device_from_device_data(
                 for attr in cluster["attributes"]:
                     attrid = int(attr["id"], 16)
 
-                    if attr["value"] is not None:
+                    if attr.get("value", None) is not None:
```
TheJulianJES, line 400: nit: I don't think we need to specify `None` here, as that should be the default return type for non-existent keys in dictionaries, but we can also keep it if that's more explicit or whatever.

PR #477, opened 2025-07-19
Initialize Inovelli VZM30-SN manufacturer specific attributes

codyhackw:
This PR adds the manufacturer specific attributes for the VZM30-SN so that the custom entities are created.

### DIFF
--- a/zha/zigbee/cluster_handlers/manufacturerspecific.py
+++ b/zha/zigbee/cluster_handlers/manufacturerspecific.py
@@ -271,7 +271,46 @@ class InovelliConfigEntityClusterHandler(ClusterHandler):
     def __init__(self, cluster: zigpy.zcl.Cluster, endpoint: Endpoint) -> None:
         """Initialize Inovelli cluster handler."""
         super().__init__(cluster, endpoint)
-        if self.cluster.endpoint.model == "VZM31-SN":
+        if self.cluster.endpoint.model == "VZM30-SN":
+            self.ZCL_INIT_ATTRS = {
+                "dimming_speed_up_remote": True,
+                "dimming_speed_up_local": True,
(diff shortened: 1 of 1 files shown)


TheJulianJES (commented):
If you can export device diagnostics with the latest v2 quirks and add the changes to this location: https://github.com/zigpy/zha/tree/dev/tests/data/devices, all these lines should be covered.
The tests will verify that ZHA creates the same entities that you see in your HA instance. However, as the v2 quirks aren't merged into ZHA yet, they'll fail until we merge a new quirks release. But you can already add the diagnostic/test files to this PR.

TheJulianJES (approved):
Ok, manually verified that all the attributes exist for the VZM30-SN.
We'll need somebody, who actually has this device, to get diagnostics in the future.

For now, I'll merge this without test coverage, since it's an important bugfix and a low-risk change.

TheJulianJES:
Got some diagnostics courtesy of @dmulcahey. I'll merge the quirks bump first, since that has the Inovelli quirks v2 stuff, then include the re-generated diagnostics based on the quirks bump here. 

EDIT: The diagnostics were for the VZM31-SN, not the VZM30-SN, my bad.

TheJulianJES (approved):
~~Diagnostics remaining unchanged, even though the missing initialization was added back, is due to the fact that the attributes were already all read once in the diagnostics we got.~~

Never mind, we added the -31 diagnostics, not the 30. Nobody has the VZM30-SN apparently, I'll just merge this without test coverage for now.

In the future, all/most of this will move into quirks v2 anyway, where all the attributes are automatically initialized by ZHA.

stephenjust:
I do have a VZM30-SN, submitted #495 to add the diagnostics file for it

Issue #364, opened 2025-01-27
Allow to specify number mode via quirks v2


TheJulianJES:
ZHA apparently doesn't use the quirks v2 provided modes for the `number` platforms.
Reference: https://github.com/zigpy/zha-device-handlers/pull/3719#issuecomment-2608651314

... because we apparently don't expose that via quirks v2. So, more of a feature request.

PR #482, opened 2025-07-26
Set number mode for quirks v2 entities

### LINKED ISSUE #202: Quirks v2 number entities ignoring mode
Mode parameter is missing when initializing quirk v2 number entities  

https://github.com/zigpy/zha/blob/2ca4115e7384c2e386b5b5091ba0ff729f88c077/zha/application/platforms/number/__init__.py#L248-L271

franortiz:
Initialize mode from quirks number metadata
Fixes zigpy/zha#202

### DIFF
--- a/tests/test_discover.py
+++ b/tests/test_discover.py
@@ -57,6 +57,7 @@ from zha.application.discovery import ENDPOINT_PROBE, EndpointProbe
 from zha.application.gateway import Gateway
 from zha.application.helpers import DeviceOverridesConfiguration
 from zha.application.platforms import PlatformEntity, binary_sensor, sensor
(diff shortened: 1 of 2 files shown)


TheJulianJES:
Huh, I managed to duplicate that issue in https://github.com/zigpy/zha/issues/364.
I do think we should probably include the `NumberMode` enum in "zigpy quirks v2", but we already have some v2 quirks setting it via string, so this PR might be fine to accept as is for now. Depending on how we implement it that, it could also be a breaking API change for few number entities.

TheJulianJES:
I've expanded an existing test to also verify the `NumberMode` is set correctly for number entities created from quirks v2 metadata. Overall, I think we can just go ahead with this implementation for now.

PR #483, opened 2025-07-28
Implement quirks V2 `change_entity_metadata` API

puddly:
See https://github.com/zigpy/zigpy/pull/1626.


zha/zigbee/device.py
```diff
@@ -868,6 +869,81 @@ def _is_entity_removed_by_quirk(self, entity: PlatformEntity) -> bool:
...
+        """Apply entity metadata changes from quirks v2."""
+        if self.quirk_metadata is None:
+            _LOGGER.debug("No quirk metadata for device %s", self)
+            return
+
+        _LOGGER.debug(
+            "Processing entity %s for metadata changes. Quirk metadata has %d change entries",
+            entity,
+            len(self.quirk_metadata.changed_entity_metadata),
+        )
```
TheJulianJES: Even though this is debug level, I could see this become a bit spammy, since one of the messages will be printed for **every entity**, even if there's no quirk applied or no change entries for a quirk. 

But yeah, it's debug level, so it's *probably* fine.

TheJulianJES (approved):
Looks good!

zha/zigbee/device.py
puddly: Ah, oops, that was for my own debugging 😄. This needs a bit of a cleanup before merging.

PR #495, opened 2025-07-31
Add test data for Inovelli VZM30-SN

stephenjust:
Also including fixes that were required to import the diagnostics files, in addition to #494


puddly:
Thanks! These test files are a big help, especially for these complex Inovelli devices.

CI is failing because the Inovelli devices were recently migrated to a different quirk format, you may need to run `python -m tools.regenerate_diagnostics` (with the ZHA venv dependencies fully up-to-date) to load and recreate them.

stephenjust:
> Thanks! These test files are a big help, especially for these complex Inovelli devices.
> 
> CI is failing because the Inovelli devices were recently migrated to a different quirk format, you may need to run `python -m tools.regenerate_diagnostics` (with the ZHA venv dependencies fully up-to-date) to load and recreate them.

I had to remove the diagnostics file for the VZM35-SN. It looks like the cluster id "3" on that device may or may not be present depending on firmware version. I suspect that quirk migration will need to be reverted. Will file a bug over there.

Edit: https://github.com/zigpy/zha-device-handlers/issues/4217

PR #498, opened 2025-08-01
Clean up electrical measurement sensors

### LINKED ISSUE #501: Strange warning logged with HASS 2025.8.0b1 about Unexpected extra state attributes found for sensor <zha.application.platforms.sensor.ElectricalMeasurementTotalActivePower...>
Hi,
while testing Home Assistant 2025.8.0b1 with the ne ZHA/zigpy stack this new warning sometimes appears. I have no idea, when exactly, but it is shown quite often. I have some devices with electrical power management cluster, but this message does not shwo which device is causing this.

```
2025-08-01 20:26:07.634 WARNING (MainThread) [homeassistant.components.zha.sensor] Unexpected extra state attributes found for sensor <zha.application.platforms.sensor.ElectricalMeasurementTotalActivePower object at 0x7f7097898350>: {'total_active_power_max'}
```

The message did not appear with 2025.7.3 and also not in the patched version I used for testing the SONOFF device and the new OnOfCluster code. So the problem might have been introduced later. Digging through commit logs.

TheJulianJES:
### Summary of the PR

This PR cleans up the electrical measurement sensors and removes an unexpected extra state attribute for the recently added "total active power" sensor.

Fixes https://github.com/zigpy/zha/issues/501

### Background

https://github.com/zigpy/zha/pull/458 added the "total active power" entity, but due to some magic we did in the base EM class, it also included an extra state attribute that HA doesn't expect (and which we should not have added).

This causes the following error in HA:
```python
Unexpected extra state attributes found for sensor <zha.application.platforms.sensor.ElectricalMeasurementTotalActivePower object at 0x7f791aece120>: {'total_active_power_max'}
```

Now, removing the extra state attribute just for this one sensor was slightly hacky, so since we may wanna land this during the beta period, I kept everything backwards compatible, so keeping the `_max` state attributes for all existing entities, yet still cleaned up everything *a lot* to prevent a bunch of inadvertent side effects when creating a new entity.

See this comment for some explanation on the changes in this PR: https://github.com/zigpy/zha/pull/498#discussion_r2246882172

### Future

In the future, we can remove all of the `_max` extra state attributes, since they're pretty much useless anyway, as no device supports them, yet we currently all poll them across phases A, B, C for current, power, and voltage, as well as one for frequency.

We may also want to apply a similar refactoring to the `SmartEnergy` sensors.

### Testing

Tests pass, which is good. The regenerated diagnostics are just due to a class name change.
Other than removing the (unpolled) extra state attribute for the "total active power" entity, that was mistakenly added in HA 2025.8.0b1, there should not be any behavioral changes.

So, this is just a bugfix, although it required a bit of refactoring first.

### DIFF
--- a/zha/application/platforms/sensor/__init__.py
+++ b/zha/application/platforms/sensor/__init__.py
@@ -665,15 +665,12 @@ class Battery(Sensor):
 class BaseElectricalMeasurement(PollableSensor):
     """Base class for electrical measurement."""
 
+    _attr_max_attribute_name: str | None = None
     _use_custom_polling: bool = False
-    _attribute_name = "active_power"
     _attr_suggested_display_precision = 1
-    _attr_device_class: SensorDeviceClass = SensorDeviceClass.POWER
+    _divisor_attribute_name: str | None = None
+    _multiplier_attribute_name: str | None = None
     _attr_state_class: SensorStateClass = SensorStateClass.MEASUREMENT
-    _attr_native_unit_of_measurement: str = UnitOfPower.WATT
-    _divisor_attribute_name: str | None = "ac_power_divisor"
-    _multiplier_attribute_name: str | None = "ac_power_multiplier"
-    _attr_max_attribute_name: str = None
 
     def __init__(
         self,
@@ -686,13 +683,8 @@ class BaseElectricalMeasurement(PollableSensor):
         super().__init__(cluster_handlers, endpoint, device, **kwargs)
         self._attr_extra_state_attribute_names: set[str] = {
             "measurement_type",
-            self._max_attribute_name,
+            self._attr_max_attribute_name,  # None values are ignored in HA
         }
-
-    @property
-    def _max_attribute_name(self) -> str:
-        """Return the max attribute name."""
-        return self._attr_max_attribute_name or f"{self._attribute_name}_max"
 
     @property
     def state(self) -> dict[str, Any]:
@@ -701,8 +693,7 @@ class BaseElectricalMeasurement(PollableSensor):
         if self._cluster_handler.measurement_type is not None:
             response["measurement_type"] = self._cluster_handler.measurement_type
 
-        max_attr_name = self._max_attribute_name
-        if not hasattr(self._cluster_handler.cluster.AttributeDefs, max_attr_name):
+        if (max_attr_name := self._attr_max_attribute_name) is None:
             return response
 
         if (max_v := self._cluster_handler.cluster.get(max_attr_name)) is not None:
@@ -733,13 +724,27 @@ class BaseElectricalMeasurement(PollableSensor):
         raise AttributeError("Cannot set divisor directly")
 
 
+# this entity will be created by ReportingEM or PolledEM class below
+class ElectricalMeasurementActivePower(BaseElectricalMeasurement):
+    """Active power phase measurement."""
+
+    _attribute_name = "active_power"
+    # no unique id suffix for backwards compatibility
+    # no translation key due to device class
+    _attr_max_attribute_name = "active_power_max"
+    _divisor_attribute_name = "ac_power_divisor"
+    _multiplier_attribute_name = "ac_power_multiplier"
+    _attr_device_class: SensorDeviceClass = SensorDeviceClass.POWER
+    _attr_native_unit_of_measurement: str = UnitOfPower.WATT
+
+
 @MULTI_MATCH(
     cluster_handler_names=CLUSTER_HANDLER_ELECTRICAL_MEASUREMENT,
     stop_on_match_group=CLUSTER_HANDLER_ELECTRICAL_MEASUREMENT,
     models={"VZM31-SN", "SP 234", "outletv4", "INSPELNING Smart plug"},
 )
-class ElectricalMeasurement(BaseElectricalMeasurement):
-    """Active power measurement."""
+class ReportingElectricalMeasurement(ElectricalMeasurementActivePower):
+    """Unpolled active power measurement."""
 
     pass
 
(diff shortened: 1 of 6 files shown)


zha/application/platforms/sensor/__init__.py
```diff
@@ -733,13 +724,27 @@ def _divisor(self, value: int | float | None) -> None:
...
+    """Active power phase measurement."""
+
+    _attribute_name = "active_power"
+    # no unique id suffix for backwards compatibility
+    # no translation key due to device class
+    _attr_max_attribute_name = "active_power_max"
+    _divisor_attribute_name = "ac_power_divisor"
+    _multiplier_attribute_name = "ac_power_multiplier"
+    _attr_device_class: SensorDeviceClass = SensorDeviceClass.POWER
+    _attr_native_unit_of_measurement: str = UnitOfPower.WATT
```
TheJulianJES, line 738: Explanation:
Previously, `PolledElectricalMeasurement` was the base class for everything, yet also the `active_power` sensor.

Now, the base class `BaseElectricalMeasurement` truly is just a base class. It doesn't set the attribute name to "active_power" as before and it also doesn't device device class + unit of measurement, which then needed to be overridden in all other EM sensor classes.

Due to how polling works for the EM sensors, we now have a `ElectricalMeasurementActivePower` class that does nothing on its own, as it's never created as a sensor for ZHA.
The entity is only created by either `ReportingElectricalMeasurement` (which does no polling) or by `PolledElectricalMeasurement`, which then polls all relevant EM attributes as before (using `async_update` in the EM cluster handler).

This generally allows for a much cleaner class hierarchy and we don't need to set `_use_custom_polling = False` in all other classes anymore.
It's still not as clean as we might like it to be, since the `PolledElectricalMeasurement` is responsible for handling the polling for all EM cluster attributes, but it's as clean as we can get with our current approach to this. And honestly, it's way better than before.

zha/application/platforms/sensor/__init__.py
```diff
@@ -686,23 +683,17 @@ def __init__(
...
         self._attr_extra_state_attribute_names: set[str] = {
             "measurement_type",
-            self._max_attribute_name,
+            self._attr_max_attribute_name,  # None values are ignored in HA
         }
 
-    @property
-    def _max_attribute_name(self) -> str:
-        """Return the max attribute name."""
-        return self._attr_max_attribute_name or f"{self._attribute_name}_max"
```
TheJulianJES, line 695: Explanation:
If a `_max` extra state attribute is wanted for an entity, it has to be explicitly specified using `_attr_max_attribute_name` now. This is done for all existing sensors, except the recently added "total active power" sensor, where the inclusion was inadvertent, as we do not even poll that `_max` attribute.

zha/application/platforms/sensor/__init__.py
```diff
@@ -907,7 +905,6 @@ class ElectricalMeasurementPowerFactorPhB(ElectricalMeasurementPowerFactor):
     _unique_id_suffix = "power_factor_ph_b"
     _attr_translation_key: str = "power_factor_ph_b"
     _skip_creation_if_no_attr_cache = True
-    _attr_max_attribute_name = "power_factor_max_ph_b"
```
TheJulianJES, line 910: Note: `power_factor_max_ph_b` is an attribute that did not even exist, so we can safely remove it.

zha/application/platforms/sensor/__init__.py
```diff
@@ -918,7 +915,6 @@ class ElectricalMeasurementPowerFactorPhC(ElectricalMeasurementPowerFactor):
     _unique_id_suffix = "power_factor_ph_c"
     _attr_translation_key: str = "power_factor_ph_c"
     _skip_creation_if_no_attr_cache = True
-    _attr_max_attribute_name = "power_factor_max_ph_c"
```
TheJulianJES, line 921: Note: `power_factor_max_ph_c` is an attribute that did not even exist, so we can safely remove it.

zha/application/platforms/sensor/__init__.py
```diff
@@ -686,23 +683,17 @@ def __init__(
         super().__init__(cluster_handlers, endpoint, device, **kwargs)
         self._attr_extra_state_attribute_names: set[str] = {
             "measurement_type",
-            self._max_attribute_name,
+            self._attr_max_attribute_name,  # None values are ignored in HA
```
TheJulianJES, line 686: ZHA in Home Assistant filters out `None` values for sensor extra state attributes. The behavior isn't changed, just adds a comment for an explanation.
Since we'll likely remove the `_max` extra state attributes soon, we can just remove this then as well.

zha/application/platforms/sensor/__init__.py
```diff
@@ -686,23 +683,17 @@ def __init__(
...
-
     @property
     def state(self) -> dict[str, Any]:
         """Return the state for this sensor."""
         response = super().state
         if self._cluster_handler.measurement_type is not None:
             response["measurement_type"] = self._cluster_handler.measurement_type
 
-        max_attr_name = self._max_attribute_name
-        if not hasattr(self._cluster_handler.cluster.AttributeDefs, max_attr_name):
```
TheJulianJES, line 705: Explanation:
Previously, `_max_attribute_name` returned the main sensor attribute name with a `_max` suffix if no explicit attribute name was provided, so we needed to check if the attribute even exists.

Now, we need to explicitly set the `_attr_max_attribute_name` for each sensor if we want to include an extra state attribute, hence we don't need to check if the attribute actually exists anymore.

(If we still manage to set `_attr_max_attribute_name` to an invalid attribute name, we probably want it to error out, instead of silently failing and not including the extra state attribute.)

TheJulianJES (commented):
I think this PR should be good to go. It fixes the root cause for the "extra state attribute" logging seen in 2025.8.0b1.
(Commit below just moved one line, without any functional changes.)

uschindler:
Message went away in HA 2025.8.0b2. Well done!

PR #503, opened 2025-08-04
Add extra state attributes to diagnostics JSON

puddly:
We currently have a small bug where a `None` state attribute name has made it into `extra_state_attributes`, causing downstream warnings. We should include the list of extra state attributes in the JSON to make sure changes like this are caught.

### DIFF
--- a/tests/data/devices/adurosmart-eria-ad-rgbw3001.json
+++ b/tests/data/devices/adurosmart-eria-ad-rgbw3001.json
@@ -429,7 +429,11 @@
           "off_with_transition": false,
           "off_brightness": null,
           "available": true
-        }
+        },
+        "extra_state_attributes": [
+          "off_brightness",
+          "off_with_transition"
+        ]
       }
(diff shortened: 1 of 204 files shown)


TheJulianJES (commented):
This is the extra state attribute names. We could also add the extra state attributes with values to diagnostics IMO. We'd have to get that from `state` per extra state attribute name defined.

This is already good/important of course, but might still be nice to have diagnostics stay close to what HA entities provide. I can also look at this in the next few days or so.

puddly:
I think they're (partially?) included in the info object's `state` dictionary. Maybe we should filter out unsupported ones?

TheJulianJES:
Oh, if they are present at all, it's fine. I hadn't checked. No need to filter out unsupported ones probably, as the names + their state is what HA constructs its extra state attributes out of.
So if either will have a diff in our diagnostics, HA will also have a difference.

All good then!

PR #505, opened 2025-08-04
Fix `None` extra state attribute names

TheJulianJES:
### Proposed change

This fixes an issue where we included `None` as an extra state attribute name on some EM sensors to HA, causing (only) an unexpected warning to be logged. With this PR, the `None` name is removed from the `_attr_extra_state_attribute_names`.

### Background

Before the EM sensors clean up PR (https://github.com/zigpy/zha/pull/498), we always provided HA the extra state attributes with the attribute name and a `_max` suffix, even if that attribute did not exist in ZCL at all.

For `power_factor`, where no max attribute exists, the property evaluated to `power_factor_max`. This was the code: <https://github.com/zigpy/zha/blob/51dda469ad6869146e75a2e402b65a733b17f2a4/zha/application/platforms/sensor/__init__.py#L687-L690>
In the new ZHA release, the code for that is basically the same, except that the max attribute can be `None`. 
I thought we filter stuff, so left the `None` value until we remove `_max` state attributes soon, as this matched previous behavior more closely. This is the part in HA responsible for extra state attributes: <https://github.com/home-assistant/core/blob/b76f47cd9ff3d14c84be44ac6cce74a5776f3675/homeassistant/components/zha/sensor.py#L151-L170>

But that should also have logged with `power_factor_max`, since that's set in the state attributes that are supposedly provided. Except, we apparently include `power_factor_max` in the extra state attributes in HA, even though it doesn't exist: <https://github.com/home-assistant/core/blob/b76f47cd9ff3d14c84be44ac6cce74a5776f3675/homeassistant/components/zha/sensor.py#L37-L77>

### HA Future TO-DO

We can later remove non-existent ZCL attributes like `power_factor_max` from HA: https://github.com/home-assistant/core/blob/b76f47cd9ff3d14c84be44ac6cce74a5776f3675/homeassistant/components/zha/sensor.py#L37-L77

### Previous PR

This is a follow-up PR to:
- https://github.com/zigpy/zha/pull/498

### DIFF
--- a/tests/test_sensor.py
+++ b/tests/test_sensor.py
@@ -377,6 +377,7 @@ async def async_test_em_rms_voltage(
     zha_gateway: Gateway, cluster: Cluster, entity: PlatformEntity
 ) -> None:
     """Test electrical measurement RMS Voltage sensor."""
+    assert entity.extra_state_attribute_names == {"measurement_type", "rms_voltage_max"}
 
     await send_attributes_report(zha_gateway, cluster, {0: 1, 0x0505: 1234})
     assert_state(entity, 123.4, "V")
--- a/tests/test_sensor.py
+++ b/tests/test_sensor.py
@@ -331,6 +331,8 @@ async def async_test_em_power_factor(
     zha_gateway: Gateway, cluster: Cluster, entity: PlatformEntity
 ):
     """Test electrical measurement Power Factor sensor."""
+    assert entity.extra_state_attribute_names == {"measurement_type"}
+
     # update divisor cached value
     await send_attributes_report(zha_gateway, cluster, {"ac_power_divisor": 1})
     await send_attributes_report(zha_gateway, cluster, {0: 1, 0x0510: 100, 10: 1000})
--- a/zha/application/platforms/sensor/__init__.py
+++ b/zha/application/platforms/sensor/__init__.py
@@ -683,8 +683,9 @@ class BaseElectricalMeasurement(PollableSensor):
         super().__init__(cluster_handlers, endpoint, device, **kwargs)
         self._attr_extra_state_attribute_names: set[str] = {
             "measurement_type",
-            self._attr_max_attribute_name,  # None values are ignored in HA
         }
+        if self._attr_max_attribute_name is not None:
+            self._attr_extra_state_attribute_names.add(self._attr_max_attribute_name)
 
     @property
     def state(self) -> dict[str, Any]:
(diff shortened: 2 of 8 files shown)


tests/test_sensor.py
```diff
@@ -331,6 +331,8 @@ async def async_test_em_power_factor(
     zha_gateway: Gateway, cluster: Cluster, entity: PlatformEntity
 ):
     """Test electrical measurement Power Factor sensor."""
+    assert entity.extra_state_attribute_names == {"measurement_type"}
```
TheJulianJES, line 334: This would have included {`measurement_type`, `None`} before this fix.

zha/application/platforms/sensor/__init__.py
```diff
@@ -683,8 +683,9 @@ def __init__(
         super().__init__(cluster_handlers, endpoint, device, **kwargs)
         self._attr_extra_state_attribute_names: set[str] = {
             "measurement_type",
-            self._attr_max_attribute_name,  # None values are ignored in HA
         }
+        if self._attr_max_attribute_name is not None:
+            self._attr_extra_state_attribute_names.add(self._attr_max_attribute_name)
```
TheJulianJES, line 688: I don't think there's any cleaner way to do when creating the set, but open to suggestions.

Only alternatives I can think of right now would be these (worse ones IMO):
```python
        self._attr_extra_state_attribute_names = set(
            filter(None, ["measurement_type", self._attr_max_attribute_name])
        )
```

```python
        self._attr_extra_state_attribute_names = {"measurement_type"} | (
            {self._attr_max_attribute_name}
            if self._attr_max_attribute_name is not None
            else set()
        )
```

```python
        self._attr_extra_state_attribute_names = {
            "measurement_type",
            *(
                {self._attr_max_attribute_name}
                if self._attr_max_attribute_name is not None
                else set()
            ),
        }
```

puddly:
This would have been caught by https://github.com/zigpy/zha/blob/5c933fcb028e32f6c2966241192633c8fa5ce1ea/pyproject.toml#L72. I'm going to see if it's possible to quickly remove that disable.

zha/application/platforms/sensor/__init__.py
puddly: I think the first `if` statement is fine, you can at least read it in one pass 😄

tests/test_sensor.py
```diff
@@ -375,6 +377,7 @@ async def async_test_em_rms_voltage(
     zha_gateway: Gateway, cluster: Cluster, entity: PlatformEntity
 ) -> None:
     """Test electrical measurement RMS Voltage sensor."""
+    assert entity.extra_state_attribute_names == {"measurement_type", "rms_voltage_max"}
```
TheJulianJES, line 380: We never checked if the behavior actually correctly include the "max extra state attributes" before, so also added this.

PR #506, opened 2025-08-04
Remove mypy `assignment` ignore

puddly:
It would have flagged https://github.com/zigpy/zha/pull/505.

This PR reshuffles types in a few platforms and slightly cleans up the `fan` platform. A lot of cluster handlers are pulled from the cluster handler dict via `.get` but very, very few are ever `None`, this has been made stricter.

Some TODOs:
- Group light entities directly use clusters instead of cluster handlers, which expose an identical command interface to `ClusterHandler`. This is pretty messy but required to utilize the `Light` base class logic.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,7 +69,6 @@ disable_error_code = [
     "attr-defined",
     "no-any-return",
     "no-untyped-call",
-    "assignment",
     "arg-type",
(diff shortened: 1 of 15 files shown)


puddly:
Everything runs fine for me at home, merging.

PR #507, opened 2025-08-04
Fix mypy `attr-defined` and `union-attr` ignores

puddly:
These are simple changes and should be safe to merge.

Only a few left to go:

| Error           | Count |
| --------------- | ----- |
| no-untyped-def  | 118   |
| no-any-return   | 104   |
| no-untyped-call | 50    |
| arg-type        | 40    |

### DIFF
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -545,7 +545,7 @@ class BaseLight(BaseEntity, ABC):
         """Process ZCL color commands."""
 
         if self._color_cluster_handler is None:
-            self.debug("no color cluster handler available")
+            self.warning("no color cluster handler available")
             return False
 
         transition_time = (
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -333,7 +333,7 @@ class BaseLight(BaseEntity, ABC):
 
         t_log = {}
 
-        if new_color_provided_while_off:
+        if new_color_provided_while_off and self._level_cluster_handler is not None:
             # If the light is currently off, we first need to turn it on at a low
             # brightness level with no transition.
             # After that, we set it to the desired color/temperature with no transition.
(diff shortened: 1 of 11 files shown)


zha/application/platforms/light/__init__.py
```diff
@@ -333,7 +333,7 @@ async def async_turn_on(self, **kwargs: Any) -> None:
 
         t_log = {}
 
-        if new_color_provided_while_off:
+        if new_color_provided_while_off and self._level_cluster_handler is not None:
```
TheJulianJES: Do we want to put this in `new_color_provided_while_off` variable above where the "brightness supported" is also checked?

zha/application/platforms/light/__init__.py
```diff
@@ -426,7 +427,7 @@ async def async_turn_on(self, **kwargs: Any) -> None:
                 self.debug("turned on: %s", t_log)
                 return
 
-        if new_color_provided_while_off:
+        if new_color_provided_while_off and self._level_cluster_handler is not None:
```
TheJulianJES: Then we also don't need it here again

zha/application/platforms/light/__init__.py
```diff
@@ -540,6 +544,10 @@ async def async_handle_color_commands(
     ):
         """Process ZCL color commands."""
 
+        if self._color_cluster_handler is None:
+            self.debug("no color cluster handler available")
```
TheJulianJES: We can probably change this to `warning` as well, similar to the warning for the fan cluster handler missing.
Realistically, we should never hit this if the light doesn't support colors/has a `Color` cluster (handler). If we do, we want a warning IMO.

zha/application/platforms/light/__init__.py
puddly: I've moved it into the `brightness_supported` flag but...mypy doesn't actually support narrowing like this. So this doesn't work:

```python
obj: SomeType | None = ...
flag = (obj is not None)

if flag:
    obj.method()  # Can't call this on possibly `None`
```

I've instead used `assert`s.

puddly:
Coverage is quite poor, we seemingly lack tests for colorloop and a few others.

PR #510, opened 2025-08-05
Reprioritize joining and initialization requests

puddly:
Requires https://github.com/zigpy/zigpy/pull/1635 to keep things more usable during startup. For now, this makes sure that device initialization takes precedence over most other requests.

### DIFF
--- a/zha/application/gateway.py
+++ b/zha/application/gateway.py
@@ -624,10 +624,8 @@ class Gateway(AsyncUtilMixin, EventBase):
         zha_device.available = True
         zha_device.on_network = True
 
-        async with self.request_priority(t.PacketPriority.CRITICAL):
+        async with self.request_priority(t.PacketPriority.HIGH):
             await zha_device.async_configure()
-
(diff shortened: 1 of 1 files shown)


TheJulianJES (approved):
This looks good to me. Do we want to get this in for the beta? (without https://github.com/zigpy/zigpy/pull/1635)

puddly:
Sure. Both should actually be okay to merge (I've been running the zigpy PR for a while without issues) but I suspect it'll cause errors for people who manually override the request concurrency to a number incompatible with the divisions.

TheJulianJES:
Hmm, yeah. We could change that to automatically correct to a smaller "biggest number that works" and log a warning, just so we can see how many people this affects – you never know how many people just paste stuff in their `configuration.yaml`. But it's *probably fine* as-is.
I'd go without that for this beta, since I'm not sure how much time I'll have this week to get to issues and so on. But if you want it in like how it is, ping me and I can include it in another bump for the beta.

PR #512, opened 2025-08-07
Fix IKEA remote cluster handler binding

puddly:
In Zigpy, we recently started being more strict with cluster types and routing packets to the correct cluster. For the vast majority of devices there is no ambiguity (one endpoint has one cluster of a specific type) but some devices implement both server and client clusters on the same endpoint.

IKEA remotes were one of those devices and the existing code to react to button events was relying on the old behavior, listening to the wrong cluster. I've added an explicit test for this device to make sure this doesn't regress.

https://github.com/home-assistant/core/issues/150148
https://github.com/home-assistant/core/issues/150179

### DIFF
--- a/zha/zigbee/cluster_handlers/manufacturerspecific.py
+++ b/zha/zigbee/cluster_handlers/manufacturerspecific.py
@@ -475,8 +475,8 @@ class IkeaAirPurifierClusterHandler(ClusterHandler):
 
 
 @registries.CLUSTER_HANDLER_ONLY_CLUSTERS.register(IKEA_REMOTE_CLUSTER)
-@registries.CLUSTER_HANDLER_REGISTRY.register(IKEA_REMOTE_CLUSTER)
-class IkeaRemoteClusterHandler(ClusterHandler):
+@registries.CLIENT_CLUSTER_HANDLER_REGISTRY.register(IKEA_REMOTE_CLUSTER)
+class IkeaRemoteClientClusterHandler(ClusterHandler):
     """Ikea Matter remote cluster handler."""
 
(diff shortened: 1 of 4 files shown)


zha/zigbee/cluster_handlers/manufacturerspecific.py
```diff
@@ -475,8 +475,8 @@ async def async_update(self) -> None:
 
 
 @registries.CLUSTER_HANDLER_ONLY_CLUSTERS.register(IKEA_REMOTE_CLUSTER)
-@registries.CLUSTER_HANDLER_REGISTRY.register(IKEA_REMOTE_CLUSTER)
-class IkeaRemoteClusterHandler(ClusterHandler):
+@registries.CLIENT_CLUSTER_HANDLER_REGISTRY.register(IKEA_REMOTE_CLUSTER)
+class IkeaRemoteClientClusterHandler(ClusterHandler):
```
TheJulianJES: Should this be a `ClientClusterHandler`? I guess it doesn't really need to be, as that only emits events IIRC

TheJulianJES: Looking at tests, we do seem to generally want client cluster handlers to be a subclass of `ClientClusterHandler`.

puddly: We have a unit test that fails if a client cluster handler isn't suffixed with `ClientClusterHandler` 😄

puddly: Ah, I forgot to actually make it one 🤦

PR #515, opened 2025-08-13
Fix SYMFONISK sound remote events

puddly:
Unit test event data was gathered from 2025.7.4.

It looks like there is a bit of a problem with events more broadly.

This device's custom IKEA remote client cluster was duplicated and made a server cluster when in reality it is only a client cluster (https://github.com/zigpy/zha-device-handlers/blob/48191cc6ca52b7ad9746d433251638f6966004ac/zhaquirks/ikea/symfonisk2.py#L156). The old zigpy ZCL matching logic preferred the server cluster and because we only create default cluster handlers (and thus listen to events) for *server* clusters, the emitted event worked.

Because we now route ZCL commands correctly, the event is emitted by the client cluster. Client cluster handlers, however, are not created unless explicitly registered. In this PR I have a commit that globally registers them but this isn't something that I think we can release mid-month. For now, I will manually add a client cluster handler for it. We should see if enabling client cluster handlers will cause any issues.

This requires a quirks PR to pass: https://github.com/zigpy/zha-device-handlers/pull/4257

CC https://github.com/home-assistant/core/issues/150360


zha/zigbee/cluster_handlers/const.py
```diff
@@ -84,6 +84,7 @@
 AQARA_OPPLE_CLUSTER: Final[int] = 0xFCC0
 IKEA_AIR_PURIFIER_CLUSTER: Final[int] = 0xFC7D
 IKEA_REMOTE_CLUSTER: Final[int] = 0xFC80
+IKEA_SHORTCUT_CLUSTER_V1_CLUSTER: Final[int] = 0xFC7F
```
TheJulianJES: nit: Can we rename this to `IKEA_SHORTCUT_V1_CLUSTER`, so we don't have `CLUSTER_V1_CLUSTER` at the end?

tests/test_device.py
```diff
@@ -1154,3 +1154,50 @@ async def test_somrig_events(zha_gateway: Gateway) -> None:
...
+        zha_gateway.application_controller,
+        "tests/data/devices/ikea-of-sweden-symfonisk-sound-remote-gen2.json",
+    )
+
+    zha_device = await join_zigpy_device(zha_gateway, zigpy_dev)
+
+    listener = mock.Mock()
+    zha_device.on_all_events(listener)
+
+    # ShortcutV1Cluster:shortcut_v1(shortcut_button=1, shortcut_event=1)
```
TheJulianJES: nit: I think the comment should be renamed back to the actual cmd name `shortcut_v1_events`?
```suggestion
    # ShortcutV1Cluster:shortcut_v1_events(shortcut_button=1, shortcut_event=1)
```

TheJulianJES (approved):
Looks good

PR #517, opened 2025-08-14
Entity recomputation and add/remove at runtime

puddly:
This PR introduces two new events to the ZHA device object:
- `DeviceEntityAddedEvent(unique_id: str)`
- `DeviceEntityRemovedEvent(unique_id: str)`

And a new method:
- `recompute_entities()`

`recompute_entities()` will emit the above events when entities are added/removed at runtime. I've added it to the main device discovery unit test to make sure that it's idempotent.

Future TODOs:
- Re-interview devices after OTA.
- Call `recompute_entities` on device rejoin to deal with devices that change modes.
- Call `recompute_entities` after device reconfiguration button is clicked.
- This is too expensive to call on every attribute update but it would be nice to figure out how to do this dynamically, to supersede the above manual calls?


TheJulianJES:
When HA receives a `DeviceEntityRemovedEvent`, do we want to completely remove the entity from the entity registry, like seen in the video below? Or do we only want to make it "no longer provided by ZHA", so the user can manually delete it?

HA does keep metadata, like names, icon, and so on for deleted entities since a year or so. So if it's re-added, it will be re-applied by HA. I guess it kind of depends on when this would happen? E.g. after a firmware update or re-interview?

https://github.com/user-attachments/assets/7ca17bef-1f76-4a55-94a9-de20592ab960

Last month, someone showed me how they reinterviewed a Z-Wave keypad which had 100+ entities. The old names were all "Value 1" and so on, but "worked". After the reinterview, all those entities became unavailable (likely "no longer provided") and completely new entities with proper names (and different unique IDs) were added.
They then had to manually delete them in HA (though you can use the entities list and select all unavailable ones).

I guess we may want to not completely remove the entity from the registry, for now? I'm not really sure... 
We probably only want to fully remove them if we're certain that the entity is no longer provided by the device?

HA does seem to prefer the pattern of completely removing the entity from the registry in these cases..

zha/zigbee/device.py
```diff
@@ -166,6 +168,27 @@ class DeviceFirmwareInfoUpdatedEvent:
...
+
+
+@dataclass(kw_only=True, frozen=True)
+class DeviceEntityRemovedEvent:
+    """Event generated when a new entity is added to a device."""
+
+    event_type: Final[str] = ZHA_DEVICE_ENTITY_REMOVED_EVENT
+    event: Final[str] = ZHA_DEVICE_ENTITY_REMOVED_EVENT
+
+    unique_id: str
```
TheJulianJES, line 198: The `unique_id` is only unique per platform in HA, so we should probably include that as well.

zha/zigbee/device.py
```diff
@@ -166,6 +168,27 @@ class DeviceFirmwareInfoUpdatedEvent:
...
 
+@dataclass(kw_only=True, frozen=True)
+class DeviceEntityAddedEvent:
+    """Event generated when a new entity is added to a device."""
+
+    event_type: Final[str] = ZHA_DEVICE_ENTITY_ADDED_EVENT
+    event: Final[str] = ZHA_DEVICE_ENTITY_ADDED_EVENT
+
+    # TODO: allow all entity information to be serialized and include it here
+    unique_id: str
```
TheJulianJES, line 187: Why would we need to do that?

zha/zigbee/device.py
```diff
@@ -1053,9 +1144,11 @@ async def on_remove(self) -> None:
                     exc_info=True,
                 )
 
-        for platform_entity in self._platform_entities.values():
+        for platform_entity in list(self._platform_entities.values()):
             try:
-                await platform_entity.on_remove()
+                # TODO: To avoid unnecessary traffic during shutdown, we don't
+                # need to emit an event for every entity, just the device
+                await self._remove_entity(platform_entity, emit_event=False)
```
TheJulianJES: Is there still something to do here?

TheJulianJES:
This PR rebased + adding `Platform` to added/removed events, as the `unique_id` is only unique per platform:
- https://github.com/zigpy/zha/compare/dev...TheJulianJES:tjj/puddly/dynamic-entities
- The changes are relatively small, so I can (force-) push them to this branch, if you want.

Very experimental implementation in HA:
- https://github.com/home-assistant/core/compare/dev...TheJulianJES:tjj/dynamic_entities
- The last commit can be reverted if we want to go back to completely removing it from the entity registry. That seems like what most integrations do, and makes removing a bit less awkward? Not sure what's better, yet.

puddly:
@TheJulianJES Sure, feel free to force push

zha/zigbee/device.py
puddly: It's been a while 😅.

For the server+client refactor, I think so that ZHA in Core can react to the "added" event and have all of the entity state ready to go. We then would send state changes as separate events, completing the entire entity lifecycle. Right now, ZHA in Core needs to peek into the entity object's state after it receives the "added" event.

zha/zigbee/device.py
puddly: I don't believe so, no. The original issue was event churn during reconfiguration and startup/shutdown (one event per entity, which was excessive). I think `emit_event=False` (or a event-catching context manager) is enough.

In the eventual client+server split, we maybe should see if it's worth consolidating these into a single "device added" message that includes all entity state or if it's not really a problem to really send state piecemeal with TCP socket buffering.

zha/zigbee/device.py
TheJulianJES: Ok, that makes sense. I think we can do that in a future PR with the refactor (and especially the `.state` stuff from https://github.com/zigpy/zha/pull/663). For now, I'd just leave this as is.

TheJulianJES:
I think an initial version of this should be ready to merge, so we can get the HA part ready soon and I can put up my re-interview PR for ZHA, so we may be able to get that all in for next month. 😅 

One change I've made (which might change again in the future) is to not emit the individual entity events on startup (first initialization) of the device, as HA already gets a signal to add entities from the "device fully initialized event".
This is somewhat similar to not emitting an event when removing the device, I guess.

zha/zigbee/device.py
```diff
@@ -976,53 +1003,131 @@ def _discover_new_entities(self) -> None:
...
+
+        entities = list(self._platform_entities.values())
+
+        # Remove all entities that are no longer supported
+        for entity in entities[:]:
+            entity.recompute_capabilities()
 
-        # At this point we can compute a primary entity
-        self._compute_primary_entity()
+            if not entity.is_supported() or not entity.is_supported_in_list(entities):
```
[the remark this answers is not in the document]
TheJulianJES: This was already the original PR and I think it's fine. We can revisit this later. `self` is filtered out in `is_supported_in_list`.

zha/zigbee/device.py
```diff
@@ -976,53 +1003,131 @@ def _discover_new_entities(self) -> None:
...
 
-        self._discover_new_entities()
+        if key in self._platform_entities:
+            raise ValueError(
+                f"Cannot add entity {entity!r}, unique ID already taken by {self._platform_entities[key]!r}"
+            )
 
-        await self._zdo_handler.async_initialize(from_cache)
-        self._zdo_handler.debug("'async_initialize' stage succeeded")
+        _LOGGER.debug("Discovered new entity %s", entity)
```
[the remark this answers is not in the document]
TheJulianJES: We're not consistent here, though I've changed this occasion, for now.
This also changes the logger class at the start, making you think the message comes from zigpy.

zha/zigbee/device.py
```diff
@@ -1061,9 +1166,11 @@ async def on_remove(self) -> None:
                     exc_info=True,
                 )
 
-        for platform_entity in self._platform_entities.values():
+        for platform_entity in list(self._platform_entities.values()):
             try:
-                await platform_entity.on_remove()
+                # XXX: To avoid unnecessary traffic during shutdown, we don't
+                # need to emit an event for every entity, just the device
+                await self._remove_entity(platform_entity, emit_event=False)
```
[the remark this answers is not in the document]
TheJulianJES: Explicitly cleared with [`20f225c` (this PR)](https://github.com/zigpy/zha/pull/517/commits/20f225c6503a42fe372ab931deec459560370aa1), though this shouldn't be easy to trigger.

TheJulianJES:
The "Comment[s] suppressed due to low confidence" from here is valid IMO: https://github.com/zigpy/zha/pull/517#pullrequestreview-4002720310

But I think we can't really fix unless we can determine whether the `primary` attribute was set via election or via a v2 quirk. I think all of these issues are valid, though not really new:

## Pre-existing issues flagged by Copilot (not introduced by this PR)

### `_compute_primary_entity` can get stuck after entity removal

Weight-election losers are set to `_attr_primary = False`, which permanently excludes them from future elections (filtered by `_attr_primary is not False`). If the winner is later removed, no remaining entity can become primary.

Fixing this requires distinguishing quirk-set `_attr_primary` from weight-computed values, since both use the same field. Needs a separate PR.

### `_discover_new_entities()` in `async_configure` leaves entities half-added

`async_configure()` calls `_discover_new_entities()`, which runs `entity.on_add()` (e.g. RSSI sensor registers a global updater listener) and queues entities in `_pending_entities` — but never calls `_add_pending_entities()`. These entities are cleaned up when `async_initialize()` runs later, but until then they have active side effects without being finalized.

### `is_supported_in_list` receives `self` in the entity list

`recompute_entities` passes the full entity list including the entity being evaluated. Implementations (RSSI sensor, button) already filter out `self` defensively. A prior commit on this branch explicitly ensured this works, but the caller could exclude `self` to avoid requiring defensive checks in every implementation.

TheJulianJES:
Everything seems to work fine (without any HA changes – just not getting new functionality then). I think we can merge this for now. I'll create an issue for the discovery issues mentioned above. Issue:
- https://github.com/zigpy/zha/issues/725

zha/zigbee/device.py
```diff
@@ -976,53 +1003,131 @@ def _discover_new_entities(self) -> None:
             entity.on_add()
             self._pending_entities.append(entity)
 
-    async def async_initialize(self, from_cache: bool = False) -> None:
-        """Initialize cluster handlers."""
-        self.debug("started initialization")
+    def _add_entity(self, entity: PlatformEntity, *, emit_event: bool = True) -> None:
+        """Add an entity to the device."""
```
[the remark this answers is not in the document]
TheJulianJES: See comment below and this issue:
- https://github.com/zigpy/zha/issues/725

This will be investigated at a later point.

PR #525, opened 2025-08-31
Add postfix number to ZHA entities

Markus98:
This is a PR created based on the discussion in https://github.com/home-assistant/core/pull/151434

## Proposed change
This PR implements a feature from the ZHA backlog. It adds an optional `postfix` attribute to ZHA entities. It is used to differentiate between entities that have the same type and name on the same device. The attribute is needed for proper numbered naming in the UI.

## Additional information
- Scroll all the way down for the relevant code files diff. Code files changed:
  - `zha/application/platforms/__init__.py`
  - `zha/zigbee/device.py`
- This PR is a part of closing issue: https://github.com/zigpy/backlog/issues/7
- I will update the hass-core PR with the required changes to use the `postfix` attribute https://github.com/home-assistant/core/pull/151434
- I did not write explicit unit tests for this, but regressions should be covered in testing since I added the `postfix` attribute to `BaseEntityInfo`.

## Checklist

- [x] The code change is tested and works locally.
- [x] Local tests pass. **Your PR cannot be merged unless tests pass**
- [x] There is no commented out code in this PR.
- [x] I have followed the development checklist
- [x] I have followed the perfect PR recommendations
- [x] The code has been formatted using Ruff
- [ ] Tests have been added to verify that the new code works.


TheJulianJES (commented):
Awesome, this looks pretty good!
I'll take a closer final look in the next few days, maybe adjust one of the comments, but I think this and the HA implementation are good otherwise.

One note for possible future PRs: it's a bit nicer to commit the regenerated diagnostics separately, since the GitHub diff viewer gets a bit laggy with all the changed files. We squash all PRs when merging them anyway, so it'll still be one commit in the repo.

TheJulianJES:
The Matter integration in Home Assistant has a similar feature. However, it uses the endpoint ID for the postfix. For reference, see: https://github.com/home-assistant/core/blob/180f898bfa8b4bc3a1a5a97f64866c64fe1f551e/homeassistant/components/matter/entity.py#L111

I would not do this in the Zigbee integration, as quite a few Zigbee devices start with endpoint 11 or some other number. This would be confusing if there's "Voltage (11)" and "Voltage (12)", for example.
Counting upwards from 1 (if there are duplicate entities) is a better approach, at least for Zigbee.
Directly creating entities (per device, not per endpoint, e.g. quirks v2 entities) also play a role here.

Minor side note:
The Matter implementation also sets its postfix before all entities "are discovered", just by looking at duplicate clusters. The reason for this might be that this was a good option which doesn't require checking other entities, or endpoint IDs are generally used more sensible with Matter.

For us, I think the approach in this PR is already the best.

epenet:
Have `translation_placeholders` been considered?

Similar to what I added to Tuya:
- https://github.com/home-assistant/core/pull/149248
- https://github.com/home-assistant/core/pull/149249
- https://github.com/home-assistant/core/pull/149250
- https://github.com/home-assistant/core/pull/149251
- https://github.com/home-assistant/core/pull/149664

TheJulianJES:
Thanks for linking this! 

I'm not sure that approach would work for us here, since we can potentially have almost every entity twice, needing a postfix. Or just once, not needing one.

In that case, we'd need two translation key/value pairs per current one, from what I can see, as we'd have a trailing space otherwise.

We could make use of it for some quirks v2 defined entities though, where e.g. left/right would be used instead. For these, the entities are defined/fixed per device, so we always know exactly which ones exist.
Translation placeholders could be useful here, though the postfix approach from this PR can be, too.

Markus98:
Thank you for taking a look and I'm glad that it seems good!

> One note for possible future PRs: it's a bit nicer to commit the regenerated diagnostics separately, since the GitHub diff viewer gets a bit laggy with all the changed files. We squash all PRs when merging them anyway, so it'll still be one commit in the repo.

Good point, thanks for the note. I'm used to working with atomic commits that do not break tests but I don't need to do that here if we squash anyways.

> I would not do this in the Zigbee integration, as quite a few Zigbee devices start with endpoint 11 or some other number. This would be confusing if there's "Voltage (11)" and "Voltage (12)", for example.
> Counting upwards from 1 (if there are duplicate entities) is a better approach, at least for Zigbee.

I made the same observation and agree. 

> Have `translation_placeholders` been considered?

I also considered implementing it this way, but decided against it for the same reasons @TheJulianJES outlined.

epenet:
I don't know if you have thought of this:
- 3 "production" (kWh) sensors (total / A / B)
- 3 "power" (W) sensors (total / A  / B)
- only 2 "current" (A) sensors (A / B)

With the automatic suffix, you end up with misalignment of the suffix (`current B` would end up with same suffix 2 as `power A`)

Is it really that bad to use the cluster ID?
If we don't want to start number from say 11, we would use a hex string of the cluster ID so that we at least keep the "grouping" of the cluster entities.

Markus98:
@epenet I would argue that those cases should be solved via quirks for the devices that need them. From what I saw, it is quite common that the cluster ID starts with 11 or some other number that isn't as nice to use as a postfix directly. I assume that the case you pointed out is rarer, and I would say that it's more important for this to work cases which are more common.

My opinion is mostly based on looking at the mock device test data, so feel free to disagree if you have better knowledge on what is most common.

TheJulianJES:
> I don't know if you have thought of this:
> - 3 "production" (kWh) sensors (total / A / B)
> - 3 "power" (W) sensors (total / A  / B)
> - only 2 "current" (A) sensors (A / B)

I've never encountered any real devices that provide entities like this, but one solution to this could be the following:
- internally use cluster IDs for counting first
- then, go over all IDs in ascending order and make them start with 1

So, for these entity names (with cluster id), we'd get these names:
- Power (11) -> Power (1)
- Power (12) -> Power (2)
- Power (13) -> Power (3)
- Current (12) -> Current (2)
- Current (13) -> Current (3)

zha/zigbee/device.py
```diff
@@ -946,6 +946,30 @@ def _discover_new_entities(self) -> None:
...
+            tuple[type[BaseEntity], str | None], list[BaseEntity]
+        ] = defaultdict(list)
+        # Group entities by type & name, pick groups that are greater than one entity,
+        # and add postfix numbers to the entities of those groups.
+        for entity in entities:
+            # Pick the name key in the same priority order as which attribute affects the name.
+            name_key = (
+                entity.translation_key or entity.device_class or entity.fallback_name
+            )
+            entity_type_name_dict[(type(entity), name_key)].append(entity)
```
puddly: `type(entity)` would theoretically group entities with the same "outward appearance" (e.g. `name_key`) apart from one another if they do not share the exact same type. What about using the platform as a key instead?

Markus98: Sounds logical to me. I implemented the change and it looks like it affected a couple of devices in the test diagnostics dataset.

Markus98:
We had a short discussion about this in a meeting with the HA core team. With the new entity naming changes coming in, we should probably look at implementing sub-devices instead of the postfix number for the name.

Listing the large protocol integrations and making sure they migrate to sub-devices instead is already a task on the roadmap. ZHA falls under this umbrella.
https://github.com/home-assistant/core/issues/168901

So implementing this PR could be considered an antipattern. We should probably reconsider if we want to go forward with this if the sub-device pattern is adopted. I also am aware that matter already implements this pattern, but that one needs to be figured out separately.

Marking this as draft until a decision has been made.

Markus98:
I am wondering whether we could use the logic in this PR as-is for creating sub-devices in HA instead of using the numbers for postfixes.

Related task:
https://github.com/home-assistant/core/issues/172450

TheJulianJES:
Hmm, maybe. I feel like creating sub-devices per endpoint might be too much, but creating ones only if there's duplicate names could be too little? What about the entities which aren't duplicated then, do they all land on the first sub-device, or on the endpoint they're actually on?

It is definitely worth investigating. We have a lot of devices in our diagnostics database, so I guess we can implement some sort of index and output that to the ZHA diagnostics, then run the different index/sub-device calculations to see which one works best.

Markus98:
I looked at what I implemented here and the logic is very much tied to naming, which we can discuss whether it is a valid strategy for creating sub devices or not. But anyways, I think this PR can serve as reference for implementations, but we should create separate PRs for them. For that reason I am closing this one.

PR #533, opened 2025-09-12
Add support for translation_placeholders

epenet:
~Needs https://github.com/zigpy/zigpy/pull/1668~


zha/application/platforms/__init__.py
```diff
@@ -205,6 +207,13 @@ def translation_key(self) -> str | None:
             return self._attr_translation_key
         return None
 
+    @property
+    def translation_placeholders(self) -> dict[str, str] | None:
+        """Return the translation placeholders."""
+        if hasattr(self, "_attr_translation_placeholders"):
+            return self._attr_translation_placeholders
```
TheJulianJES: The attribute should always be present, though we also check this for `translation_key` and `_attr_entity_category`. I think the `hasattr` check can be removed for all off them. Not sure why it's there in the first place. But we can also clean up the other attributes in another PR.

TheJulianJES (commented):
Otherwise, this looks good to me as well.

zha/application/platforms/__init__.py
```diff
@@ -205,6 +207,13 @@ def translation_key(self) -> str | None:
...
+        return None
```
TheJulianJES: Removing the `hasattr` check per my comment here: https://github.com/zigpy/zha/pull/533#discussion_r2361297687
```suggestion
        return self._attr_translation_placeholders
```

puddly:
I've finished tests for this PR and added a missing branch for directly-defined `translation_placeholders` (as opposed to `new_translation_placeholders` being redefined by a quirk).

TheJulianJES (approved):
Looks good!

PR #548, opened 2025-10-22
Fix not claiming cluster handler for some quirks v2 entities

TheJulianJES:
## Proposed change

This fixes an issue where ZHA doesn't claim the cluster handler for quirks v2 entities that only require the specified attribute to be read during device initialization (pairing), but not bound, causing the cluster handler to not be configured, thus never reading the required attributes, leading to "unknown" state for affected quirks v2 entities after initial pairing.

Two new tests are also added to (1) test that the cluster handler is claimed in the above mentioned scenario, where no binding is configured, and (2) to test that setting up a quirks v2 "command button" entity doesn't claim the cluster handler unnecessarily.

### Note

We currently claim cluster handlers for all quirks v2 metadata that contains `attribute_name`, which I believe basically all do, except for command button. We don't really need to claim a cluster handler + read the attribute for "write attribute buttons" though. We may wanna change this behavior in this PR.

## Additional information

For more information on the underlying issues this fixes, see: https://github.com/zigpy/backlog/issues/52

### Note on tests

The tests may be a bit misplaced in `test_sensor` and `test_button`. They do (need to) utilize quirks v2 entities, but it's really the underlying (quirks v2 device) discovery that is tested. 

For now, I'd leave them in-place to minimize the diff, and they do test quirks v2 entities somewhat. In the future, we may wanna move the tests though.

### DIFF
--- a/tests/test_button.py
+++ b/tests/test_button.py
@@ -221,7 +221,7 @@ async def custom_button_device(zha_gateway: Gateway):
         .command_button(
             FakeManufacturerCluster.ServerCommandDefs.self_test.name,
             FakeManufacturerCluster.cluster_id,
-            command_args=(5,),
+            command_kwargs={"identify_time": 5},
             translation_key="self_test",
             fallback_name="Self test",
         )
@@ -387,7 +387,7 @@ async def test_cluster_handler_quirks_unnecessary_claiming(
         .command_button(
             FakeManufacturerCluster.ServerCommandDefs.self_test.name,
             OppleCluster.cluster_id,
-            command_args=(5,),
+            command_kwargs={"identify_time": 5},
             translation_key="self_test",
             fallback_name="Self test",
         )
--- a/tests/test_button.py
+++ b/tests/test_button.py
@@ -221,7 +221,7 @@ async def custom_button_device(zha_gateway: Gateway):
         .command_button(
             FakeManufacturerCluster.ServerCommandDefs.self_test.name,
             FakeManufacturerCluster.cluster_id,
-            command_kwargs={"identify_time": 5},
+            command_args=(5,),
             translation_key="self_test",
             fallback_name="Self test",
         )
@@ -268,7 +268,7 @@ async def test_quirks_command_button(
         assert len(cluster.request.mock_calls) == 1
         assert cluster.request.call_args[0][0] is False
         assert cluster.request.call_args[0][1] == 0
-        assert cluster.request.call_args[0][3] == 5  # duration in seconds
+        assert cluster.request.call_args[0][3] == 5
 
 
 async def test_quirks_write_attr_button(
--- a/tests/test_button.py
+++ b/tests/test_button.py
@@ -268,7 +268,7 @@ async def test_quirks_command_button(
         assert len(cluster.request.mock_calls) == 1
         assert cluster.request.call_args[0][0] is False
         assert cluster.request.call_args[0][1] == 0
-        assert cluster.request.call_args[0][3] == 5
+        assert cluster.request.call_args[0][3] == 5  # duration in seconds
 
 
 async def test_quirks_write_attr_button(
--- a/tests/test_button.py
+++ b/tests/test_button.py
@@ -391,14 +391,6 @@ async def test_cluster_handler_quirks_unnecessary_claiming(
             translation_key="self_test",
             fallback_name="Self test",
         )
-        # TODO: we claim the cluster handler + read attributes for write_attr_button...
-        # .write_attr_button(
-        #     "last_feeding_size",
-        #     20,
-        #     OppleCluster.cluster_id,
-        #     translation_key="last_feeding_size",
-        #     fallback_name="Last Feeding Size",
-        # )
         .add_to_registry()
     )
 
--- a/tests/test_button.py
+++ b/tests/test_button.py
@@ -385,7 +385,7 @@ async def test_cluster_handler_quirks_unnecessary_claiming(
         )
         .replaces(OppleCluster)
         .command_button(
-            FakeManufacturerCluster.ServerCommandDefs.self_test.name,
+            OppleCluster.ServerCommandDefs.self_test.name,
             OppleCluster.cluster_id,
             command_kwargs={"identify_time": 5},
             translation_key="self_test",
--- a/tests/test_button.py
+++ b/tests/test_button.py
@@ -12,6 +12,7 @@ from zhaquirks.const import (
     PROFILE_ID,
 )
 from zhaquirks.tuya.tuya_valve import ParksideTuyaValveManufCluster
+import zigpy
 from zigpy.exceptions import ZigbeeException
 from zigpy.profiles import zha
 from zigpy.quirks import CustomCluster, CustomDevice, DeviceRegistry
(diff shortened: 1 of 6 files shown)


tests/test_button.py
```diff
@@ -354,3 +357,89 @@ async def test_quirks_write_attr_buttons_uid(zha_gateway: Gateway) -> None:
...
+    registry = DeviceRegistry()
+    (
+        QuirkBuilder(
+            "Fake_Manufacturer_sensor_2", "Fake_Model_sensor_2", registry=registry
+        )
+        .replaces(OppleCluster)
+        .command_button(
+            FakeManufacturerCluster.ServerCommandDefs.self_test.name,
+            OppleCluster.cluster_id,
+            command_args=(5,),
```
puddly: Tiny nitpick:
```suggestion
            command_kwargs={"identify_time": 5},
```

tests/test_button.py
```diff
@@ -354,3 +357,89 @@ async def test_quirks_write_attr_buttons_uid(zha_gateway: Gateway) -> None:
...
+            translation_key="self_test",
+            fallback_name="Self test",
+        )
+        .add_to_registry()
+    )
+
+    zigpy_device = create_mock_zigpy_device(
```
puddly, line 397: Another nitpick, feel free to ignore if it's too much trouble. Is there any way to test using JSON diagnostics from a real device? I'd like to phase out use of `create_mock_zigpy_device` and other synthetic testing setups in favor of using real devices.

zha/application/discovery.py
```diff
@@ -299,6 +306,19 @@ def discover_quirks_v2_entities(self, device: Device) -> Iterator[PlatformEntity
                     [cluster_handler.name],
                 )
 
+            # if the cluster handler is unclaimed, claim it and set BIND accordingly,
```
puddly, line 309: I wonder: should we move this to happen before the `yield`? That way, the cluster handler is fully set up before the entity initialization happens.

zha/application/discovery.py
```diff
@@ -299,6 +306,19 @@ def discover_quirks_v2_entities(self, device: Device) -> Iterator[PlatformEntity
...
+            # so ZHA configures the cluster handler: reporting + reads attributes
+            if (attribute_initialization_found or reporting_found) and (
+                cluster_handler not in endpoint.claimed_cluster_handlers.values()
```
puddly, line 312: Is there any way for the cluster handler to already be claimed?

puddly (approved):
Looks good to me! I've left a few comments.

Long-term, I would like to explore a way to get rid of cluster handlers. Maybe we can have class-based entities (internal to ZHA) provide a nicer way to set up binding and reporting config.

tests/test_button.py
TheJulianJES: Hmm, I explicitly used `create_mock_zigpy_device`, as that's what the similar tests in `test_sensor` used and especially here, we really want to make sure that we get this exact device with that exact cluster setup.

I feel like we always run into a bit of a risk with using JSON diagnostics in tests, as it's not clear how endpoints and clusters are layed out by just looking at the tests. You have to look at the (huge) JSON file. It's also harder to get a device exactly in the way you want it, without adding unnecessary endpoints/clusters/quirk interference.

Some parts of the device/diagnostics can change over time as well (i.e. ZHA or quirk doing something different, firmware update, ...), when we may not want that for all tests where the diagnostics are used in.

zha/application/discovery.py
TheJulianJES: We can't really do that in a nice way, since we need to iterate over all quirks v2 entity metadata for a specific cluster handler before we can make the final decision on whether we can set `BIND` to `False`.

This is all still done "pre initialization", since they're just added to `Device._pending_entities` in discovery here: https://github.com/zigpy/zha/blob/b17254948690f740a2b795d28dfad5304ad98291/zha/zigbee/device.py#L939-L950
Nothing in `entity.on_add()` can also ever rely on the cluster handler being claimed or not.
The actual cluster handler initialization is done *way* later, where it then matters if a cluster handler is claimed or not. I think it's fine to keep it like this tbh.

`self._discover_new_entities()` does everything we do here. Only afterwards is `endpoint.async_initialize` called:
https://github.com/zigpy/zha/blob/b17254948690f740a2b795d28dfad5304ad98291/zha/zigbee/device.py#L956-L969

zha/application/discovery.py
TheJulianJES: The cluster handler can already be claimed by entities from `EndpointProbe`. Since those ZHA entities may rely on the default of `BIND = True`, I've added the `cluster_handler not in endpoint.claimed_cluster_handlers.values()` check to make sure we only ever set `BIND = False` for cluster handlers only claimed for quirks v2 entities.

This is also checked by the tests.

tests/test_button.py
TheJulianJES: Good catch. Copied that from some other test, but I'll change that.
We'll probably want to adjust this in a lot of existing tests as well at a later date, including (completely) using kwargs for creating quirks v2 entities.

TheJulianJES:
> Long-term, I would like to explore a way to get rid of cluster handlers. Maybe we can have class-based entities (internal to ZHA) provide a nicer way to set up binding and reporting config.

Yeah, I think they're not *too bad*, but they could definitely use some improvement (or yeah, possibly be removed entirely).

For quirks v2, we now automatically handle setting up cluster handlers under the hood (claiming + attribute initialization + reporting), so we can definitely change how we do that, without breaking the quirks v2 API.
As a first step, we might be able to introduce something similar to that for the "ZHA entities", so that relevant attributes are automatically read and configured from reporting, just by defining the entity alone.

tests/test_button.py
TheJulianJES: `FakeManufacturerCluster.ServerCommandDefs.self_test.name` should also be changed to `OppleCluster[...]`. It's the same command name and we never execute this, since we just need it for testing the quirks v2 metadata processing in the device discovery, but it should still be adjusted.

TheJulianJES: Done with [`6c12760` (#548)](https://github.com/zigpy/zha/pull/548/commits/6c127608e951476e5fd41ea7417e29e0f34b13ee), [`51682e9` (#548)](https://github.com/zigpy/zha/pull/548/commits/51682e9e648ee45a4189ce487d2e61b00efcab44), and [`dfa5653` (#548)](https://github.com/zigpy/zha/pull/548/commits/dfa5653527d5219404d8e4725d5f49317a523e70).
I'll fix the existing command button test in a follow-up PR. I don't want to include those unrelated changes here.

tests/test_button.py
TheJulianJES: We can address this in the future IMO. The new/modified tests are all based on one very similar (existing) test. If so, we should change all of them then.

Maybe we can have some JSON diagnostics files in a separate folder that aren't real devices (and cannot be "modified" if we add new diagnostics). That way we could avoid `create_mock_zigpy_device`.

We could add a comment in the tests then to explain what the diagnostics/tests devices do.
I do think that manually modifying JSON files isn't as easy as using `create_mock_zigpy_device`. For these tests, `create_mock_zigpy_device` is kind of ideal.

TheJulianJES:
I think this should be mergeable for now. Answered your comments above, including [one more comment](https://github.com/zigpy/zha/pull/548#discussion_r2470476610) on `create_mock_zigpy_device`. Let me know if you have more feedback/questions.

puddly (approved):
Looks good to me. Thanks!

PR #553, opened 2025-10-25
Don't await for identify effect on join

abmantis:
Some devices do not reply to identify requests. For those, the message will get sent 3 times and timeout after 15 seconds.
When the request times-out, the join process will (silently) be interrupted with a asyncio.exceptions.CancelledError, keeping the join stuck at "Configuring".

Catching the exception would solve the issue, but it would make the join unecessarily long. Since identify is not critical this PR moves it to a separate fire-and-forget task.

Partly addresses https://github.com/zigpy/backlog/issues/58

### DIFF
--- a/zha/zigbee/device.py
+++ b/zha/zigbee/device.py
@@ -840,9 +840,13 @@ class Device(LogMixin, EventBase):
             and self.identify_ch is not None
             and not self.skip_configuration
         ):
-            await self.identify_ch.trigger_effect(
-                effect_id=Identify.EffectIdentifier.Okay,
-                effect_variant=Identify.EffectVariant.Default,
+            self._gateway.async_create_task(
+                self.identify_ch.trigger_effect(
+                    effect_id=Identify.EffectIdentifier.Okay,
+                    effect_variant=Identify.EffectVariant.Default,
+                ),
+                name=f"({self.nwk},{self.model}) trigger_effect identify",
+                eager_start=True,
             )
 
     def _is_entity_removed_by_quirk(self, entity: PlatformEntity) -> bool:


puddly (approved):

puddly:
We should figure out eventually what is causing the cancellation to happen or propagate. It isn't an exception we throw in zigpy.

zha/zigbee/device.py
```diff
@@ -840,9 +840,13 @@ async def async_configure(self) -> None:
             and self.identify_ch is not None
             and not self.skip_configuration
         ):
-            await self.identify_ch.trigger_effect(
-                effect_id=Identify.EffectIdentifier.Okay,
-                effect_variant=Identify.EffectVariant.Default,
+            self._gateway.async_create_task(
```
TheJulianJES, line 843: nit: At some point in the future, we may want to decide whether to use the `self._gateway` attribute directly or the `self.gateway` property 😄 . There's seemingly no pattern to how they're used in this class.

TheJulianJES (approved):
Nice catch!

abmantis:
> We should figure out eventually what is causing the cancellation to happen or propagate. It isn't an exception we throw in zigpy.

Catching and logging it, this is the trace:

```
Traceback (most recent call last):
  File "/home/acosta/dev/repos/homeassistant/core/dev/.venv/lib/python3.13/site-packages/zigpy/device.py", line 603, in request
    return await future
           ^^^^^^^^^^^^
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/acosta/dev/repos/homeassistant/core/dev/.venv/lib/python3.13/site-packages/zha/zigbee/cluster_handlers/__init__.py", line 68, in wrap_zigpy_exceptions
    yield
  File "/home/acosta/dev/repos/homeassistant/core/dev/.venv/lib/python3.13/site-packages/zha/zigbee/cluster_handlers/__init__.py", line 88, in wrapper
    return await RETRYABLE_REQUEST_DECORATOR(func)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/acosta/dev/repos/homeassistant/core/dev/.venv/lib/python3.13/site-packages/zigpy/util.py", line 133, in retry
    return await func()
           ^^^^^^^^^^^^
  File "/home/acosta/dev/repos/homeassistant/core/dev/.venv/lib/python3.13/site-packages/zigpy/zcl/__init__.py", line 401, in request
    return await self._endpoint.request(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<9 lines>...
    )
    ^
  File "/home/acosta/dev/repos/homeassistant/core/dev/.venv/lib/python3.13/site-packages/zigpy/endpoint.py", line 234, in request
    return await self.device.request(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<11 lines>...
    )
    ^
  File "/home/acosta/dev/repos/homeassistant/core/dev/.venv/lib/python3.13/site-packages/zigpy/device.py", line 602, in request
    async with asyncio_timeout(timeout):
               ~~~~~~~~~~~~~~~^^^^^^^^^
  File "/home/acosta/.local/share/uv/python/cpython-3.13.3-linux-x86_64-gnu/lib/python3.13/asyncio/timeouts.py", line 116, in __aexit__
    raise TimeoutError from exc_val
TimeoutError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/acosta/dev/repos/homeassistant/core/dev/.venv/lib/python3.13/site-packages/zha/zigbee/device.py", line 841, in async_configure
    await self.identify_ch.trigger_effect(
    ...<2 lines>...
    )
  File "/home/acosta/dev/repos/homeassistant/core/dev/.venv/lib/python3.13/site-packages/zha/zigbee/cluster_handlers/__init__.py", line 87, in wrapper
    with wrap_zigpy_exceptions():
         ~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/acosta/.local/share/uv/python/cpython-3.13.3-linux-x86_64-gnu/lib/python3.13/contextlib.py", line 162, in __exit__
    self.gen.throw(value)
    ~~~~~~~~~~~~~~^^^^^^^
  File "/home/acosta/dev/repos/homeassistant/core/dev/.venv/lib/python3.13/site-packages/zha/zigbee/cluster_handlers/__init__.py", line 70, in wrap_zigpy_exceptions
    raise ZHAException("Failed to send request: device did not respond") from exc
```

zha/zigbee/device.py
abmantis: Ah, didn't even noticed the property! I just copy/pasted this from some other place in the file :sweat_smile: 
Should I change it?

puddly:
I wonder if this is caused by `Task.cancel()`? If we await a task that gets cancelled (e.g. if a device rejoins and we cancel outstanding requests), this seems like it would inadvertently propagate `CancelledError`.

PR #557, opened 2025-10-28
Diagnostics import for October

puddly:
The previous script I was using for this task (TODO: clean it up and move it here to be run by CI) had a bug and missed diagnostics JSON present in the issue body, picking up only those in comments.

Over half are Tuya devices.


puddly:
### This has been moved to https://github.com/zigpy/zha/pull/565.

---

The last commit is due to a failure here:

```console
$ python -m tools.regenerate_diagnostics tests/data/devices/tze200-a7sghmms-ts0601.json
```

For some reason, the cluster handler-specific initialization code did not run. The cluster handler computed something that realistically should be handled by the entity so I ported the code over into there. This has the benefit of now allowing the summation formatting to change at runtime.

PR #565, opened 2025-11-03
Move `Metering` cluster formatting out of cluster handler and into sensor entity

puddly:
When working on https://github.com/zigpy/zha/pull/557 I ran into a bug: diagnostics could not be regenerated for [a specific device](https://github.com/zigpy/zha/pull/557#issuecomment-3475086107). For some reason, the cluster handler-specific initialization code did not run.

The cluster handler computed something that realistically should be handled by the entity so I ported the code over into there. This has the benefit of now allowing the summation formatting to change at runtime.

### DIFF
--- a/tests/data/devices/tze200-a7sghmms-ts0601.json
+++ b/tests/data/devices/tze200-a7sghmms-ts0601.json
@@ -1,0 +1,897 @@
+{
+  "version": 1,
+  "ieee": "ab:cd:ef:12:b5:0f:0b:2f",
+  "nwk": "0x5C8B",
+  "manufacturer": "_TZE200_a7sghmms",
+  "model": "TS0601",
+  "friendly_manufacturer": "_TZE200_a7sghmms",
+  "friendly_model": "TS0601",
+  "name": "_TZE200_a7sghmms TS0601",
+  "quirk_applied": true,
+  "quirk_class": "zigpy.quirks.v2.CustomDeviceV2",
+  "quirk_id": null,
+  "manufacturer_code": 4417,
+  "power_source": "Battery or Unknown",
+  "lqi": null,
+  "rssi": null,
+  "last_seen": "2025-07-05T08:46:21.109364+00:00",
+  "available": true,
+  "device_type": "EndDevice",
+  "active_coordinator": false,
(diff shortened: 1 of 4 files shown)


tests/data/devices/tze200-a7sghmms-ts0601.json
```diff
@@ -0,0 +1,897 @@
+{
+  "version": 1,
+  "ieee": "ab:cd:ef:12:b5:0f:0b:2f",
+  "nwk": "0x5C8B",
+  "manufacturer": "_TZE200_a7sghmms",
```
TheJulianJES, line 5: Hmm, interesting that it affected this Tuya valve with a v2 quirk. We have a local `Metering` cluster for that and it's one of the few times we actually do not use `kW(h)` as the unit for it, but instead water/liters:
https://github.com/zigpy/zha-device-handlers/blob/928268cca5d22a048e72d45ae1c24a1b6dd92a9c/zhaquirks/tuya/builder/__init__.py#L128-L148

TheJulianJES: Looked through the endpoint/ch initialization code and we skip important parts if "skip configuration" is enabled, both binding and attribute reporting in `ch.async_configure` but also the ch specific init code in [`ch.async_initialize`](https://github.com/zigpy/zha/blob/54975cb3387c0e3edf6beef3acab552ed4394d44/zha/zigbee/cluster_handlers/__init__.py#L462-L465), which later runs the ch specific init.
And sure enough, the quirk enables it: [zhaquirks/tuya/tuya_valve.py#L115](https://github.com/zigpy/zha-device-handlers/blob/928268cca5d22a048e72d45ae1c24a1b6dd92a9c/zhaquirks/tuya/tuya_valve.py#L115)

So, that at least solves the *why* behind the cluster handler specific initialization code not running.

Moving the `Metering` cluster formatting out of the cluster handler might actually break the entity displaying the delivered summation, now that the code is executed.
But agreed this is something we should do.

TheJulianJES (approved):
Should be good, though it might break that one Tuya device until the quirk is updated

PR #570, opened 2025-11-06
Don't use magic number for gateway shut down delay

emontnemery:
Don't use magic number for gateway shut down delay

Apart from a minor cleanup of the code, this also means we can patch the delay in tests, which will allow removing this workaround:
https://github.com/home-assistant/core/blob/6bed6d19fec829927267aea22f17b667eb4ccbd6/tests/components/zha/test_button.py#L48-L55

### DIFF
--- a/zha/application/gateway.py
+++ b/zha/application/gateway.py
@@ -58,7 +58,7 @@ from zha.zigbee.device import Device, DeviceInfo, DeviceStatus, ExtendedDeviceIn
 from zha.zigbee.group import Group, GroupInfo, GroupMemberReference
 
 BLOCK_LOG_TIMEOUT: Final[int] = 60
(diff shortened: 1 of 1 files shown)


MartinHjelmare (approved):

zha/application/gateway.py
```diff
@@ -58,6 +58,7 @@
 from zha.zigbee.group import Group, GroupInfo, GroupMemberReference
 
 BLOCK_LOG_TIMEOUT: Final[int] = 60
+SHUT_DOWN_DELAY_S: Final[int] = 0.1
```
MartinHjelmare:
```suggestion
SHUT_DOWN_DELAY_S: Final[float] = 0.1
```

PR #571, opened 2025-11-07
Refactor light entities to use a proper base class

puddly:
To bring `Light` in line with the other entities, I'm refactoring `BaseLight` to actually be a base entity type. The current base class was more of a base implementation and expected cluster handlers to exist, in addition to making other assumptions. I've split things off into:

- `BaseLight`: implements the external light entity API, nothing more.
- `BaseClusterHandlerLight`: implements the gargantuan ZCL-ish API to control lights, based on the presence of various cluster handlers and so on.

I took this opportunity to get rid of `kwargs.get(...` and have replaced this pattern with typed kwargs. This allowed the logic in `Light._assume_group_state` to safely be absorbed into `GroupLight`.

A few things have been renamed, to match the `BaseLight` API:
1. `_supported_color_modes` is now `_internal_supported_color_modes`
2. `_external_supported_color_modes` is now `_supported_color_modes`

This has revealed a problem: `state` exposes `_external_supported_color_modes`, not ` _supported_color_modes`. ZHA doesn't actually use this API (it directly accesses `entity.supported_color_modes`) but it's still a strange behavior. I think we should change this and regenerate diagnostics.

---

The purpose of this change is to allow light entities to be created for coordinators that have LEDs. These light entities are not going to be communicating with devices via Zigbee, using any sort of ZCL functionality, or even really attached to a device object in any way (like counter entities).


TheJulianJES (approved):
Changes look good

PR #572, opened 2025-11-08
Ignore `color_temperature` attribute when computing light capabilities

puddly:
See https://github.com/home-assistant/core/issues/156133.

The device switches modes. Color temperature is the only mode where we peek at attribute values to determine capabilities (all others just use the `color_capabilities` bitmap). The number of affected devices according to diagnostics is very small and all of the changes look reasonable to me.

### DIFF
--- a/tests/data/devices/ericsity-gl-c-009p.json
+++ b/tests/data/devices/ericsity-gl-c-009p.json
@@ -407,16 +407,15 @@
           "on": false,
           "brightness": 254,
           "xy_color": null,
-          "color_temp": 500,
+          "color_temp": null,
           "effect_list": [
             "off"
           ],
           "effect": "off",
           "supported_features": 40,
-          "color_mode": "color_temp",
+          "color_mode": "brightness",
           "supported_color_modes": [
             "brightness",
-            "color_temp",
             "onoff"
           ],
           "off_with_transition": false,
--- a/tests/data/devices/homr-hrmsn01.json
+++ b/tests/data/devices/homr-hrmsn01.json
@@ -519,7 +519,7 @@
             0.0,
             0.0
           ],
-          "color_temp": 0,
+          "color_temp": null,
           "effect_list": [
             "off"
           ],
@@ -528,7 +528,6 @@
           "color_mode": "xy",
           "supported_color_modes": [
             "brightness",
-            "color_temp",
             "onoff",
             "xy"
           ],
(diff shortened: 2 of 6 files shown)


TheJulianJES (commented):
For the original GLEDOPTO GL-C-009P issue, there's an [old quirk](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/gledopto/glc009p.py) that just removes the color cluster. If that were changed to a v2 quirk, it would also fix the issue.

But I agree this is the more proper fix. Going through the affected devices, the only devices we may wanna have a look at are:
- LEDVANCE RGB light, as it seems to support color temperature in Z2M:
  [zigbee-herdsman-converters/src/devices/sylvania.ts#L155-L159](https://github.com/Koenkk/zigbee-herdsman-converters/blob/3e441fd2fb76e235bca736d9eb3b948423ebf834/src/devices/sylvania.ts#L155-L159)
  (no idea why the range is "undefined", but this seems to apply to other "tunable white" devices as well: [zigbee-herdsman-converters/src/devices/ledvance.ts#L21-L25](https://github.com/Koenkk/zigbee-herdsman-converters/blob/3e441fd2fb76e235bca736d9eb3b948423ebf834/src/devices/ledvance.ts#L21-L25))
- the Tuya light `_TZ3210_bfwvfyx1`, but it apparently has lackluster support in ZHA, yet at least allows controlling the color temperature range [*somehow*](https://github.com/zigpy/zha-device-handlers/issues/1429#issuecomment-2507027418)

There may be other we are missing, but I think this change should be fine (for the next major release). Only "important issue" may be this (and other?) LEDVANCE lights..

GTerplan:
Is there any testing that needs to be done with this change?  If so I am willing to apply any file changes for the issue I reported until it's put into a future release.

puddly:
@GTerplan Give it a try, it should resolve your issue. We've identified that making this change will cause _other_ devices to break so we're working on fixing those first.

MattWestb:
light.ikea_of_sweden_tradfri_bulb_gu10_ws_345lm and 
light.dresden_elektronik_conbee_ii_no_name_group_0x549a is doing the same after adding the new IKEA Bilresa that is likely the reason the remote setting up the group and then sending X-Y commands like IKEA always have doing / using X-Y for color management.

GTerplan:
> @GTerplan Give it a try, it should resolve your issue. We've identified that making this change will cause _other_ devices to break so we're working on fixing those first.

@puddly Since I am using the HA OS VM I wasn't able to apply any of your changes, but I did create a zha_quirks folder within the config folder and had to modify the glc009p.py file based on my device signature.  I have included the updated file here since I am hoping that can just be updated in the main zha quirks folder during an update which should not break anything.
[glc009p.py](https://github.com/user-attachments/files/24539225/glc009p.py)

PR #575, opened 2025-11-13
Rename `quirk_id` to `exposes_features` + allow multiple features

TheJulianJES:
### Proposed change

This renames `quirk_id` to `exposes_features` internally and allows quirks to set multiple exposed features.

### Why

This will be useful for a number of issues, including Frient sensors that expose other IAS bits (alarm 1/2 + tamper + test bit), so we don't have to use custom quirks v2 entities for all devices using standard ZCL functionality. It's also useful to remove certain features we currently expose for all siren entities, like a lot of the siren specific parameters or the configuration entities.

In v1/v2 quirks, we can simply add a line that's something like `.exposes_feature(IAS_TAMPER_FEATURE)` or `.exposes_feature(SIREN_LIMITED_FEATURE)` and ZHA will add or change the behavior of some entities for those devices.

### Other notes

For now, I've left the attribute we check for the exposed features in v1 quirks called `quirk_id`. A previous iteration of this PR only renamed all internal references to `quirk_ids`, but I think we can just go ahead and rename it to `exposes_features` like done in this PR.
A small follow-up PR would add support from specifying exposed features in v2 quirks.

`quirk_id` can be a `set[str]`/`list[str]` or `str` as before, to make sure we don't break existing integrated v1 quirks or custom v1 quirks that may be used for some Tuya (plug) devices.

### Additional information

Addresses (see for more background info):
- https://github.com/zigpy/zha/issues/311

Other related issues:
- https://github.com/zigpy/backlog/issues/62 (with future PRs)
- https://github.com/zigpy/backlog/issues/61 (with future PRs)

For a diff without the regenerated diagnostics, see: https://github.com/zigpy/zha/compare/dev...TheJulianJES:zha:tjj/multiple_quirk_id_without_diagnostics

### DIFF
--- a/zha/application/platforms/select.py
+++ b/zha/application/platforms/select.py
@@ -283,10 +283,10 @@ class TuyaPowerOnState(types.enum8):
 
 
 @CONFIG_DIAGNOSTIC_MATCH(
-    cluster_handler_names=CLUSTER_HANDLER_ON_OFF, quirk_ids=TUYA_PLUG_ONOFF
+    cluster_handler_names=CLUSTER_HANDLER_ON_OFF, exposed_features=TUYA_PLUG_ONOFF
 )
 @CONFIG_DIAGNOSTIC_MATCH(
-    cluster_handler_names="tuya_manufacturer", quirk_ids=TUYA_PLUG_MANUFACTURER
+    cluster_handler_names="tuya_manufacturer", exposed_features=TUYA_PLUG_MANUFACTURER
 )
 class TuyaPowerOnStateSelectEntity(ZCLEnumSelectEntity):
     """Representation of a ZHA power on state select entity."""
@@ -306,7 +306,7 @@ class TuyaBacklightMode(types.enum8):
 
 
 @CONFIG_DIAGNOSTIC_MATCH(
-    cluster_handler_names=CLUSTER_HANDLER_ON_OFF, quirk_ids=TUYA_PLUG_ONOFF
+    cluster_handler_names=CLUSTER_HANDLER_ON_OFF, exposed_features=TUYA_PLUG_ONOFF
 )
 class TuyaBacklightModeSelectEntity(ZCLEnumSelectEntity):
     """Representation of a ZHA backlight mode select entity."""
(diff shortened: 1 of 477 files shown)


zha/application/platforms/select.py
```diff
@@ -283,10 +283,10 @@ class TuyaPowerOnState(types.enum8):
 
 
 @CONFIG_DIAGNOSTIC_MATCH(
-    cluster_handler_names=CLUSTER_HANDLER_ON_OFF, quirk_ids=TUYA_PLUG_ONOFF
+    cluster_handler_names=CLUSTER_HANDLER_ON_OFF, exposed_features=TUYA_PLUG_ONOFF
 )
```
TheJulianJES, line 287: In few places, I've explicitly used `exposed_features` internally, as we specify the device exposed features to match against. We don't expose any features ourselves here.

TheJulianJES:
I think we could get this in as well. It does what the PR title says, but it only processes the (v1) `quirk_id` in this PR, though it can now also be a `set`, `list`, or `str` like before.
This is needed to keep backwards compatibility for custom quirks (mostly Tuya plug quirks).

Another PR would add the exposed features from the v2 quirks API, something like https://github.com/TheJulianJES/zha/compare/tjj/multiple_quirk_id...TheJulianJES:zha:tjj/quirks_v2_exposed_features.

For a diff of this PR without the regenerated diagnostics (so GitHub dies less), see: https://github.com/zigpy/zha/compare/dev...TheJulianJES:zha:tjj/multiple_quirk_id_without_diagnostics

TheJulianJES:
This would require one small change in Core here: https://github.com/home-assistant/core/blob/9ebc6cbb23f470c5c371b1b5abb189513a9086af/homeassistant/components/zha/helpers.py#L344

You're good with this PR?

puddly (approved):
LGTM!

TheJulianJES:
> This would require one small change in Core here: https://github.com/home-assistant/core/blob/9ebc6cbb23f470c5c371b1b5abb189513a9086af/homeassistant/components/zha/helpers.py#L344

Core patch for this PR included here: https://github.com/home-assistant/core/compare/dev...TheJulianJES:tjj/zha-deps-bump-20251126

PR #586, opened 2025-11-26
Use quirks v2 exposed features

TheJulianJES:
This also adds the quirks v2 exposed features to the `device.exposed_features` set[str].

Related:
- https://github.com/zigpy/zha/pull/575
- https://github.com/zigpy/zigpy/pull/1711
- https://github.com/zigpy/zigpy/pull/1712

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -11,7 +11,12 @@ import pytest
 from zigpy.exceptions import ZigbeeException
 import zigpy.profiles.zha
 from zigpy.quirks.registry import DeviceRegistry
-from zigpy.quirks.v2 import DeviceAlertLevel, DeviceAlertMetadata, QuirkBuilder
+from zigpy.quirks.v2 import (
+    DeviceAlertLevel,
+    DeviceAlertMetadata,
+    ExposesFeatureMetadata,
+    QuirkBuilder,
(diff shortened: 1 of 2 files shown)


tests/test_device.py
```diff
@@ -1123,6 +1128,36 @@ async def test_quirks_v2_translation_placeholders(zha_gateway: Gateway) -> None:
...
+    zha_device = await join_zigpy_device(zha_gateway, zigpy_dev)
+
+    # can access the set of exposed features, similar to v1 quirks
+    assert zha_device.exposes_features == {"some_feature", "another_feature"}
+
+    # can access the quirk metadata for config features for now
+    assert zha_device.quirk_metadata.exposes_features == (
+        ExposesFeatureMetadata(feature="some_feature"),
+        ExposesFeatureMetadata(feature="another_feature", config={"option": True}),
+    )
```
TheJulianJES: We can access the config parameters using the `quirk_metadata` for now.

In the future, we can consider changing the `device.exposes_features` from a `set[str]` to something else, but as we don't need the config options at this very moment, it's something we can do later.
Not sure in how many situations this is an issue anyway.

puddly (approved):

PR #591, opened 2025-11-30
Avoid adding switch if not supported as attribute

elupus:
Some weird devices have OnOff clusters with only partial attributes. Tuya
remotes are one such device where both input and output endpoints have the OnOff
cluster added. Events go on outputs, manufacture attributes go on input.


puddly:
The `OnOff` attribute of the OnOff cluster is listed as mandatory in the ZCL spec. I'm thankful they at least explicitly marked it as unsupported 😄.

elupus:
Sometimes you get lucky :)

PR #594, opened 2025-12-03
Add events for LevelControl, Color and OnOff

elupus:
- Add LevelControl, On/Off, Color events 

Depends on https://github.com/home-assistant/core/pull/157853 for support in HA


elupus:
Just noticed that https://github.com/zigpy/zha/pull/419 exists too. I think this is simpler and stands on the main standardized clusters.

zha/application/platforms/event.py
```diff
@@ -0,0 +1,119 @@
...
+    def __init__(
+        self,
+        cluster_handlers: list[ClusterHandler],
+        endpoint: Endpoint,
+        device: Device,
+        **kwargs,
+    ) -> None:
+        """Initialize the fan."""
+        super().__init__(cluster_handlers, endpoint, device, **kwargs)
+        self._cluster_handler = self.cluster_handlers[self._cluster_handler_name]
```
puddly: For base platforms, I'm trying to move away from having any cluster handler code in the base entity type. This will allow for the public API to be specified more exactly and for entities to be created in the future that do not rely on cluster handlers. For example: https://github.com/puddly/zha/blob/26454b62db2a4fc8a302280a7326bbd1a8b261a6/zha/application/platforms/light/__init__.py#L106

elupus: Sounds reasonable. I've split the the base into a base and a cluster handler base

elupus:
Btw. I've explicitly avoided some fixed structure here. Tried to keep the original cluster commands as is. Currently HA has very few devices classes for events, so the structure of the event data there has not given correct structure.

Given the above situation, inventing some in-between structure that isn't 1 to 1 mapped to zigbee spec seem pointless and only adds complexity.

zha/application/discovery.py
```diff
@@ -478,19 +480,19 @@ def discover_by_cluster_id(self, endpoint: Endpoint) -> Iterator[PlatformEntity]
...
 
-        # until we can get rid of registries
-        yield from self.handle_on_off_output_cluster_exception(endpoint)
+        profile_id = endpoint.zigpy_endpoint.profile_id
+        device_type = endpoint.zigpy_endpoint.device_type
+        if device_type in REMOTE_DEVICE_TYPES.get(profile_id, []):
+            yield from self._handle_single_output_clusters(endpoint)
+        else:
+            # until we can get rid of registries
+            yield from self.handle_on_off_output_cluster_exception(endpoint)
```
elupus: I really don't understand this weird cluster exception here. It seems to add extra cluster handler at this stage.

puddly: It's from 2019. The more we delete from `discovery.py` the better 😄

elupus: Ok. Will look at it after goes in. I'd rather not make this larger than it is. The split of input/output handler is no longer required since the client cluster handlers have names and id's that separate them from server side. So this stuff can be simplified.

But I'd rather avoid making this larger than it already is, given how hard the test fixtures are to review.

puddly: Your PR motivated me to get rid of the old ZHA discovery, it's now gone, with https://github.com/zigpy/zha/pull/615 being merged yesterday. A few more PRs will get rid of cluster handlers entirely but those will come later.

The above PR aimed for bug-for-bug compatibility with the current discovery schema but I think it removes the need for the extra discovery code in this PR and we can make use of a `PlatformFeatureGroup` to match existing `OnOff` client clusters with specific device types.

elupus: Nice will have a look!

elupus:
So rebased on the new cluster rework. Must say it looks much cleaner and less complicated.

Hedda:
Any updates on this? PR for initial event support for zha in Home Assistant depends on it -> https://github.com/home-assistant/core/pull/157853

Also related to https://github.com/zigpy/zha/pull/419

It was btw mentioned in this other pull request that adding that event entity support pull request first "***is the way to go"***:

* https://github.com/zigpy/zha/pull/657

elupus:
I wonder the same

Hedda:
For reference, see https://github.com/zigpy/zha/pull/864 which implements the base platforms from https://github.com/zigpy/zha/pull/594 but is opt-in only for now to get ZGP devices working

elupus:
Will see if i get some time to rebase it.

Hedda:
Is this work to implement a "_Base IR platform_" also affect how this should be architected and implemented?

* https://github.com/zigpy/zha/pull/867

@elupus Should this pull request maybe also be rebased if and when https://github.com/zigpy/zha/pull/864 and https://github.com/zigpy/zha/pull/867 has been merged?

elupus:
Rebased with a lil-help from a friend. Much cleaner and simpler now. Nice work again on the removal of cluster handlers. I've not yet tested this thou, so must remain draft until tested, but the main ideas could be reviewed if wanted.

elupus:
Working well with the updated HA change. Seem to be an issue with hiding on/off cluster somewhere on my rotary dimmer (_TZ3000_ixla93vd). But i suspect that is a bug in the quirks. It has On/Off clusters both on server side and client side.

puddly:
Thanks for rebasing!

Right now, my main concern with generic discovery is that the emitted events are exposing low-level ZCL functionality rather than user-understandable button names: they all are cluster operations. Devices can emit `move_with_on_off` with different rates per button (and then a generic `stop`), or use `move_to_level`, or bare `on`, or `on` then `off`, and so on. These are all device-specific because no two devices expose events the same way...

This is an annoying situation because spec-complaint ZCL discovery and user-friendliness really don't intersect 😅. ZHA _already_ exposes this information via `zha_event` and (for quirked devices) automation triggers + blueprints, which is even less user-friendly. Automation triggers help because they map specific events to human-readable button names but there is no grouping like there is with event entities.

The only alternative is per-device quirks to map these to buttons, which is what we do now. I was thinking of maybe amending the quirks automation triggers method to _also_ create event entities, since those are "vetted" and already do button mapping?

@TheJulianJES what are your thoughts?

elupus:
Just small comment.
They are not really device specific. They are standardised, sure some (many maybe) bad devices exists, but there is a spec on behaviour. Which is more than there is for standard events in HA.

My though is that, we should map to "standardized" HA events where they exists. But in absence, we should show the original events that it is possible to read a spec on how they are expected to work.

These are proper events with proper translations which can be automated on, in comparison to the global device event emitter that is hard to match and see.

puddly:
I don't think it's a matter of non-compliance in this case. Server clusters have attributes that indicate support for commands and such. But a client cluster is entirely open ended: the device is free to send _any_ command with any command parameters (i.e. different move rates for different directions or even buttons).

PR #599, opened 2025-12-11
Import new device diagnostics

puddly:
I've also modified the legacy diagnostics importer to require every failure to be handled on a case-by-case basis: many old devices (mostly Tuya) failed to import and were silently ignored.


tools/import_diagnostics.py
```diff
@@ -110,17 +198,33 @@ def zigpy_device_from_legacy_diagnostics(
...
+                    cluster,
+                )
+                continue
+
             if patch_cluster:
                 patch_cluster_for_testing(real_cluster)
             for attr_id, attr in cluster["attributes"].items():
                 if (
-                    attr["value"] is None
+                    attr.get("value", None) is None
```
TheJulianJES: nit: `None` should already be the default:
```suggestion
                    attr.get("value") is None
```

tools/import_diagnostics.py
```diff
@@ -135,17 +239,33 @@ def zigpy_device_from_legacy_diagnostics(
...
```
TheJulianJES: nit: `None` should already be the default:
```suggestion
                    attr.get("value") is None
```

TheJulianJES:
Oh wow, the GitHub diff viewer is actually useful for this now:
image

TheJulianJES (approved):
Nice!

PR #605, opened 2025-12-23
Zigpy 0.91.0: attribute events, manufacturer codes, and signature fixes

puddly:
ZHA fixes for https://github.com/zigpy/zigpy/pull/1719.

A few tests needed tweaking because of attribute "collisions" but otherwise nothing is breaking as far as I can tell and a few new entities are showing up for existing test devices.

The most visible change is switching from `attribute_updated` to explicit events for every possible attribute change type.


zha/zigbee/cluster_handlers/general.py
```diff
@@ -466,13 +471,22 @@ def cluster_command(self, tsn, command_id, args):
...
-        else:
-            super().attribute_updated(attrid, value, timestamp)
+        self.debug(
+            "received attribute: %s update with value: %s",
+            event.attribute_id,
+            event.value,
+        )
+        if event.attribute_id == self.CURRENT_LEVEL:
+            self.dispatch_level_change(SIGNAL_SET_LEVEL, event.value)
+        super()._handle_attribute_updated_event(event)
```
TheJulianJES: This level dispatching should likely be cleaned up to use `ClusterAttributeUpdatedEvent` directly, but we specifically avoided emitting a ZHA event in HA for these before. Now, I think we do.

This would be a lot of logbook entries when turning on a light on/off (even with a default 1s transition).

For now, I think we can just add that `else` back in before calling `super()._handle_attribute_updated_event(event)` (if it's not the current level attribute that was updated), so:

```suggestion
        if event.attribute_id == self.CURRENT_LEVEL:
            self.dispatch_level_change(SIGNAL_SET_LEVEL, event.value)
        else:
            super()._handle_attribute_updated_event(event)
```

zha/zigbee/cluster_handlers/general.py
```diff
@@ -466,13 +471,22 @@ def cluster_command(self, tsn, command_id, args):
...
+        | AttributeReportedEvent
+        | AttributeUpdatedEvent
+        | AttributeWrittenEvent,
+    ) -> None:
         """Handle attribute updates on this cluster."""
-        self.debug("received attribute: %s update with value: %s", attrid, value)
-        if attrid == self.CURRENT_LEVEL:
-            self.dispatch_level_change(SIGNAL_SET_LEVEL, value)
-        else:
-            super().attribute_updated(attrid, value, timestamp)
```
TheJulianJES, line 475: For reference, state before.

zha/zigbee/cluster_handlers/manufacturerspecific.py
```diff
@@ -256,7 +262,13 @@ def attribute_updated(self, attrid: int, value: Any, timestamp: datetime) -> Non
...
 
-    def attribute_updated(self, attrid: int, value: Any, _: Any) -> None:
+    def _handle_attribute_updated_event(
+        self,
+        event: AttributeReadEvent
+        | AttributeReportedEvent
+        | AttributeUpdatedEvent
+        | AttributeWrittenEvent,
+    ) -> None:
         """Handle an attribute updated on this cluster."""
```
TheJulianJES, line 272: Side note: Could be nice to add some sort of `EMIT_EVENTS` attribute for `ClusterHandler`s in general, so we don't need this empty override, just to block emitting ZHA events.
Could just set `EMIT_EVENTS = False` here.

tests/test_sensor.py
TheJulianJES, line 1: This was a nice lesson on how to spell "instantaneous_demand" 😄

TheJulianJES (commented):
This looks good. Only the comment about emitting ZHA events for `LevelControl` changes should be addressed IMO, as it'll get spammy otherwise. That part needs to be reworked a bit in the future anyway.

Otherwise, why do the diagnostics change?

zha/zigbee/cluster_handlers/__init__.py
TheJulianJES, line 643: Side note: We should also update the batched `get_attributes` when initializing a cluster handler to properly handle manufacturer codes based on attribute definitions. The code is really flawed right now.

`configure_reporting` should likely be adjusted as well (and likely everything where we do the `self.cluster.cluster_id >= 0xFC00 and manufacturer_code` check).

We can also do that in a separate PR.

zha/zigbee/cluster_handlers/general.py
puddly: Thanks! I've added the `else` back in.

puddly:
I tracked down why the new entities were being created for some of these Tuya devices and it's a diagnostics loading bug triggered by zigpy changing `set_attribute` semantics: if an attribute is marked unsupported but then receives a value, the "unsupported" state is cleared. Some quirks mark the attributes as unsupported within the cluster's `__init__`.

With that, I think there's no unexpected diagnostics JSON changes with this PR.

TheJulianJES (approved):

PR #612, opened 2026-01-03
Fix test compatibility with looptime 0.7

TheJulianJES:
## Proposed change

`TestGateway` calls are now wrapped with `looptime.enabled()` to temporarily enable looptime time compaction during fixture setup and teardown.

## Additional information

- [looptime 0.3](https://github.com/nolar/looptime/releases/tag/0.3) (https://github.com/nolar/looptime/pull/9) intended to run fixtures in real wall-clock time while only test functions use mocked time. However, a bug caused fixtures to use a fake zero-based clock instead.
- [looptime 0.6](https://github.com/nolar/looptime/releases/tag/0.6) (https://github.com/nolar/looptime/pull/20) added `looptime.enabled()` as an explicit workaround for fixtures needing mocked time.
- [looptime 0.7](https://github.com/nolar/looptime/releases/tag/0.7) (https://github.com/nolar/looptime/pull/23) fixed the bug so fixtures now correctly use real wall-clock time, breaking our tests as we apparently relied on the buggy behavior somewhat.

It also seems like tests are running a bit quicker now.

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -12,6 +12,6 @@ pytest
 python-slugify
 zigpy>=0.65.2
 ruff
-looptime
+looptime>=0.7
 freezegun
 coloredlogs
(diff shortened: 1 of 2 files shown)


puddly (approved):
Thanks!

PR #615, opened 2026-01-06
Move from centralized discovery to entity-driven discovery

puddly:
As part of a medium-term goal of getting rid of cluster handlers and eventually rewriting entity classes themselves to request attribute reporting config/binding/attribute values, I think a first stepping stone would be to move away from implicit entity registration decorators and have entity objects themselves decide what cluster handlers they want (or if they are not applicable).

Instead of this:

```python
@STRICT_MATCH(
    cluster_handler_names=CLUSTER_HANDLER_ON_OFF,
    aux_cluster_handlers={CLUSTER_HANDLER_COLOR, CLUSTER_HANDLER_LEVEL},
)
class Light(BaseClusterHandlerLight, PlatformEntity):
    ...
```

We do this:

```python
@register_entity
class Light(BaseClusterHandlerLight, PlatformEntity):
    ...
    _cluster_handler_match = ClusterHandlerMatch(
        cluster_handlers=frozenset({CLUSTER_HANDLER_ON_OFF}),
        optional_cluster_handlers=frozenset(
            {CLUSTER_HANDLER_COLOR, CLUSTER_HANDLER_LEVEL}
        ),
        profile_device_types=LIGHT_PROFILE_DEVICE_TYPES,
        feature_priority=(PlatformFeatureGroup.LIGHT_OR_SWITCH_OR_SHADE, 0),
    )
```

There is no more complexity in `ClusterHandlerMatch`. It lists what cluster handlers are required, what cluster handlers are optional, a bit of model/manufacturer filtering, and `feature_priority` for exclusivity within a group of overlapping clusters. That's it.

This has a few benefits:
1. It moves unique ID calculation logic into the platform base entity object and makes it 100% explicit, allowing us to more easily migrate.
2. Allows us to fine-tune cluster handler matching per-entity without adding more matching rules.
3. Will allow us to move cluster handler ZCL attribute reporting and binding config into the ZCL entity objects themselves, allowing ZHA to granularly merge binding/reporting config when setting up a device.


tests/test_button.py
```diff
@@ -412,11 +412,7 @@ async def test_cluster_handler_quirks_unnecessary_claiming(
     )
     zigpy_device = registry.get_device(zigpy_device)
 
-    # Suppress normal endpoint probing, as this will claim the Opple cluster handler
-    # already due to it being in the "CLUSTER_HANDLER_ONLY_CLUSTERS" registry.
-    # We want to test the handler also gets claimed via quirks v2 attributes init.
-    with patch("zha.application.discovery.EndpointProbe.discover_entities"):
-        zha_device = await join_zigpy_device(zha_gateway, zigpy_device)
```
TheJulianJES, line 419: The test would already always pass if this patch is removed before. It no longer tests that the quirks v2 entities claim the `OppleRemoteClusterHandler` then. The default ZHA entities will always claim that, hence discovery for those was patched out.

I'll need to have a proper look at this later, but I think these tests essentially test nothing (correctly) now.

zha/application/platforms/light/__init__.py
```diff
@@ -782,6 +779,64 @@ def __init__(
...
+            (zha.PROFILE_ID, zha.DeviceType.ON_OFF_LIGHT),
+            # ZLL
+            (zll.PROFILE_ID, zll.DeviceType.COLOR_LIGHT),
+            (zll.PROFILE_ID, zll.DeviceType.COLOR_TEMPERATURE_LIGHT),
+            (zll.PROFILE_ID, zll.DeviceType.DIMMABLE_LIGHT),
+            (zll.PROFILE_ID, zll.DeviceType.DIMMABLE_PLUGIN_UNIT),
+            (zll.PROFILE_ID, zll.DeviceType.EXTENDED_COLOR_LIGHT),
+            (zll.PROFILE_ID, zll.DeviceType.ON_OFF_LIGHT),
+        }:
+            return None
```
TheJulianJES: We should put these into a constant [somewhere](https://github.com/zigpy/zha/blob/dev/zha/application/platforms/light/const.py) IMO.

zha/application/platforms/light/__init__.py
```diff
@@ -782,6 +779,64 @@ def __init__(
...
+
+        # Or with `ForceOnLight`
+        if endpoint.device.manufacturer in {
+            "Jasco",
+            "Jasco Products",
+            "Quotra-Vision",
+            "eWeLight",
+            "eWeLink",
+        }:
+            return None
```
TheJulianJES: This doesn't scale really well, as we have to keep repeating ourselves if we add another entity that's more specific compared to the base one, with the intention of replacing the base one.

The discovery system we had before with the weighting and `stop_on_match_group` was a bit awkward, but I think it's closer to the new discovery schemas used in Matter (and slowly in Z-Wave).

For reference, see [Matter's discovery here](https://github.com/home-assistant/core/blob/4fb89e68a790cac50665c1d8f0a371a1af4478f0/homeassistant/components/matter/discovery.py#L60-L225) and an example of a specific [binary sensor replacement](https://github.com/home-assistant/core/blob/4fb89e68a790cac50665c1d8f0a371a1af4478f0/homeassistant/components/matter/binary_sensor.py#L66-L90) here. (The Matter `key` is used as a unique ID suffix.)
By default, a single Matter entity claims an attribute, adding that to `discovered_attributes`.
Unless `allow_multi=True` is set, a second entity will not be discovered for that attribute.

As far as I can see, the Hue sensor replacement only works, as it's first in the list (before the generic/fallback one) (and the default is `allow_multi=False`). I do think that's not a perfect solution and I'd have switched the default the other way around, as that can be easier seen in tests, but our weighting system sorting the discovered entities based on specificity was actually a nice solution for that IMO.

---

Maybe we should talk about this a bit. What do you think of introducing something similar to `allow_multi` but in the opposite direction, e.g. `exclusive_claim`? This would only be set on the more specific entities (e.g. vendor-specific Hue sensor replacing generic one for the Matter example, or, in our case, `ForceOnLight` replacing the generic one, for example).

If we need to match in a much more specific way, we can simply do it manually, like done here, but in most cases, we should be able to set `exclusive_claim=True`, only for the more specific entities.
The more specific entities would automatically get higher priority compared to ones that do not have `exclusive_claim` set, so these entities would be discovered first. For the `exclusive_claim=True` entities, the primary attribute (or a key, same for both entities) would be added to a list/set of "claimed attributes/keys".
Later, all normal entities with `exclusive_claim=False` check if their attribute/key is in in the claimed attributes/keys. If not, they're created normally. But if it is, they would skip creation.

This would be a much more simple system than our previous weighting system, and we should be able to keep all benefits you mentioned in the PR description, at least from a quick look, but also simplify and clean this up a lot IMO.

puddly: I completely agree! The PR in its current state is still very much a WIP (I haven't actually run it) and the entity-based discovery needs to be completely rewritten; I was curious to see how far I can get with making all of the matching logic painfully verbose and entirely isolated to individual entities.

In this branch, 99% of all entities are simply created based on the presence of a cluster with no further filtering. The 1% of exceptions are now very explicit:
- The `OnOff` server cluster can be turned into a `Light` or a `Switch` based on the endpoint device type.
- The `OnOff` server cluster can _also_ be turned into a `Shade` if `Level` and `Shade` clusters also exist and the device type is correct.
- The `Fan` server cluster is not created if a `Thermostat` entity is currently claiming it, to prevent creating a duplicate entity for fan functionality that's currently part of the complex thermostat entity.

I think that's actually _all_ of them. If we migrate all of the device-specific quirks out of the ZHA codebase and into quirks, the ZCL-only functionality becomes super straightforward. Something like:

```python
from zha.platforms.climate import BaseThermostat

class MyWeirdThermostat(BaseThermostat):
    def __init__(self, device) -> None:
        self._device = device

    async def async_turn_on(self) -> None:
        await self._device.send_packet(...)

(
    QuirkBuilder()
    .applies_to("manuf", "model")
    # Prevent ZCL entity auto-discovery
    .prevent_default_entity_creation(endpoint_id=1)
    # And substitute our own
    .exposes_entity(MyWeirdThermostat)
    .add_to_registry()
)
```

> What do you think of introducing something similar to `allow_multi` but in the opposite direction, e.g. `exclusive_claim`?

Having an entity _prevent_ future entities from doing something is an API that I'd like to avoid. Once you move the device-specific "quirks" out of core ZHA, we may not even need claiming/exclusivity as part of the entity matching API. We could eventually use static discovery schemas:

```python
class EntityDiscovery:
    # Mandatory clusters
    server_clusters: frozenset[type[Cluster]]
    client_clusters: frozenset[type[Cluster]]

    # Missing clusters
    not_server_clusters: frozenset[type[Cluster]]
    not_client_clusters: frozenset[type[Cluster]]

    # A set of (profile, device_type)s
    device_types: frozenset[tuple[int, int]]

    # Some way to specify attributes to peek at??

    # Optional clusters will be pulled from the endpoint during initialization
```

puddly: I've added a feature group weighting system [nearly identical to the above](https://github.com/zigpy/zha/pull/615/changes#diff-e6aafd60b21a74e66657008f459f013056b7995e95af6fd134ebe54397307b66R1043-R1056). It slightly increases discovery complexity but removes the need for all filtering logic in `match_cluster_handlers()`.

At this point, all but a few `def match_cluster_handlers` functions are entirely static. The only dynamic logic is to determine the `unique_id` format based on the endpoint profile and device type, which I think we can get rid of.

tests/test_button.py
puddly: The entire unit test could actually be removed, I think. This PR removes the concept of claiming cluster handlers: we claim them to tell ZHA to "use" them, nothing more.

TheJulianJES: Ok, I'll still need to take a look at this. Previously, claiming them determined setting up attribute reporting and if they're bound to the coordinator or not.

puddly:
I'm successfully running this branch locally, as a final test. No entity issues to report.

tests/data/devices/centralite-3405-l.json
TheJulianJES, line 273: Uh, why did we even have two alarm control panel entities for this keypad before? 😅

TheJulianJES: Ah, that was the issue I also ran into with the Frient keypad: client and server cluster handlers, though there was only one entity with two cluster handlers causing inconsistent state.

Are we removing the right one now?

puddly: This device has both a client and a server `IasAce` cluster on the same endpoint. Previous discovery code didn't really distinguish the two, now we explicitly only create one when it's a client cluster.

tests/test_sensor.py
```diff
@@ -1689,12 +1689,7 @@ async def test_state_class(
 
 async def test_cluster_handler_quirks_attribute_reporting(zha_gateway: Gateway) -> None:
     """Test quirks sensor setting up ZCL_INIT_ATTRS and REPORT_CONFIG correctly."""
-
-    # Suppress normal endpoint probing, as this will claim the Opple cluster handler
-    # already due to it being in the "CLUSTER_HANDLER_ONLY_CLUSTERS" registry.
-    # We want to test the handler also gets claimed via quirks v2 reporting config.
-    with patch("zha.application.discovery.EndpointProbe.discover_entities"):
-        zha_device, cluster = await zigpy_device_aqara_sensor_v2_mock(zha_gateway)
```
TheJulianJES, line 1697: Can you also re-add the `with patch("zha.application.discovery.discover_entities_for_endpoint"):` patch here and below? It's used in three separate tests..

tests/test_sensor.py
```diff
@@ -1773,11 +1768,7 @@ async def test_cluster_handler_quirks_attribute_reading(zha_gateway: Gateway) ->
     )
     zigpy_device = registry.get_device(zigpy_device)
 
-    # Suppress normal endpoint probing, as this will claim the Opple cluster handler
-    # already due to it being in the "CLUSTER_HANDLER_ONLY_CLUSTERS" registry.
-    # We want to test the handler also gets claimed via quirks v2 attributes init.
-    with patch("zha.application.discovery.EndpointProbe.discover_entities"):
-        zha_device = await join_zigpy_device(zha_gateway, zigpy_device)
```
TheJulianJES, line 1780: And here:

tests/test_sensor.py
puddly: Oops, I forgot we had a few more. Thanks!

tests/data/devices/centralite-3405-l.json
puddly: I believe so, I can double check with a real device a bit later.

zha/application/discovery.py
```diff
@@ -136,512 +129,412 @@ def inner(*args: P.args, **kwargs: P.kwargs) -> Iterator[T]:
...
+        # if the cluster handler is unclaimed, claim it and set BIND accordingly,
+        # so ZHA configures the cluster handler: reporting + reads attributes
+        if attribute_initialization_found or reporting_found:
+            endpoint.claim_cluster_handlers([cluster_handler])
+            # BIND is True by default, so only set to False if no reporting found.
+            # We can safely do this, since quirks v2 entities are initialized last,
+            # so if the cluster handler wasn't claimed by endpoint probing so far,
+            # only v2 entities need it.
+            if not reporting_found:
+                cluster_handler.BIND = False
```
TheJulianJES, line 389: Hmm, so previously, we only executed this code for unclaimed cluster handlers (as still in the comment). 
I think the main reason for doing so was to only run the `cluster_handler.BIND = False` if ZHA doesn't already use it.
(Running `endpoint.claim_cluster_handlers([cluster_handler])` multiple times should be fine though?)

Now, if ZHA already claimed a cluster handler (for one of its entities) to set up both binding and attribute reporting, and we have a v2 quirks entity that's using an attribute from that cluster (for initialization, but not for attribute reporting), we'll set `BIND = False` here, but the ZHA cluster handler likely wants that cluster to be bound. So, we cause some issues.

---

This all needs to be redone anyway when init + report config is moving to entities (and I guess something like `request_bind` to only bind a cluster if at least one entity is discovered and requires it to be bound?). 

[`BIND = True`](https://github.com/zigpy/zha/blob/3c342c46818e9b71fbc7ff7807d165e81557eb72/zha/zigbee/cluster_handlers/__init__.py#L193) in the base `ClusterHandler`. It's only selectively disabled in some other cluster handlers (so those are never bound). But for most cluster handlers, the cluster would only be bound (at least before this PR) if it's a claimed cluster handler, e.g. used/claimed by at least one entity.

So, with the code above, we checked if it was an unclaimed cluster handler, so we knew that no ZHA entity relied on it at all (/for binding), letting us safely set `BIND = False` then, as claiming that cluster handler above would have bound the cluster handler, otherwise (even though we just want to claim it here for `ZCL_INIT_ATTRS`). But if no reporting is configured by v2 quirks, we can safely disable binding completely for that cluster handler.

With the changes from this PR, we will disable binding for all cluster handlers that would otherwise be bound by ZHA if there's just one quirks v2 entity on that same cluster (with it not having attribute reporting configured, but possibly still with the base cluster handler doing so).
So, I think there's a bug here? Can we just keep the unclaimed cluster handler check here, for now, just so we know that no ZHA entity relies on that cluster handler (being bound)?

.. Does my explanation make sense? 😅

puddly: Ah. I just changed this a few hours ago. I'll revert the change. We apparently don't have test coverage for this specific scenario 😅

TheJulianJES: Oops, you're right. Should have been done in https://github.com/zigpy/zha/pull/548  😅 
I can look into adding a test for that tomorrow-ish, but hopefully we'll *all make this better* soon, anyway.

TheJulianJES (approved):
I've not spotted any (obvious) issues either, so here we go 😄

zha/application/platforms/update.py
```diff
@@ -324,11 +328,14 @@ def _get_cluster_version(self) -> str | None:
...
 
-@CONFIG_DIAGNOSTIC_MATCH(cluster_handler_names=CLUSTER_HANDLER_OTA_SERVER)
+@register_entity(Ota.cluster_id)
 class FirmwareUpdateServerEntity(BaseFirmwareUpdateEntity):
     """Representation of a ZHA firmware update entity."""
 
     _unique_id_suffix = "firmware_update"
+    _cluster_handler_match = ClusterHandlerMatch(
+        cluster_handlers=frozenset({CLUSTER_HANDLER_OTA})
+    )
```
elupus, line 338: This looks strange @puddly. This entity used to match to the server side cluster handler. Now it is identical to the one above. So should match to same.

puddly: There are `cluster_handlers` and `client_cluster_handlers`. They're registered by name so the names will be the same for both cluster handlers but they now match correctly.

elupus: Oh. I missed that detail. Then it make sense. Looking at rebasing my event series.

PR #621, opened 2026-01-12
Remove legacy hardcoded attributes for lumi.airrtc.agl001

### LINKED ISSUE #620: Remove legacy hardcoded attributes for lumi.airrtc.agl001 (Aqara E1 TRV)
## Summary

The `manufacturerspecific.py` file contains hardcoded `ZCL_INIT_ATTRS` for `lumi.airrtc.agl001` (Aqara E1 TRV) that creates duplicate entities when used alongside the quirk in zha-device-handlers.

## Location

`zha/zigbee/cluster_handlers/manufacturerspecific.py` lines ~128-141:

```python
elif self.cluster.endpoint.model == "lumi.airrtc.agl001":
    self.ZCL_INIT_ATTRS = {
        "system_mode": True,
        "preset": True,
        "window_detection": True,
        "valve_detection": True,
        "valve_alarm": True,
        "child_lock": True,
        "away_preset_temperature": True,
        "window_open": True,
        "calibrated": True,
        "schedule": True,
        "sensor": True,
    }
```

## Problem

Users see duplicate entities:
- Legacy entities created by this hardcoded config (German translations from HA core)
- Quirk-defined entities from zha-device-handlers (English names from quirk)

## Proposed Solution

Remove this `elif` block once the improved quirk is merged in zigpy/zha-device-handlers#4604.

## Related

- zigpy/zha-device-handlers#4604 - Improved Aqara E1 TRV quirk with proper entity definitions

cl445:
## Summary

Removes legacy hardcoded entity definitions for the Aqara E1 TRV (`lumi.airrtc.agl001`).

The device now has a proper quirk in zha-device-handlers that defines all necessary entities using the quirks v2 API (see zigpy/zha-device-handlers#4604).

## Problem

These legacy definitions create duplicate entities alongside the quirk-defined ones:
- Legacy entities have translated names (e.g., "Kindersicherung", "Fenstererkennung")
- Quirk entities have English names (e.g., "Child lock", "Window detection")

This confuses users who see both sets of entities for the same functionality.

## Changes

**Removed from `manufacturerspecific.py`:**
- `ZCL_INIT_ATTRS` block for `lumi.airrtc.agl001` in `OppleRemoteClusterHandler`

**Removed from platform files:**

| File | Removed Classes |
|------|-----------------|
| `binary_sensor/__init__.py` | `AqaraThermostatWindowOpen`, `AqaraThermostatValveAlarm`, `AqaraThermostatCalibrated`, `AqaraThermostatExternalSensor` |
| `switch.py` | `AqaraThermostatWindowDetection`, `AqaraThermostatValveDetection`, `AqaraThermostatChildLock` |
| `select.py` | `AqaraThermostatPresetMode` (enum), `AqaraThermostatPreset` |
| `number/__init__.py` | `AqaraThermostatAwayTemp` |

**Total: 128 lines removed**

## Test Plan

- [x] Verified quirk provides all entities (zigpy/zha-device-handlers#4604)
- [ ] Test with real device after both PRs merged

## Related

- Closes #620
- Companion PR: zigpy/zha-device-handlers#4604 (improved Aqara E1 TRV quirk)

Thanks to @TheJulianJES for explaining that `ZCL_INIT_ATTRS` alone doesn't create entities - the actual entity classes in platform files needed to be removed as well!

### DIFF
--- a/zha/application/platforms/binary_sensor/__init__.py
+++ b/zha/application/platforms/binary_sensor/__init__.py
@@ -331,49 +331,6 @@ class XiaomiPlugConsumerConnected(BinarySensor):
     _attr_translation_key: str = "consumer_connected"
 
 
-@MULTI_MATCH(cluster_handler_names="opple_cluster", models={"lumi.airrtc.agl001"})
-class AqaraThermostatWindowOpen(BinarySensor):
-    """ZHA Aqara thermostat window open binary sensor."""
-
-    _attribute_name = "window_open"
-    _unique_id_suffix = "window_open"
-    _attr_device_class: BinarySensorDeviceClass = BinarySensorDeviceClass.WINDOW
-
-
-@MULTI_MATCH(cluster_handler_names="opple_cluster", models={"lumi.airrtc.agl001"})
-class AqaraThermostatValveAlarm(BinarySensor):
-    """ZHA Aqara thermostat valve alarm binary sensor."""
-
-    _attribute_name = "valve_alarm"
-    _unique_id_suffix = "valve_alarm"
-    _attr_device_class: BinarySensorDeviceClass = BinarySensorDeviceClass.PROBLEM
-    _attr_translation_key: str = "valve_alarm"
-
-
-@CONFIG_DIAGNOSTIC_MATCH(
-    cluster_handler_names="opple_cluster", models={"lumi.airrtc.agl001"}
-)
-class AqaraThermostatCalibrated(BinarySensor):
-    """ZHA Aqara thermostat calibrated binary sensor."""
-
-    _attribute_name = "calibrated"
-    _unique_id_suffix = "calibrated"
-    _attr_entity_category: EntityCategory = EntityCategory.DIAGNOSTIC
-    _attr_translation_key: str = "calibrated"
-
-
-@CONFIG_DIAGNOSTIC_MATCH(
-    cluster_handler_names="opple_cluster", models={"lumi.airrtc.agl001"}
-)
-class AqaraThermostatExternalSensor(BinarySensor):
-    """ZHA Aqara thermostat external sensor binary sensor."""
-
-    _attribute_name = "sensor"
-    _unique_id_suffix = "sensor"
-    _attr_entity_category: EntityCategory = EntityCategory.DIAGNOSTIC
-    _attr_translation_key: str = "external_sensor"
-
-
 @MULTI_MATCH(cluster_handler_names="opple_cluster", models={"lumi.sensor_smoke.acn03"})
 class AqaraLinkageAlarmState(BinarySensor):
     """ZHA Aqara linkage alarm state binary sensor."""
--- a/zha/application/platforms/number/__init__.py
+++ b/zha/application/platforms/number/__init__.py
@@ -758,24 +758,6 @@ class AqaraPetFeederPortionWeight(NumberConfigurationEntity):
 
 
 @CONFIG_DIAGNOSTIC_MATCH(
-    cluster_handler_names="opple_cluster", models={"lumi.airrtc.agl001"}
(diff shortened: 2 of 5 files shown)


TheJulianJES (changes requested):
This only initializes the attributes during pairing/reconfiguration for ZHA entities. (This is automatically done by entities defined with v2 quirks.)

The ZHA entities themselves are defined in the relevant ZHA platforms. Search the repo for the model string [(here)](https://github.com/search?q=repo%3Azigpy%2Fzha%20lumi.airrtc.agl001&type=code). They need to be removed from there.

Also, we should set the `unique_id_suffix` for the v2 quirk defined entities accordingly to match what the ZHA entities auto-generated. It should be something like this: `1-64704-child_lock` (endpoint id + cluster id + ZHA unique id suffix).
Otherwise, users would see new entities and we'd break old ones. We can't do that.

cl445:
Thanks for the detailed explanation @TheJulianJES! You're absolutely right - I misunderstood where the actual entity definitions come from.

I've now added `unique_id_suffix` to all entity definitions in the v2 quirk (zigpy/zha-device-handlers#4604) to match the legacy entity IDs:

- Switches: `child_lock`, `window_detection`, `valve_detection`, `schedule`
- Binary sensors: `valve_alarm`, `window_open`, `calibrated`
- Enums: `preset`, `sensor`
- Number: `away_preset_temperature`

This should ensure that the quirk entities "take over" the existing legacy entity IDs, preventing duplicates and preserving user automations/dashboards.

Given this approach, is this PR (#621) still needed? Or would the `unique_id_suffix` solution in the quirk be sufficient to handle the migration?

TheJulianJES:
The ZHA entities can be removed if they're all properly replaced by the quirks entities. Refer to my message above on what else needs to be done in this PR:
> The ZHA entities themselves are defined in the relevant ZHA platforms. Search the repo for the model string [here](https://github.com/search?q=repo%3Azigpy%2Fzha%20lumi.airrtc.agl001&type=code). They need to be removed from there.

cl445:
I've expanded this PR to also remove the legacy entity class definitions from the platform files:

**Removed:**
- `binary_sensor/__init__.py`: 4 classes (WindowOpen, ValveAlarm, Calibrated, ExternalSensor)
- `switch.py`: 3 classes (WindowDetection, ValveDetection, ChildLock)
- `select.py`: 1 enum + 1 class (PresetMode, Preset)
- `number/__init__.py`: 1 class (AwayTemp)

**Total: 128 lines removed across 5 files**

I tested locally and confirmed that these legacy entity classes were indeed creating duplicates:
- Legacy: `switch.tvr_badezimmer_oben_kindersicherung` (German translation)
- Quirk: `switch.lumi_lumi_airrtc_agl001_child_lock` (English)

The `unique_id_suffix` approach I tried earlier didn't work because the unique_id base format is completely different between legacy (device-name based) and quirk (model-name based) entities.

This PR should now properly eliminate all duplicates when paired with the quirk PR (zigpy/zha-device-handlers#4604).

TheJulianJES:
> The `unique_id_suffix` approach I tried earlier didn't work because the unique_id base format is completely different between legacy (device-name based) and quirk (model-name based) entities.

From my previous message:

> Also, we should set the `unique_id_suffix` for the v2 quirk defined entities accordingly to match what the ZHA entities auto-generated. It should be something like this: `1-64704-child_lock` (endpoint id + cluster id + ZHA unique id suffix).
Otherwise, users would see new entities and we'd break old ones. We can't do that.

You need to add the endpoint id and cluster id in v2 quirks UID, only then add the ZHA unique ID suffix. The example I mentioned should work for the child lock entity. The rest are very similar. `1-64704-` should work fine as a prefix.

PR #623, opened 2026-01-14
Update Aqara E1 thermostat diagnostics `lumi.airrtc.agl001`

TheJulianJES:
The diagnostics that were previously added must have been from some PR that causes them to not match existing the quirk, so update with actual diagnostics.

Also, the tool to add them doesn't override them when re-adding, instead failing with unhelpful errors 😄

### DIFF
--- a/tests/data/devices/lumi-lumi-airrtc-agl001.json
+++ b/tests/data/devices/lumi-lumi-airrtc-agl001.json
@@ -12,9 +12,9 @@
   "exposes_features": [],
   "manufacturer_code": 4447,
   "power_source": "Battery or Unknown",
-  "lqi": 132,
-  "rssi": -67,
-  "last_seen": "2026-01-12T21:06:46.129234+00:00",
+  "lqi": 140,
+  "rssi": -65,
+  "last_seen": "2026-01-14T04:37:29.927773+00:00",
   "available": true,
   "device_type": "EndDevice",
   "active_coordinator": false,
@@ -143,7 +143,7 @@
               "id": "0x0000",
               "name": "local_temperature",
               "zcl_type": "int16",
-              "value": 2190
+              "value": 2090
             },
             {
               "id": "0x0010",
@@ -185,7 +185,7 @@
               "id": "0x0012",
               "name": "occupied_heating_setpoint",
               "zcl_type": "int16",
-              "value": 500
+              "value": 2300
             },
             {
               "id": "0x0030",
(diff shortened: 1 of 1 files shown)


puddly:
I wonder....should we drop the requirement to keep duplicates out of the testing DB?

TheJulianJES:
Ideally, we'd only need to differentiate by firmware version and nothing else. But since (v2) quirks directly modify the device and do not keep the `original_signature`, we're in a bit of an bad situation here.
Soo... maybe yes on dropping that requirement? Or we wait until we persist more of the unmodified device with v2 quirks?

puddly:
Ahhh, that's not ideal. I think it would be best to fix that ASAP and just cache the original device signature at appdb load time?

TheJulianJES:
Yeah. I think we already stop persisting any updates to the "device signature" in the DB [here*](https://github.com/zigpy/zigpy/blob/f7550584b04d0df2aa8af3514ec0931fb6fee6d5/zigpy/appdb.py#L461-L470), as soon there's a quirk applied. So, when restarting ZHA, the original/unmodified signature should still be available.
I think the issue is more that we can't access that unmodified signature at runtime (from the device object), as the quirk already modified it at that point. With v1 quirks, that really isn't an issue, because we know the original signature from the v1 quirk, but that's not possible with the v2 quirk. 

Like, this [zha/zigbee/device.py#L1511-L1512](https://github.com/zigpy/zha/blob/37d2efbf4aa2b15394d4279f2b105dbd8c4d979b/zha/zigbee/device.py#L1511-L1512) just doesn't work at all (and is what the diagnostics tool uses), as we always clear that in zigpy when initializing `CustomDeviceV2` here: [zigpy/quirks/v2/__init__.py#L84-L85](https://github.com/zigpy/zigpy/blob/f7550584b04d0df2aa8af3514ec0931fb6fee6d5/zigpy/quirks/v2/__init__.py#L84-L85).

*: We do seem to persist a changed node descriptor into the DB though ([here](https://github.com/zigpy/zigpy/blob/f7550584b04d0df2aa8af3514ec0931fb6fee6d5/zigpy/appdb.py#L458-L463))?
It's probably best for the `Device` object to more clearly differentiate between the original clusters/signature and modified ones..?

We may also need to think about that when we eventually allow re-interviewing of existing devices, e.g. after an OTA update. They'll have a quirk applied when we re-interview it, so we need to make sure we do persist the new signature to the DB, but the unmodified variant, not the quirked variant. I guess that's for a later date..

PR #624, opened 2026-01-14
Update default precision for instantaneous demand + current

TheJulianJES:
This updates the suggested display precision for the instantaneous demand entity and the RMS current entity.

The instantaneous demand entity then aligns with the EM active power entities (suggested precision 1).
The RMS current entity uses suggested precision 2 to not cut off low-power draw devices, often attached to smart plugs. HA would already use 2 by default, but we currently override this to 1 in the Base EM sensor class, which is not ideal.

### DIFF
--- a/tests/data/devices/adeo-sin-4-fp-21-equ.json
+++ b/tests/data/devices/adeo-sin-4-fp-21-equ.json
@@ -552,7 +552,7 @@
           "endpoint_id": 1,
           "available": true,
           "group_id": null,
-          "suggested_display_precision": null,
+          "suggested_display_precision": 1,
           "unit": "W"
         },
         "state": {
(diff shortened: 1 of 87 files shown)


TheJulianJES:
**TL;DR:** IMO we should merge this now but go through all EM related sensors and see if the HA defaults are better at a later date.

---

I think both changes make sense. We set `_attr_suggested_display_precision = 1` in the base EM class. We should reevaluate doing that, since HA has better defaults depending on device class + unit.

For current, it would already use 2 decimals, but due to the EM base class, it's set to 1. So we just fix it to 2 decimals, just for current now.
I don't want to unintentionally affect a bunch of entities by setting everything to `None` and letting HA use its defaults (at the moment). That wouldn't show up in our diagnostics. I'll reevaluate all these suggested precisions in the future.

For instantaneous demand, HA uses 0 decimals currently. That seems too little, as a lot of smart plugs offer higher resolution than that. I guess that's why we also set it to 1 for all EM entities. So to match that, it's set to 1 in this PR.

TheJulianJES:
Merging this per the above comment. It's better aligned with HA standards and existing power sensors.

PR #625, opened 2026-01-15
Combine OTA release notes when possible

TheJulianJES:
## Proposed change

This changes the update entity to combine the release notes of multiple updates when we jump to the most recent one, 
skipping an install of an intermediate update. A test for this is also added. 

A header with the hex version of the update is also added in the update entity release notes.

## Additional information

This is useful for the zigpy-ota OTA provider, as it keeps a history of all images in the zigpy index, even older ones.
Zigpy will only install the newest image possible, but we still list older release notes now.

### Notes

- Do note there could still be a newer update that shows after this one is installed, e.g. due to `min_file_version` constraints. It's expected that those release notes don't show up during the first install, since that update isn't installed. They'll show up when updates are checked for again. But I guess this is obvious.
- Also note that the friendly version strings ("1.2.3") in the screenshots below are part of the release notes. They are not saved in the index, since there's no consistent way to extract them from OTA info nor are they consistently provided by manufacturers. If we have them, we can try to include them in the release notes directly, like we did with Innr images and in the example screenshots below.
  Only the headers with the hex version strings are included automatically.
- The release notes are ordered from newest to oldest. I think this is what HA also does for add-on updates.

### Screenshots

Changelog with three firmware versions that are skipped | Changelog with one firmware version that is skipped | Changelog with no firmware versions that are skipped (no header) (not done in this PR now) | Changelog with no firmware versions that are skipped (with header) (done in this PR now)
--- | --- | --- | ---
Bildschirmfoto 2026-01-16 um 02 15 59 | Bildschirmfoto 2026-01-16 um 02 19 29 | Bildschirmfoto 2026-01-16 um 02 18 26 | Bildschirmfoto 2026-01-16 um 03 14 13

### Markdown example of what we pass to HA

```markdown
## 0x12345696
Release notes for v3.

## 0x1234568c
Release notes for v1.
```

### DIFF
--- a/tests/test_update.py
+++ b/tests/test_update.py
@@ -620,7 +620,7 @@ async def test_firmware_update_latest_version_even_if_downgrade(
 async def test_firmware_update_metadata(zha_gateway: Gateway) -> None:
     """Test ZHA update platform - firmware metadata (changelog, release_notes, release_url)."""
     zigpy_device = zigpy_device_mock(zha_gateway)
-    zha_device, ota_cluster, fw_image, installed_fw_version = await setup_test_data(
+    zha_device, ota_cluster, _, installed_fw_version = await setup_test_data(
         zha_gateway, zigpy_device
     )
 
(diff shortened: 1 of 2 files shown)


TheJulianJES:
I've edited the PR description to include some screenshots of how this looks like in Home Assistant.
For single images, we do not include a version header in the release notes at the moment. I wonder if that is something we should do? (Compare the third screenshot to the other ones)

puddly:
Nice! I think the one that always includes the version header doesn't actually look too bad.

TheJulianJES:
Hadn't attached a concrete example for that, so here's that: (edited the PR description to include fourth image as well)
Changelog with no firmware versions that are skipped (no header) | Changelog with no firmware versions that are skipped (with header)
--- | ---
Bildschirmfoto 2026-01-16 um 02 18 26 | Bildschirmfoto 2026-01-16 um 03 14 13

Especially if the "This updates the device firmware to version 1.2.3" part is missing from release notes (which can happen), I think the big hex version header might be nice to further differentiate the actual release notes from the blue info box?

EDIT: Switched to the second variant, always including the header now. I think it's ok. [`73428db` (#625)](https://github.com/zigpy/zha/pull/625/commits/73428dbddfc1ce22b3765d301a38c9e2de1d2275)

tests/test_update.py
```diff
@@ -620,7 +620,7 @@ async def test_firmware_update_latest_version_even_if_downgrade(
 async def test_firmware_update_metadata(zha_gateway: Gateway) -> None:
     """Test ZHA update platform - firmware metadata (changelog, release_notes, release_url)."""
     zigpy_device = zigpy_device_mock(zha_gateway)
-    zha_device, ota_cluster, fw_image, installed_fw_version = await setup_test_data(
+    zha_device, ota_cluster, _, installed_fw_version = await setup_test_data(
```
TheJulianJES, line 623: Made this small change, as we don't actually use the `fw_image` from `setup_test_data` in these tests. In this first test, `fw_image` is assigned another image a few lines below.  To be less confusing, it's probably better not to reuse the variable name / instantly overwrite that again, hence just using `_` now. We can also use something different if we don't like this.

PR #635, opened 2026-01-29
Fix `OnOff` client cluster commands not updating attribute cache

TheJulianJES:
https://github.com/zigpy/zha/pull/615 removed `handle_on_off_output_cluster_exception`, so the `OnOffClusterHandler` is no longer used as a cluster handler for client/output devices. However, the `cluster_command` logic for `on_with_timed_off` was only present there.

This PR adds it to the `OnOffClientClusterHandler` to fix the issue of motion binary sensors not updating on cluster commands. The logic is also removed from the (server) `OnOffClusterHandler`, as that should (hopefully) not be used anymore.


TheJulianJES:
We may need a similar fix for alarm control panels, though they were already broken somewhat before these changes.

tests/test_binary_sensor.py
```diff
@@ -287,3 +289,91 @@ async def test_quirks_binary_sensor_attr_converter(zha_gateway: Gateway) -> None
...
+    on_off_ch = zha_device.endpoints[1].client_cluster_handlers["1:0x0006_client"]
+    assert on_off_ch is not None
+
+    # Simulate motion sensor sending on_with_timed_off command
+    # on_off_control=0 means always accept, on_time=1800 (180 seconds in 10ths)
+    on_off_ch.cluster_command(
+        106,  # tsn
+        OnOff.ServerCommandDefs.on_with_timed_off.id,
+        [0, 1800, 0],  # on_off_control, on_time, off_wait_time
+    )
```
TheJulianJES: We probably shouldn't directly call the `cluster_command` method on the client cluster handler.
Instead, we can use `listener_event` with `"cluster_command"` or use `make_zcl_header`:
```python
# Option 1: call `listener_event` with `"cluster_command"`, similar to existing `async_test_iaszone_on_off` test
cluster.listener_event(
    "cluster_command", 1, OnOff.ServerCommandDefs.on_with_timed_off.id, [0, 1800, 0]
)

# Option 2: Use `make_zcl_header`
hdr = make_zcl_header(
    OnOff.ServerCommandDefs.on_with_timed_off.id, global_command=False
)
cluster.handle_message(hdr, [0, 1800, 0])
```

What do we prefer? `handle_message ` with `make_zcl_header`, so we're not relying on `listener_event` `cluster_command` staying the same over time?

tests/test_cluster_handlers.py
```diff
@@ -1241,8 +1241,19 @@ async def test_zha_send_event_from_quirk(zha_gateway: Gateway):
...
+        call(
+            "attribute_updated",
+            {
+                "attribute_id": 0,
+                "attribute_name": "on_off",
+                "attribute_value": t.Bool.true,
+                "value": t.Bool.true,
+            },
+        ),
+    ]
```
TheJulianJES, line 1256: Remotes will also update the attribute cache now. Previously, the `handle_on_off_output_cluster_exception` only created the server `OnOffClusterHandler` for binary sensor platform entities.

Now, we (correctly) only use `OnOffClientClusterHandler` everywhere. But since remotes and motion sensors are essentially the same, we also update the `on_off` attribute in our cache now.

I think this should be fine? We don't create a binary sensor entity for them, though this is another emitted ZHA event. In the future, we can add a config option to the ZHA UI to not show them in the logbook on the HA side, as that's been heavily requested anyway.

Unless we want to make really weird exceptions and couple cluster handlers and entities again, we could only check the device type and avoid `update_attribute` then?
Do we want that or just keep as is and see if anyone complains?

puddly: We may have default event entities for some devices in the future with https://github.com/zigpy/zha/pull/594 so I think it's fine to have a slightly chatty logbook for now

tests/test_binary_sensor.py
TheJulianJES: I've opted to use `handle_message` with `make_zcl_header` in the tests now. It's different to the `async_test_iaszone_on_off` test, which also checks cluster handler functionality, but that can be changed in the future. This also tests more of the "whole stack".

zha/zigbee/cluster_handlers/general.py
```diff
@@ -594,46 +646,6 @@ async def turn_off(self) -> None:
...
-                    )
-        elif cmd == "toggle":
-            self.cluster.update_attribute(
-                OnOff.AttributeDefs.on_off.id, not bool(self.on_off)
-            )
-
-    def set_to_off(self, *_):
-        """Set the state to off."""
-        self._off_listener = None
-        self.cluster.update_attribute(OnOff.AttributeDefs.on_off.id, t.Bool.false)
```
TheJulianJES, line 635: I've removed the entire `cluster_command` handling for the server `OnOff` cluster handler. Unfortunately, it was never covered by tests, hence we didn't catch this.

I don't think it can be used at all for the server handler, so should be fine to remove.

TheJulianJES:
Tested these changes with an IKEA TRADFRI motion sensor and it works properly again.

puddly (approved):
Thanks!

PR #637, opened 2026-01-29
Update ZHA manufacturer code handling for zigpy 0.91.x

TheJulianJES:
This PR fixes an issue where multiple read/write methods overwrote the default manufacturer code with `None`, even though we should now be using `UNDEFINED` with the recent zigpy 0.91.x changes.

Since we use the `UNDEFINED` default in ZHA for reading attributes (during initialization), this also works around the issue of quirks overriding `read_attributes` with `manufacturer=None` as the default. 

Due to the improved manufacturer code handling in zigpy, we do not need to determine the manufacturer code ourselves anymore. It should also fix an issue where batch-reading attributes during initialization always used a manufacturer code or not, messing up mixed reads.

`manufacturer=UNDEFINED` calls are expected in a lot of places in tests now.

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -127,30 +127,75 @@ def make_attribute(attrid: int, value: Any, status: int = 0) -> zcl_f.Attribute:
 
 
 async def send_attributes_report(
-    zha_gateway: Gateway, cluster: zigpy.zcl.Cluster, attributes: dict
+    zha_gateway: Gateway,
+    cluster: zigpy.zcl.Cluster,
+    attributes: dict[str | int | zcl_f.ZCLAttributeDef, Any],
+    *,
+    tsn: int | None = None,
 ) -> None:
-    """Cause the sensor to receive an attribute report from the network.
+    """Mock attribute reports on a cluster."""
+    reports = []
+    manufacturer_codes: set[int | None] = set()
 
-    This is to simulate the normal device communication that happens when a
-    device is paired to the zigbee network.
-    """
-    attrs = []
-
-    for attrid, value in attributes.items():
-        if isinstance(attrid, str):
-            attrid = cluster.attributes_by_name[attrid].id
+    for attr, value in attributes.items():
+        if isinstance(attr, int):
+            # Raw attribute ID for unknown attributes
+            manufacturer_codes.add(None)
+            reports.append(
+                zcl_f.Attribute(
+                    attrid=attr,
+                    value=zcl_f.TypeValue(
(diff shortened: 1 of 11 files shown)


puddly:
Thanks! I've reverted the changes in the other PR to keep it small.

puddly:
It looks like the other PR still needs the `send_attributes_report` from zigpy, as we do not account for the manufacturer code when creating the attribute reports.

TheJulianJES:
It's also needed in this PR, so uh, we can put it in this one and rename it? (or create a separate one)

puddly:
Let's put it into this one

TheJulianJES:
The remaining error with the Tuya `FrostLockResetButton` seems a bit weird. `TuyaClusterData` seems to only use the manufacturer code for us. I guess we can just change the type there to also include `UndefinedType`. Still taking a look.. 
We may also wanna run zha-quirks tests (especially the fragile Tuya conversion) with the latest zigpy changes.

Failing test here:
```python
 tests/test_button.py::test_frost_unlock - ValueError: Failed to convert manufacturer=<UndefinedType._singleton: 0> from type <enum 'UndefinedType'> to int | None
```

puddly:
Looks like a few fail:

```
FAILED tests/test_bosch.py::test_bosch_radiator_thermostat_II_write_attributes - KeyError: None
FAILED tests/test_bosch.py::test_bosch_radiator_thermostat_II_read_attributes_paused - KeyError: None
FAILED tests/test_xiaomi.py::test_xiaomi_e1_thermostat_rw_redirection[system_mode-unoccupied_heating_setpoint] - KeyError: None
FAILED tests/test_xiaomi.py::test_xiaomi_e1_thermostat_rw_redirection[28-20] - KeyError: None
```

TheJulianJES:
If it's only the attribute redirection, I think that shouldn't affect ZHA with this PR, since we explicitly pass `UNDEFINED` now (which is also not ideal, but we can remove that when quirks are all updated I guess).

I'll take a look at zha-quirks, since it needs to be updated for the `TuyaClusterData` change anyway. We may be able to just update the defaults for the `manufacturer` in `read_attributes`/`write_attributes`.

puddly:
I think most methods that override attribute reading/writing can probably pass through `**kwargs`, without needing to care about `manufacturer`

puddly:
ZHA tests pass with the above quirks PR so I think this should resolve the `manufacturer` problem!

puddly (approved):

PR #645, opened 2026-02-05
Fix missing humidity entity on Centralite 3310-G

puddly:
This one device uses a cluster handler for an overridden cluster whose name is `humidity`. The generic cluster handler matching logic did not work for it and we need to explicitly state we want the generic cluster handler ID. No other uses like this seem to exist.

https://github.com/home-assistant/core/issues/162340

### DIFF
--- a/tests/data/devices/centralite-3310-g.json
+++ b/tests/data/devices/centralite-3310-g.json
@@ -441,7 +441,7 @@
           "entity_category": null,
           "entity_registry_enabled_default": true,
           "enabled": true,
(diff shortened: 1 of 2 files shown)


puddly:
Closed in favor of https://github.com/zigpy/zha-device-handlers/pull/4736

TheJulianJES:
Would be nice to still add the diagnostics file from this sensor

EDIT: Now in:
-  https://github.com/zigpy/zha/pull/649

PR #647, opened 2026-02-06
Fix platform overrides

puddly:
The discovery rewrite this release introduced a few bugs:
- The unique ID computation from the Light platform was not used when switching to a Switch, causing the swapped entities to be recreated.
- Overrides were applied to entities not belonging to a feature group and ended up replacing the entire set of entities without a feature group.

This PR removes the weight based boosting system and replaces it with explicit override logic.

### DIFF
--- a/tests/test_discover.py
+++ b/tests/test_discover.py
@@ -127,17 +127,22 @@ async def test_device_override_entities(zha_gateway: Gateway) -> None:
     }
 
     zha_device = await join_zigpy_device(zha_gateway, zigpy_device)
+
+    # The light is gone
+    with pytest.raises(KeyError):
+        get_entity(zha_device, platform=Platform.LIGHT)
+
+    # And has been replaced by a switch with the same unique ID
+    switch = get_entity(zha_device, platform=Platform.SWITCH)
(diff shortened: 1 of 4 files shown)


tests/test_discover.py
```diff
@@ -110,6 +111,94 @@ async def test_device_override(
...
+    expected_loaded_device_data = device_data
+
+    # The light is gone
+    light_entities = expected_loaded_device_data["zha_lib_entities"].pop("light")
+    assert len(light_entities) == 1
+
+    # And has been replaced with a single switch
+    expected_loaded_device_data["zha_lib_entities"]["switch"] = [
+        loaded_device_data["zha_lib_entities"]["switch"][0]
+    ]
```
TheJulianJES, line 148: Comparing diagnostics text does feel a bit weird, but if you're good with that, I'm good as well.

TheJulianJES (approved):
Looks good!

tests/test_discover.py
puddly: I guess the alternative would be to compare the `info_object`?

PR #648, opened 2026-02-07
Fix configuring attribute reporting

TheJulianJES:
## Proposed change

This fixes configuration of attribute reporting after the recent zigpy manufacturer code changes.
- `configure_reporting_multiple` is now called with `ZCLAttributeDef` and `ReportingConfig`, instead of the attribute name (`str`) and a tuple containing min/max/reportable change int values.
- `_configure_reporting_status` was updated to get the attribute name from `ZCLAttributeDef`.
- The `cluster_id >= 0xFC00` check for using the manufacturer code is removed, as zigpy now does this better and automatically, grouping respective attributes properly. 
- The tests were updated to expect the new `configure_reporting_multiple` calls.
- The condition `res[0].status == Status.SUCCESS and len(res) == 1` was updated to `len(res) == 1 and res[0].status == Status.SUCCESS`. This should now properly work with empty lists.
  - EDIT:  Zigpy should never send us an empty list though. I think it's still fine to not catastrophically break in the case zigpy sends something unexpected though, as that's just for the reporting status. We'll just mark the reports as failed then.
- A test was added for `configure_reporting_status`, which sends reports on attribute reports used by HA frontend.

This change was tested with an IKEA motion sensor, where attribute reporting is now correctly configured for the `PowerConfiguration` cluster again. This is also represented in the HA frontend.

## Additional information

Related PRs:
- https://github.com/zigpy/zigpy/pull/1719
- https://github.com/zigpy/zha/pull/605
- https://github.com/zigpy/zha/pull/637

### DIFF
--- a/tests/test_cluster_handlers.py
+++ b/tests/test_cluster_handlers.py
@@ -33,6 +33,7 @@ from zigpy.zcl.clusters.general import (
 )
 from zigpy.zcl.clusters.homeautomation import Diagnostic
 from zigpy.zcl.clusters.measurement import TemperatureMeasurement
+from zigpy.zcl.helpers import ReportingConfig
 import zigpy.zdo.types as zdo_t
 
 from tests.common import (
@@ -303,7 +304,7 @@ async def test_in_cluster_handler_config(
     reported_attrs = set()
 
     for mock_call in cluster.configure_reporting_multiple.mock_calls:
-        reported_attrs.update(mock_call.args[0].keys())
+        reported_attrs.update(attr_def.name for attr_def in mock_call.args[0])
 
     assert attrs == reported_attrs
     assert cluster.configure_reporting.call_count == 0
@@ -941,18 +942,19 @@ async def test_configure_reporting(zha_gateway: Gateway) -> None:
     cluster_handler = TestZigbeeClusterHandler(cluster, endpoint)
     await cluster_handler.async_configure()
 
-    # Since we request reporting for five attributes, we need to make two calls (3 + 1)
+    # Since we request reporting for four attributes, we need to make two calls (3 + 1)
+    Color = zigpy.zcl.clusters.lighting.Color
     assert cluster.configure_reporting_multiple.mock_calls == [
         mock.call(
             {
-                "current_x": (1, 60, 1),
-                "current_hue": (1, 60, 2),
-                "color_temperature": (1, 60, 3),
+                Color.AttributeDefs.current_x: ReportingConfig(1, 60, 1),
+                Color.AttributeDefs.current_hue: ReportingConfig(1, 60, 2),
+                Color.AttributeDefs.color_temperature: ReportingConfig(1, 60, 3),
             }
         ),
         mock.call(
             {
-                "current_y": (1, 60, 4),
+                Color.AttributeDefs.current_y: ReportingConfig(1, 60, 4),
             }
         ),
     ]
--- a/tests/test_cluster_handlers.py
+++ b/tests/test_cluster_handlers.py
@@ -33,6 +33,7 @@ from zigpy.zcl.clusters.general import (
     PowerConfiguration,
 )
 from zigpy.zcl.clusters.homeautomation import Diagnostic
+from zigpy.zcl.clusters.lighting import Color
 from zigpy.zcl.clusters.measurement import TemperatureMeasurement
 from zigpy.zcl.helpers import ReportingConfig
 import zigpy.zdo.types as zdo_t
@@ -177,7 +178,7 @@ def endpoint_mock(zigpy_coordinator_device: ZigpyDevice) -> Endpoint:
         ),
         (zigpy.zcl.clusters.hvac.Fan.cluster_id, 1, {"fan_mode"}),
         (
-            zigpy.zcl.clusters.lighting.Color.cluster_id,
+            Color.cluster_id,
             1,
             {
                 "current_x",
@@ -926,7 +927,7 @@ async def test_configure_reporting(zha_gateway: Gateway) -> None:
     mock_ep.profile_id = zigpy.profiles.zha.PROFILE_ID
     mock_ep.device.zdo = AsyncMock()
 
-    cluster = zigpy.zcl.clusters.lighting.Color(mock_ep)
+    cluster = Color(mock_ep)
     cluster.bind = AsyncMock(
         spec_set=cluster.bind,
         return_value=[zdo_t.Status.SUCCESS],  # ZDOCmd.Bind_rsp
@@ -944,7 +945,7 @@ async def test_configure_reporting(zha_gateway: Gateway) -> None:
     await cluster_handler.async_configure()
 
     # Since we request reporting for four attributes, we need to make two calls (3 + 1)
-    Color = zigpy.zcl.clusters.lighting.Color
+
     assert cluster.configure_reporting_multiple.mock_calls == [
         mock.call(
             {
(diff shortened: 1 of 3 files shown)


zha/zigbee/cluster_handlers/__init__.py
```diff
@@ -429,26 +429,27 @@ async def configure_reporting(self) -> None:
...
+                attr_names,
                 self.name,
                 res,
             )
-            for attr in attrs:
-                event_data[attr]["status"] = Status.FAILURE.name
+            for attr_name in attr_names:
+                event_data[attr_name]["status"] = Status.FAILURE.name
             return
-        if res[0].status == Status.SUCCESS and len(res) == 1:
```
TheJulianJES, line 448: Swapping the conditions here may fix a long-standing issue occasionally seen in logs 😅 
EDIT: Looking at this more closely, should we even get an empty list from zigpy here..?

TheJulianJES: Looks like there's a bug in zigpy. We always return an empty list: https://github.com/zigpy/zigpy/blob/b2ee3c3b1a6b1f0c1414ff3f4ec68676d93fd536/zigpy/zcl/__init__.py#L1443

puddly: Ahh, we never actually extend the combined results list 😅

TheJulianJES: Should be fixed by https://github.com/zigpy/zigpy/pull/1764

TheJulianJES:
We can add a test for `configure_reporting_status` in this PR or another PR. For example: https://github.com/zigpy/zha/compare/TheJulianJES:tjj/fix_attr_reporting...TheJulianJES:tjj/fix_attr_reporting_status_tests

TheJulianJES:
@puddly Want me to just add the above test to this PR or make another one? (test for reporting status used by frontend)

puddly:
Let's roll it into this one

TheJulianJES:
Do note the updated tests in this PR still would not really catch an issue where the method signature of zigpy's `cluster.configure_reporting_multiple` changes. I guess we'd have to properly test the `configure_reporting_multiple` method in ZHA then and check the final zigpy requests.. 

I'd keep this PR as is now.

tests/test_cluster_handlers.py
```diff
@@ -941,23 +950,145 @@ class TestZigbeeClusterHandler(ClusterHandler):
...
+                    status=foundation.Status.SUCCESS
+                )
+            ],
+            {"current_x": "SUCCESS", "current_y": "SUCCESS"},
+        ),
+        # Empty list: all attributes assumed successful (no failures reported)
+        (
+            [],
+            {"current_x": "SUCCESS", "current_y": "SUCCESS"},
+        ),
```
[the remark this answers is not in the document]
TheJulianJES: Oh, this should probably fail for the reporting status, since zigpy should never report an empty list (with https://github.com/zigpy/zigpy/pull/1764).

zha/zigbee/cluster_handlers/__init__.py
```diff
@@ -429,26 +429,27 @@ async def configure_reporting(self) -> None:
...
-        attrs: dict[str, tuple[int, int, float | int]],
-        res: list | tuple,
+        attrs: dict[ZCLAttributeDef, ReportingConfig],
+        res: list[ConfigureReportingResponseRecord],
         event_data: dict[str, dict[str, Any]],
     ) -> None:
         """Parse configure reporting result."""
-        if isinstance(res, (Exception, ConfigureReportingResponseRecord)):
+        attr_names = {attr_def.name for attr_def in attrs}
+        if not res or isinstance(res, (Exception, ConfigureReportingResponseRecord)):
```
TheJulianJES: We should not be able to hit the `isinstance(res, (Exception, ConfigureReportingResponseRecord))` check ever now? Like, returning an exception or just a single `ConfigureReportingResponseRecord`? Should I just remove this?

If so, should I at least keep the guard against zigpy returning an empty list? Even though that should not happen after https://github.com/zigpy/zigpy/pull/1764/ anymore. We'd just mark the attributes as failed to configure if zigpy (or some test mocks elsewhere) somehow return an empty list. Or should we just ignore that and fail however that would fail? (causing `SUCCESS`)

zha/zigbee/cluster_handlers/__init__.py
```diff
@@ -477,14 +478,14 @@ def _configure_reporting_status(
...
         self.debug(
             "Successfully configured reporting for '%s' on '%s' cluster",
-            set(attrs) - set(failed),
+            success,
             self.name,
         )
-        for attr in success:
-            event_data[attr]["status"] = Status.SUCCESS.name
+        for attr_name in success:
+            event_data[attr_name]["status"] = Status.SUCCESS.name
```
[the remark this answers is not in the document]
TheJulianJES: Addressed. We now return failure to configure for all attributes then, even though this should not happen anymore after https://github.com/zigpy/zigpy/pull/1764/.

zha/zigbee/cluster_handlers/__init__.py
```diff
@@ -399,12 +394,17 @@ async def configure_reporting(self) -> None:
             to_configure[REPORT_CONFIG_ATTR_PER_REQ:],
         )
         while chunk:
-            reports = {rec["attr"]: rec["config"] for rec in chunk}
+            reports = {
+                self.cluster.find_attribute(rec["attr"]): ReportingConfig(
+                    *rec["config"]
+                )
+                for rec in chunk
+            }
```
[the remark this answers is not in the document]
TheJulianJES: If this is an issue here, it will be more of an issue elsewhere, so I think we can ignore this.

tests/test_cluster_handlers.py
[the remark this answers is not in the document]
TheJulianJES: Addressed now.

zha/zigbee/cluster_handlers/__init__.py
TheJulianJES: I've removed it with [`182a8bc` (this PR)](https://github.com/zigpy/zha/pull/648/commits/182a8bc1697ab4631a3b6eee3cf82f8d38e69d3f). Having zigpy return an exception or a single status report can't happen at all from what I see. It's always a list at least.

zha/zigbee/cluster_handlers/__init__.py
```diff
@@ -429,26 +429,26 @@ async def configure_reporting(self) -> None:
...
-                attrs,
+                attr_names,
                 self.name,
                 res,
             )
-            for attr in attrs:
-                event_data[attr]["status"] = Status.FAILURE.name
+            for attr_name in attr_names:
+                event_data[attr_name]["status"] = Status.FAILURE.name
             return
```
TheJulianJES, line 447: Technically, zigpy should also never return an empty list. Do we want to keep this failure case in case zigpy breaks again, so we report all attributes as failed to configure then?

If we remove this, they'd report as being successfully configured otherwise, which also isn't great. But this should never happen (again), soo..

puddly: I think we can keep it in. In the near future I'd like to persist the attribute configuration state in the database so this code likely will be reworked anyways.

zha/zigbee/cluster_handlers/__init__.py
puddly: Yeah, returning exception objects (instead of raising them) is a legacy artifact and something that we've slowly been trying to get rid of. There are more instances within the ZHA library that still account for this possibility. I believe we do still have a few `asyncio.gather(..., return_exceptions=True)` invocations that were the cause of this pattern.

puddly:
I've tested this PR as well (along with all of the other zigpy + ZHA changes) and it works as expected.

jclsn:
The fixes haven't landed in 26.2.2, have they?

puddly:
They have.

jclsn:
Ah found it https://github.com/home-assistant/core/pull/162894

jclsn:
Yeah works again. Firmware versions are still unknown, but I guess this is related to something else?

PR #650, opened 2026-02-13
Re-add `CLUSTER_HANDLER_ONLY_CLUSTERS` handling for binding

puddly:
This feature was accidentally removed during the recent discovery refactor. Without it, binding and reporting does not happen for cluster handlers not claimed by any entity. In this case, it introduced a regression for a Hue remote (and other similar devices).


TheJulianJES (approved):
Looks good! We should really have v2 quirks register these clusters for binding in the future..

PR #659, opened 2026-02-19
Add Ballast min + max level entities

TheJulianJES:
## Proposed change

This adds minimum level and maximum level entities for the Ballast cluster. This is used by dimmers.

## Additional information

Diagnostics for the Ubisys D1 dimmer will be added to test this.
We'll need to add translation keys to HA.


TheJulianJES (commented):
There's a `minimum_level` translation key with "Minimum load dimming level" as the translation in HA for the Inovelli devices: [homeassistant/components/zha/strings.json#L731-L733](https://github.com/home-assistant/core/blob/37f0f1869f5a564813da98150f7e9d71ac1c87d2/homeassistant/components/zha/strings.json#L731-L733)

I think `minimum_dimming_level` with "Minimum dimming level" as the translation makes more sense. So, that's what we should add to HA IMO. We can change the Inovelli one in the future (if it has the same meaning – but if so, why didn't they use the `Ballat` cluster then?)

TheJulianJES:
Confirmed to work with Ubisys D1 dimmer. I'll add diagnostics dumps for all Ubisys deviecs in a follow-up PR when the quirks bump and all relevant ZHA entity changes are merged.

PR #660, opened 2026-02-19
Add exposed feature for SE summation polling

TheJulianJES:
## Proposed change

This adds separate entity classes based on an exposed quirks feature/hint, so that the Smart Metering's `current_summ_delivered` and `current_summ_received` attributes are polled automatically by ZHA then.

## Additional information

The `"se_poll_summation"` exposed feature will be added to zha-quirks. Then, this PR will be changed to use that exposed feature constant.

Weirdly enough, the Zigbee spec doesn't specify that these two attributes are reportable, yet some devices seem to support it. For the EM cluster, we always poll by default. For the SE cluster, we only poll for select models `"TS011F", "ZLinky_TIC", "TICMeter"`. These can (mostly) be switched to instead expose the polling feature/hint in quirks soon.

Ubisys diagnostics will be added soon. That should be enough to cover these new entities IMO, since the polling is the same and already tested elsewhere. The diagnostics will contain the class name, so check if the exposed feature matching works in this case.

### DIFF
--- a/tests/data/devices/ubisys-s1-5501.json
+++ b/tests/data/devices/ubisys-s1-5501.json
@@ -994,7 +994,7 @@
           "unique_id": "ab:cd:ef:12:e4:d6:5c:c2-3-1794-summation_delivered",
           "migrate_unique_ids": [],
           "platform": "sensor",
-          "class_name": "SmartEnergySummation",
+          "class_name": "ExposedFeaturePolledSmartEnergySummation",
           "translation_key": "summation_delivered",
           "translation_placeholders": null,
           "device_class": "energy",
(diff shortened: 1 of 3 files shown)


TheJulianJES:
This is updated to use a constant for the exposed feature from quirks. Tests will pass with the bump, though diagnostics will need to be regenerated then.

I'll mark this as ready for review, so it's not forgotten about. This is required for Ubisys devices to work properly.
(Having this as an exposed feature also allows us to include this in other quirks to hint that ZHA nedes to poll the SE entities. Right now, we hardcode some models in ZHA.)

TheJulianJES:
I'll add all the Ubisys diagnostics in a follow-up PR in the next few days. It's not critical for the release.

PR #662, opened 2026-02-19
Migrate all platforms to abstract base classes

puddly:
We've already migrated the larger platforms, this migrates the rest. I've also removed `**kwargs` from the `async_` method calls for stronger typing.

This is just a code quality improvement and should have no external effects.


TheJulianJES:
Had a quick look at this and everything seems fine

PR #667, opened 2026-02-21
Add quirks v2 siren entity based on `SwitchMetadata`

TheJulianJES:
DRAFT. TODO: Requires zigpy change adding `SIREN = "siren"` to `EntityPlatform`.

Based on:
- https://github.com/zigpy/zha/pull/665

## Proposed change

This allows quirks v2 entities to provide `entity_platform=EntityPlatform.SIREN` for `.switch()` entities to create a siren entity on the HA side, instead of a switch entity.

## Additional information

A lot of the logic and tests are copied from the switch entity, but there's no nice way around that. A downside of using `SwitchMetadata` is that we also support the "inverter attribute" logic which is not really necessary in all cases.. I thought of just ignoring that for the siren platform, but it's kept in here for now.
The logic is more similar to `ConfigurableAttributeSwitch` and its tests then.

It's a bit weird doing `.switch` to create a siren entity, but it's very similar to what we do with `.enum` to crate an enum sensor entity when providing a different `entity_platform` (and `entity_type` due to HA restrictions).

### DIFF
--- a/tests/test_siren.py
+++ b/tests/test_siren.py
@@ -8,6 +8,7 @@ from zigpy.profiles import zha
 from zigpy.quirks import DEVICE_REGISTRY
 from zigpy.quirks.v2 import CustomDeviceV2, QuirkBuilder
 from zigpy.quirks.v2.homeassistant import EntityPlatform
+import zigpy.types as t
 from zigpy.typing import UNDEFINED
 from zigpy.zcl.clusters import general, security
 import zigpy.zcl.foundation as zcl_f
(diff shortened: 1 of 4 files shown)


dmulcahey:
Maybe we add another method to the builder to make it clearer we are making a siren ex: `siren_from_switch`? or if this doesn't generalize well maybe something more generic like `entity(entity_cls, ...)` and we specify the actual class being created and what is needed? I feel like what is in the test for the builder:

```python
QuirkBuilder(zigpy_dev.manufacturer, zigpy_dev.model)
    .switch(
        general.Basic.AttributeDefs.power_source.name,
        general.Basic.cluster_id,
        on_value=1,
        off_value=0,
        entity_platform=EntityPlatform.SIREN,
        translation_key="siren",
        fallback_name="Siren",
    )
    .add_to_registry()
)
```
makes sense to us but maybe will confuse potential contributors.

PR #669, opened 2026-02-21
Add `OccupancySensing` PIR delay entities

TheJulianJES:
TODO: Translations in HA!

## Proposed change

This adds config number entities for the `OccupancySensing` `pir_o_to_u_delay` and `pir_u_to_o_delay` attributes. 
The entity class for the Sonoff sensors (`ultrasonic_o_to_u_delay`) is also moved to be near below the PIR delay classes.

## Additional information

This would avoid introducing quirks v2 entities, e.g. for the Hue sensors or the Frient sensor below:
- https://github.com/zigpy/zha-device-handlers/pull/4427

IIRC there was an issue with Hue sensors returning this attribute was unsupported during pairing, but maybe that was only an issue because of the manufacturer code stuff we had going on. I'll need to recheck that.

### Sonoff specific entity – `ultrasonic_o_to_u_delay`

In the future, we may wanna look at exposing that entity for all devices, just with the limits removed.
All these entities will end up using the `BOX` `NumberMode` in HA anyway (where you can even enter numbers outside min/max IIRC), so these limits are more artificial.

TODO: Also look at attributes in `SonoffPresenceSenorTimeout` (device class + unit of measurement + unnecessary attributes). This can be done in a separate PR.

### DIFF
--- a/zha/application/platforms/number/__init__.py
+++ b/zha/application/platforms/number/__init__.py
@@ -1226,3 +1226,37 @@ class SinopeLightLEDOffIntensityConfigurationEntity(NumberConfigurationEntity):
             }
         ),
     )
+
+
+@register_entity(OccupancySensing.cluster_id)
+class PIROccupiedToUnoccupiedDelayConfigurationEntity(NumberConfigurationEntity):
+    """Representation of a ZHA PIR occupied to unoccupied delay configuration entity."""
+
+    _unique_id_suffix = "pir_o_to_u_delay"
+    _attr_native_min_value: float = 0x0000
+    _attr_native_max_value: float = 0xFFFF
+    _attr_device_class = NumberDeviceClass.DURATION
+    _attr_native_unit_of_measurement: str = UnitOfTime.SECONDS
+    _attribute_name = "pir_o_to_u_delay"
+    _attr_translation_key: str = "pir_o_to_u_delay"
+
+    _cluster_handler_match = ClusterHandlerMatch(
+        cluster_handlers=frozenset({CLUSTER_HANDLER_OCCUPANCY})
+    )
+
+
+@register_entity(OccupancySensing.cluster_id)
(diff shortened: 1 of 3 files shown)


dmulcahey:
LGTM

zha/application/platforms/number/__init__.py
```diff
@@ -445,6 +445,59 @@ def recompute_capabilities(self) -> None:
...
+    _attribute_name = "pir_u_to_o_delay"
+    _attr_translation_key: str = "pir_u_to_o_delay"
+
+    _cluster_handler_match = ClusterHandlerMatch(
+        cluster_handlers=frozenset({CLUSTER_HANDLER_OCCUPANCY})
+    )
+
+
+@register_entity(OccupancySensing.cluster_id)
+class SonoffPresenceSenorTimeout(NumberConfigurationEntity):
```
TheJulianJES, line 483: (The Sonoff class was just moved. It's not new.)

(This is something to either introduce for all devices in the future or move to quirks. Something for another PR.)

TheJulianJES (commented):
We don't need to explicitly set `NumberMode.BOX` for the newly added entities, as there's over 256 steps, so the HA frontend will automatically use the BOX number mode.

zha/zigbee/cluster_handlers/measurement.py
```diff
@@ -85,6 +85,10 @@ class OccupancySensingClusterHandler(ClusterHandler):
             config=REPORT_CONFIG_IMMEDIATE,
         ),
     )
+    ZCL_INIT_ATTRS = {
+        "pir_o_to_u_delay": True,
+        "pir_u_to_o_delay": True,
+    }
```
TheJulianJES, line 91: Ah, I could have used `AttributeDefs.xxx.name` here. Oh well, this will be cleaned up with an eventual cluster handler refactor/removal anyway, but can also change in another PR before that. This follows existing patterns and there's no functional difference 😄

PR #670, opened 2026-02-22
Add Electrical Measurement reactive power entity

TheJulianJES:
## Proposed change

This adds the reactive power entity based on the `ElectricalMeasurement` cluster.

## Additional information

Supported by Ubisys devices (read/polling only).
For now, I've limited this PR to only include an entity for Phase A, not B, not C, and also not the total reactive power.


dmulcahey:
LGTM

PR #672, opened 2026-02-22
Fix stuck light transitioning flag after task cancellation

TheJulianJES:
The Git diff is horrible to look at, but it's just `try`/`finally` being added in those parts and one small guard (https://github.com/zigpy/zha/pull/672/changes/59611bf18c2ed308c45273248d85a39b6d08d8ce).
EDIT: Git diff is slightly improved by introducing another method where the actual implementation is moved to, with `async_turn_on` now calling that with but wrapped with `try / raise` to release the flag at the end.

## AI summary

When a `mode: restart` automation fires while a light turn-on or turn-off is in progress, Home Assistant cancels the running asyncio task. If that cancellation lands between `async_transition_set_flag()` and `async_transition_start_timer()`, the `_transitioning_individual` flag is left `True` with no timer to clear it. The light then ignores all incoming attribute reports indefinitely.

- Wrap the body of `async_turn_on` and `async_turn_off` in `try/finally` blocks. The `finally` clause calls `async_transition_complete()` when the flag is stuck (set but no timer running), covering task cancellation and unexpected exceptions.
  - EDIT: With the last commit, `async_turn_on` was essentially renamed and the new `async_turn_on` now just calls that method with `try / raise`, so we don't have to indent the entire turn on method. This reduces diff a bit and might make the intention more clear of what we're doing. I haven't done that for `async_turn_off` because it's a much shorter method. I can obviously revert this to just indent everything, also do it for turn off, or just leave as is.
- As a side effect, a pre-existing bug is also fixed: when the second `move_to_level` call in the `new_color_provided_while_off` path failed, the code returned without calling `async_transition_complete()` or starting the transition timer — leaving the flag stuck.
  - An explicit guard is now added there, consistent with all other early-return error paths (https://github.com/zigpy/zha/pull/672/changes/59611bf18c2ed308c45273248d85a39b6d08d8ce). We should consider if we even need these at all.

### DIFF
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -584,12 +584,7 @@ class BaseClusterHandlerLight(BaseLight):
             # If the task was cancelled (e.g. by a mode: restart automation) before
             # the transition timer was started, clean up the transitioning flag so
             # the light does not get stuck in a transitioning state indefinitely.
-            if (
-                set_transition_flag
-                and self._transitioning_individual
-                and not self._transition_listener
-            ):
-                self.async_transition_complete()
+            self._async_cleanup_transition_if_stuck(set_transition_flag)
 
     async def async_turn_off(self, *, transition: float | None = None) -> None:
         """Turn the entity off."""
@@ -647,12 +642,9 @@ class BaseClusterHandlerLight(BaseLight):
         finally:
             # If the task was cancelled before the transition timer was started,
             # clean up the transitioning flag so the light does not get stuck.
-            if (
+            self._async_cleanup_transition_if_stuck(
                 self._zha_config_enable_light_transitioning_flag
-                and self._transitioning_individual
-                and not self._transition_listener
-            ):
-                self.async_transition_complete()
+            )
 
     async def async_handle_color_commands(
         self,
@@ -742,6 +734,14 @@ class BaseClusterHandlerLight(BaseLight):
 
             with contextlib.suppress(ValueError):
                 self._tracked_handles.remove(self._transition_listener)
+
+    def _async_cleanup_transition_if_stuck(self, guarded: bool) -> None:
+        """Call async_transition_complete if the flag is set but no timer is running.
+
+        Used in finally blocks to handle task cancellation gracefully.
+        """
+        if guarded and self._transitioning_individual and not self._transition_listener:
+            self.async_transition_complete()
 
     def async_transition_complete(self, _=None) -> None:
         """Set _transitioning_individual to False and write HA state."""
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -311,7 +311,7 @@ class BaseClusterHandlerLight(BaseLight):
             light_options.enable_light_transitioning_flag
         )
 
-    async def async_turn_on(  # noqa: C901
+    async def async_turn_on(
         self,
         *,
         transition: float | None = None,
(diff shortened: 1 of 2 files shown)


zha/application/platforms/light/__init__.py
```diff
@@ -372,210 +372,224 @@ async def async_turn_on(
...
+                # if brightness is not 0.
+                result = await self._on_off_cluster_handler.on()
+                t_log["on_off"] = result
+                if result[1] is not Status.SUCCESS:
+                    # 'On' call failed, but as brightness may still transition
+                    # (for FORCE_ON lights), we start the timer to unset the flag after
+                    # the transition_time if necessary.
+                    self.async_transition_start_timer(transition_time)
+                    self.debug("turned on: %s", t_log)
+                    return
```
TheJulianJES: The missing coverage is because ~~all of these lines were indented~~. They don't have any coverage.

I'm also not sure if they're really needed, since we should just raise anyway if the device doesn't respond, right? I forgot about what the zigpy `request` can return though. I think we also raise for unsupported commands..? If so, we should be able to remove all of these in a future PR.

TheJulianJES:
Diff for `turn_on` is somewhat reduced with https://github.com/zigpy/zha/pull/672/changes/34e38c4ad6665e8c3b9f094b9976a5d112340d57 by moving the existing implementation to another method and just keeping the `try / finally` in `async_turn_on`.

For `async_turn_off`, the existing method is kept, just adding `try / finally` at start/end, indenting everything in between.

This is somewhat of a crucial fix, so it might be nice to get this in.

zha/application/platforms/light/__init__.py
```diff
@@ -527,6 +563,11 @@ async def async_turn_on(
             )
             t_log["move_to_level_if_color"] = result
             if result[1] is not Status.SUCCESS:
```
puddly, line 565: I'm thinking of having zigpy soon handle all ZCL default responses directly, throwing an error for any non-`SUCCESS` codes. Is it possible for a light to return a non-successful result?

puddly (approved):
Looks good to me. Thanks!

zha/application/platforms/light/__init__.py
TheJulianJES: Yeah, that's the same question I asked in the [comment above](https://github.com/zigpy/zha/pull/672#discussion_r2850122690). I hoped you would have an answer 😄 

I don't think we ever hit these paths..? Before I first touched the light class like 6(?) years ago, this code already existed. Maybe we can just remove soon?

puddly: Ah, the existing comments were collapsed in the  `/changes` view...

Yeah, let's remove them and see what breaks.

puddly: There are quite a few of these: https://github.com/search?q=repo%3Azigpy%2Fzha%20%22!%3D%20Status.SUCCESS%22%20OR%20repo%3Azigpy%2Fzha%20%22is%20not%20Status.SUCCESS%22&type=code

TheJulianJES: > Ah, the existing comments were collapsed in the `/changes` view...

I can't tell you how much I hate that new feature 😅 

I'll put up another PR to remove them. We don't have to merge that for this release yet.

TheJulianJES: > There are quite a few of these

Oh, yeah... The `isinstance(result, Exception)` check can most definitely go as well. Returning an exception instead of raising seems interesting 😄

puddly: Definitely. We got rid of most of them but the root cause seems to be `asyncio.gather(..., return_exceptions=True)`: https://github.com/search?q=repo%3Azigpy%2Fzha%20return_exceptions%3DTrue&type=code

TheJulianJES: Ah, ok. That's at least always in places where it kind of makes sense.

PR #675, opened 2026-02-24
Make tests run on python 3.14

dmulcahey:
Followed what was done here: https://github.com/home-assistant/core/pull/146699 Tests now run clean on python 3.14


TheJulianJES:
Interesting this only affected Python 3.14.3 as well then. The tests ran fine on 3.14.2 without the changes.
When we talked about this some time ago, I was referencing this issue that Core still has with 3.14.3:
- https://github.com/home-assistant/core/pull/162263#issuecomment-3854003680
- https://github.com/python/cpython/issues/144506

PR #677, opened 2026-02-25
Remove request `isinstance` `Exception` checks from switch

TheJulianJES:
This removes two dead `isinstance` check in the switch platform. Zigpy cannot return an exception as the result here. It'll always raise instead.

Another PR will remove all `Status.SUCCESS` checks for the result. We'll have to be slightly more careful with that change because ZCL errors like `UNSUPP_CLUSTER_COMMAND` don't raise(?) But they should also never occur, really.

Related discussion in:
- https://github.com/zigpy/zha/pull/672#discussion_r2850410343


TheJulianJES:
Unlike https://github.com/zigpy/zha/pull/678, it is safe to just remove these two `isinstance` checks.

PR #680, opened 2026-02-25
Use attribute names for `send_attributes_report` in tests

TheJulianJES:
## Proposed change

In preparation for bumping zigpy to 1.0.0 with https://github.com/zigpy/zha/pull/679, we need to adapt our tests to send attribute reports with the correct manufacturer code due to https://github.com/zigpy/zigpy/pull/1775. With https://github.com/zigpy/zha/pull/637, some tests were already updated to provide the attribute name (instead of ID) to `send_attributes_report`. This updates the tests that also require this for the new zigpy version.

## Additional information

We may want to revisit `send_attributes_report`, since using `None` manufacturer code when providing an ID is not fully correct: https://github.com/zigpy/zha/blob/c5a3171597848515b5a90bb6ed13b12972f75415/tests/common.py#L141-L143

`find_attribute` could be used and determine the `ZCLAttributeDef` if there's only matching definition by ID.
We could also simply prevent allowing providing attribute IDs to this. It's better if all tests provided attribute names or `ZCLAttributeDef` references directly.
... or we just leave it like this and accept that it should basically never be used.

### DIFF
--- a/tests/test_fan.py
+++ b/tests/test_fan.py
@@ -556,11 +556,11 @@ async def test_fan_ikea(
     assert entity.state["is_on"] is False
 
     # turn on at fan
-    await send_attributes_report(zha_gateway, cluster, {6: 1})
+    await send_attributes_report(zha_gateway, cluster, {"fan_mode": 1})
     assert entity.state["is_on"] is True
 
     # turn off at fan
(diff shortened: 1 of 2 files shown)


puddly (approved):
Thanks!

PR #682, opened 2026-02-27
Disable device availability checker during diagnostics regeneration

TheJulianJES:
## Proposed change

This disable device availability checker during diagnostics regeneration.
At the moment, there's a rare issue where you can manage to regenerate diagnostics with `available` being set to `false` everywhere, as the checker starts after 30 to 45 seconds of running diagnostics regeneration.

It's possible that the implementation can be cleaned up. Just wanted to put it into a PR for now.

### DIFF
--- a/tools/regenerate_diagnostics.py
+++ b/tools/regenerate_diagnostics.py
@@ -38,6 +38,12 @@ async def main(files: list[str] | None = None) -> None:
         paths = [pathlib.Path(f) for f in files]
 
     async with create_zha_gateway() as zha_gateway:
+        # Regeneration should be deterministic: disable background tasks that can
+        # asynchronously mutate device availability while fixtures are being written.
+        zha_gateway.global_updater.stop()
+        zha_gateway._device_availability_checker.stop()  # noqa: SLF001
+        zha_gateway.config.allow_polling = False
+
         for device_json in paths:
             _LOGGER.info("Migrating %s", device_json)
             zigpy_device = await zigpy_device_from_json(


tools/regenerate_diagnostics.py
```diff
@@ -38,6 +38,12 @@ async def main(files: list[str] | None = None) -> None:
         paths = [pathlib.Path(f) for f in files]
 
     async with create_zha_gateway() as zha_gateway:
+        # Regeneration should be deterministic: disable background tasks that can
+        # asynchronously mutate device availability while fixtures are being written.
+        zha_gateway.global_updater.stop()
+        zha_gateway._device_availability_checker.stop()  # noqa: SLF001
+        zha_gateway.config.allow_polling = False
```
TheJulianJES, line 45: `zha_gateway.config.allow_polling` will be set to `True` by `fetch_updated_state`: https://github.com/zigpy/zha/blob/9d03d639ee93f61135c2c02445cae23f8abd6723/zha/application/gateway.py#L389

With https://github.com/zigpy/zha/pull/654/, we make sure to run that before regenerating diagnostics but that also means it'll re-enable `allow_polling` still. So that's why we also stop the `global_updater` and `_device_availability_checker` to fix the issue.

Maybe we should just patch out `fetch_updated_state` entirely for diagnostics. EDIT: Eh, also not possible in a nice way.

TheJulianJES: If we'd set `allow_polling = False` after this line, it should work without needing to stop the `global_updater ` and `_device_availability_checker`:
https://github.com/zigpy/zha/blob/9d03d639ee93f61135c2c02445cae23f8abd6723/tools/import_diagnostics.py#L327

Hmm...

PR #684, opened 2026-02-27
Use `feature_priority` for update entities

TheJulianJES:
## Proposed change

This introduces `feature_priority` for update entities.

Currently, there's an issue where both the update entity for the client OTA cluster and the one for the server OTA cluster are discovered. If there's both an `in` and  `out` OTA cluster present on a device (endpoint), the `unique_id` will conflict, with no warnings logged. The client OTA entity class indirectly takes precedence due to definition order.

This PR changes that to have only one entity be truly discovered, making the priority more explicit: client, then server.
For all affected devices in our diagnostics, see: https://github.com/zigpy/zha/pull/683

## Additional information

Note: We can also prefer the server OTA entity if that's better, without causing breaking changes in HA. The  `unique_id` stays the same after all. But I don't know which one actually works on these devices. Client is generally the way to do it.

### DIFF
--- a/zha/application/platforms/__init__.py
+++ b/zha/application/platforms/__init__.py
@@ -72,6 +72,9 @@ class PlatformFeatureGroup(StrEnum):
     # Model-specific overrides for local temperature calibration
     LOCAL_TEMPERATURE_CALIBRATION = "local_temperature_calibration"
 
+    # Prefer OTA client update entities over OTA server update entities
+    OTA_UPDATE = "ota_update"
+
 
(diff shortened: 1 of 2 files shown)


puddly (approved):
Looks good to me. I wonder: does the zigpy side even work with a server-side OTA cluster? I think both were kept because discovery in the past didn't really differentiate between the two.

TheJulianJES:
Hmm, no idea. We reintroduced that swapped direction matching, so I guess it should..?
Eh, but not sure if it works from zigpy -> device?

PR #688, opened 2026-02-27
Fix thermostat rounding and Sinope polling idempotency

dmulcahey:
This pull request addresses two main issues with Sinope thermostat entities: preventing duplicate time update tasks when enabling the entity, and improving temperature setpoint precision. It also includes related test coverage and minor code improvements.

**Sinope time update task management:**

* Added a check in `start_polling()` to prevent duplicate time update tasks from being created when `enable()` is called multiple times. This ensures only one background task is active per entity.
* Updated `disable()` to safely remove the time update task from `_tracked_tasks` using `contextlib.suppress` to avoid errors if the task is already removed.
* Introduced a test (`test_sinope_enable_is_idempotent_for_time_update_task`) to verify that enabling a Sinope thermostat multiple times does not create duplicate time update tasks.

**Temperature setpoint precision:**

* Changed all setpoint conversions from `int(temperature * ZCL_TEMP)` to `round(temperature * ZCL_TEMP)` in `async_set_temperature()` to preserve decimal precision and avoid subtle setpoint drift.
* Added a test (`test_set_temperature_heat_decimal_precision`) to confirm that decimal precision is maintained in setpoint conversions for heating.

**Minor improvements:**

* Imported `contextlib` to support safe task removal.

## Failing tests addressed
- `tests/test_climate.py::test_set_temperature_heat_decimal_precision`
- `tests/test_climate.py::test_sinope_enable_is_idempotent_for_time_update_task`

## Verification
- `pytest` (branch-local targeted run): both tests pass.

### DIFF
--- a/tests/test_climate.py
+++ b/tests/test_climate.py
@@ -430,6 +430,48 @@ async def test_sinope_time(
     entity._async_update_time.reset_mock()
 
 
+async def test_sinope_enable_is_idempotent_for_time_update_task(
+    zha_gateway: Gateway,
+) -> None:
+    """Test Sinope enable does not duplicate time updater task."""
+    dev_climate_sinope = await device_climate_sinope(zha_gateway)
+    entity: ThermostatEntity = get_entity(
+        dev_climate_sinope, platform=Platform.CLIMATE, entity_type=ThermostatEntity
+    )
+
+    time_updater_task_name = f"sinope_time_updater_{entity.unique_id}"
+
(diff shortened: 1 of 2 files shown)


dmulcahey:
Closing as superseded by the refreshed #830. That PR retains the thermostat centidegree-rounding fix and Sinope polling lifecycle fix on current `dev`, with cancellation-aware updater ownership tests. Further review should continue there rather than on this older duplicate.

PR #694, opened 2026-02-27
Remove stale siren off-listener handles on retrigger

dmulcahey:
This pull request addresses an issue with the siren entity's timed activation logic, specifically preventing the accumulation of stale canceled handles when the siren is repeatedly turned on with a timer. The changes ensure that canceled listeners are properly removed from the entity's internal tracking, preventing unbounded growth and potential resource leaks. A new test is also added to validate this behavior.

**Bug fix: Proper cleanup of stale handles on repeated timed activation**

* Updated the `async_turn_on` method in `zha/application/platforms/siren.py` to remove canceled listeners from the `_tracked_handles` set, ensuring that only active handles are tracked and preventing memory/resource leaks.

**Testing improvements**

* Added a new test `test_siren_repeated_timed_turn_on_leaks_stale_tracked_handles` in `tests/test_siren.py` to verify that repeated timed activations do not retain stale or canceled handles, and that the cleanup logic works as intended.## Summary
Removes stale canceled timer handles when re-triggering timed siren activation.

## Failing tests addressed
- `tests/test_siren.py::test_siren_repeated_timed_turn_on_leaks_stale_tracked_handles`

## Verification
- `pytest` (branch-local targeted run): listed test passes.

### DIFF
--- a/tests/test_siren.py
+++ b/tests/test_siren.py
@@ -151,3 +151,46 @@ async def test_siren_timed_off(zha_gateway: Gateway) -> None:
 
     # test that the state has changed to off from the timer
     assert entity.state["state"] is False
+
+
+async def test_siren_repeated_timed_turn_on_leaks_stale_tracked_handles(
+    zha_gateway: Gateway,
+) -> None:
+    """Test repeated timed turn_on does not retain stale tracked handles."""
+    zha_device, _ = await siren_mock(zha_gateway)
+    entity = get_entity(zha_device, platform=Platform.SIREN)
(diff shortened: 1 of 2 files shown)


dmulcahey:
Closing as already addressed by the current siren implementation introduced in #665. `BaseSiren._cancel_off_listener` cancels the timer and removes the original handle from the removal callbacks, and the turn-on paths use that helper. The stale-handle accumulation targeted here is therefore already handled on current `dev`.

PR #695, opened 2026-02-27
Bump pytest-asyncio to 1.4.0

TheJulianJES:
This bumps the minimum required `pytest-asyncio` version to 1.4.0, though as low as 1.0.0 is probably fine as well.
It's just a test dependency though, so pinned it to what I tested with.

Multiple changes were made to make it work, since the `event_loop` fixture was finally dropped (long deprecated before).

### DIFF
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -6,7 +6,7 @@ pre-commit
 pytest-cov
 pytest-sugar
 pytest-timeout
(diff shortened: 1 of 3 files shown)


TheJulianJES:
@zigpy-review-bot Should we just require 1.4.0 (latest) at this point? If so, push commit for that.

PR #696, opened 2026-02-27
Suppress expected warnings in tests

TheJulianJES:
This filters expected warnings triggered by various tests.
Alternatively, we can also add this to our `pyproject.toml`. Adding it to the individual tests or modules doesn't really work here, except if we want to add it to basically all.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -41,6 +41,11 @@ COUNTER_NAMES = ["counter_1", "counter_2", "counter_3"]
 INSTANCES: list[Gateway] = []
 _LOGGER = logging.getLogger(__name__)
 
+EXPECTED_WARNING_FILTERS = (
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -41,6 +41,11 @@
 INSTANCES: list[Gateway] = []
 _LOGGER = logging.getLogger(__name__)
 
+EXPECTED_WARNING_FILTERS = (
+    r"ignore:Attribute '.*' has `is_manufacturer_specific` without an explicit `manufacturer_code`\. Please set `manufacturer_code=0x[0-9A-Fa-f]+`\.:DeprecationWarning",
+    r"ignore:Unique IDs are unique only with platform prefix.*:UserWarning",
```
TheJulianJES, line 46: At the moment, unique IDs only being unique within a platform is expected and this is logged quite a bit.
I can remove this, of course, but I think it might make sense to keep until we've migrated away from intentionally using that?

PR #702, opened 2026-03-05
Show update release URL for installed version

TheJulianJES:
## Proposed change

This changes ZHA update entities to show the `release_url` of the latest firmware (provided as `.downgrades[0]`).

## Additional information

For zigpy-ota provided updates, the `release_url` is generally a link to the pull request where that update was submitted. The pull request also shows more context for the update, potentially with additional release notes in the comments. This also allows us to have an easy place to collect feedback for these updates and for HA users to find that, even after they already installed the update.

We can also consider showing the latest (markdown) release notes if that update is already installed. If so, should only the latest version's release notes be shown, the last 5, or all of them?

Related:
- https://github.com/zigpy/zha/pull/625

### Newer version installed than latest in OTA repo?

Right now, we'll show the release URL for the latest version in zigpy-ota. If a user has newer firmware installed than in the zigpy-ota repo, should we not show that link then?
This sometimes happens when manufacturers install new firmware in the factory but consider it to be unimportant to upgrade existing devices with.

### Apps

ESPHome / HA apps do not use HA's `release_url` but instead provide the "release announcement link" in the release notes themselves. This may also be useful for us, since we combine multiple upgrade release notes with the PR linked above, but only ever show the link for the latest release.

More info dialog for ESPHome update entity showing release announcement link in the release notes itself, instead of using HA's dedicated release link position at the top

It's also a little easier to spot there IMO, as it's [currently a bit hidden](https://github.com/user-attachments/assets/f77632e6-610d-4136-a935-fa729b573831) at the top. The general update note is only shown below, which Matter might do the other way around even(?)

### DIFF
--- a/tests/test_update.py
+++ b/tests/test_update.py
@@ -577,13 +577,15 @@ async def test_firmware_update_no_image(zha_gateway: Gateway) -> None:
 async def test_firmware_update_latest_version_even_if_downgrade(
     zha_gateway: Gateway,
 ) -> None:
-    """Test ZHA update platform - `latest_version` always reflects the latest."""
+    """Test ZHA update platform - latest downgrade sets version and release URL."""
     zigpy_device = zigpy_device_mock(zha_gateway)
     zha_device, ota_cluster, fw_image, installed_fw_version = await setup_test_data(
         zha_gateway, zigpy_device
     )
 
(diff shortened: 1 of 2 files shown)


TheJulianJES:
IMO, we should just show all old firmware versions with full release notes. That's what all add-ons/apps apparently do.
For now, let's merge this.

PR #703, opened 2026-03-09
Prefer `device_type` over cluster/attribute values for Light

puddly:
This is a test PR to potentially address https://github.com/home-assistant/core/issues/164598#issuecomment-4024488523.

I've added the affected devices to our test database and have regenerated diagnostics, showing what devices would be affected if we use the device type more aggressively. I haven't tested the changes yet but this affects very few devices.

### DIFF
--- a/tests/data/devices/sengled-z01-a19nae26.json
+++ b/tests/data/devices/sengled-z01-a19nae26.json
@@ -505,7 +505,7 @@
           ],
           "supported_features": 40,
           "min_mireds": 153,
-          "max_mireds": 370
+          "max_mireds": 500
         },
         "state": {
           "class_name": "MinTransitionLight",
@@ -513,16 +513,15 @@
           "on": false,
           "brightness": 252,
           "xy_color": null,
-          "color_temp": 370,
+          "color_temp": null,
           "effect_list": [
             "off"
           ],
           "effect": "off",
           "supported_features": 40,
-          "color_mode": "color_temp",
+          "color_mode": "brightness",
           "supported_color_modes": [
             "brightness",
-            "color_temp",
             "onoff"
           ],
           "off_with_transition": false,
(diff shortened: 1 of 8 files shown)


puddly:
Interestingly, there is _not a single_ device in our testing database that uses a color device type (or a color temperature device type) without also having a color cluster! I would have thought this to not be the case but it looks like manufacturers use the device type pretty strictly compared to clusters themselves.

puddly:
I've added the `GLEDOPTO GL-SD-301P` to diagnostics as well, from https://github.com/home-assistant/core/issues/164598#issuecomment-4025566371. It follows the same pattern: endpoint with a `DIMMABLE_LIGHT` device type and a bogus `Color` cluster with no capabilities. We erroneously identify it as a color temperature light. It somehow manages to send attribute reports or something to make ZHA change its color mode.

tests/data/devices/sengled-z01-a19nae26.json
TheJulianJES, line 1: Checking Z2M and official instructions, the Sengled Z01-A19NAE26 does support color temperature. This patch now changes that to "brightness only". I imagine at least other Sengled lights will be affected as well.

TheJulianJES (commented):
This is definitely useful, but I fear we'll break some lights. For a patch release, that's not great.

The main [HA issue](https://github.com/home-assistant/core/issues/164598) is that the light gets into a state where `color_mode` is set to `XY` (or `HS`?), even though that is not part of the supported color modes. As far as I can see, this can only ever happen during polling.

The initialization/[capabilities set `color_mode`](https://github.com/zigpy/zha/blob/c51efa5aaedbcbe270e4bb35f69667e5b7015030/zha/application/platforms/light/__init__.py#L928-L930) based on the supported ones, so that should be fine. For an alternative approach that just fixes polling for a patch release, see:
- https://github.com/zigpy/zha/pull/707

PR #706, opened 2026-03-11
Add missing Inovelli VZM31-SN attributes to `ZCL_INIT_ATTRS`

wbyoung:
These are already [defined in the device handlers](https://github.com/zigpy/zha-device-handlers/blob/20a07776b156ad80d0bf2f09f983db1703fbe403/zhaquirks/inovelli/__init__.py#L663-L672).


TheJulianJES (approved):
Thanks!

Issue #226, opened 2024-10-10
Fan control interface does not display options correctly


Masy:
I'm currently developing a stick that can control midea AC units, and i stumbled upon a weird behavior when exposing a fan control cluster:

6.4.2.2.2 of the zigbee cluster specification states that the fan mode sequence attribute should be set to 0x02 if the fan supports low/med/high/auto modes. 
![grafik](https://github.com/user-attachments/assets/63d037d6-6c9d-4e62-bebe-6e5cf1374f04)

This does seem to be picked up correctly by home assistant / ZHA:
![grafik](https://github.com/user-attachments/assets/57cd5316-817f-45af-930c-758762cdb4d7)

However, in the interface, the only options are 'on' and 'auto', which corresponds to 0x04:
![grafik](https://github.com/user-attachments/assets/9861dcee-0847-4aa6-a533-80dad2f075bb)

Masy:
I noticed the thermostat isn't displayed correctly as well. I set the ControlSequenceOfOperation to 0x04, which should display all possible modes of SystemMode.
![grafik](https://github.com/user-attachments/assets/1c0e70d6-b9a4-4457-98bb-cc06c738ab9b)
![grafik](https://github.com/user-attachments/assets/5a27298b-8322-4014-b9b2-b3f6a9259f1e)

The only displayed modes for me are off, cool, heat and auto. If i manually set the SystemMode to dry when initializing the zigbee device on the ESP32C6 it is displayed in HA, but i can't set it via the interface.
![grafik](https://github.com/user-attachments/assets/27e02cca-c072-468c-9b1a-9dd970f59874)

puddly:
HVAC operations are sort of split up between climate modes, default climate presets, and custom climate presets. We map them to one of the supported Home Assistant climate modes here and create presets on a thermostat-by-thermostat basis: https://github.com/zigpy/zha/blob/4a4abd2e01f4e1a144550171d91e736a932ae37d/zha/application/platforms/climate/const.py#L150-L169.

I think for this attribute to work, the various modes from the ZCL would need to be turned into custom presets.

Does your device actually support all of the ZCL modes? If not, which ones does it support?

Masy:
the only thing im unsure about is if the eco mode corresponds to precooling, but i guess it supports all modes of the zigbee standard mentioned in table 6-16

Masy:
[...]

PR #708, opened 2026-03-12
Fix thermostat fan modes to respect fanModeSequence

vmvarga:
## Summary

- Fix the `Thermostat` entity to read the `fan_mode_sequence` attribute from the ZCL Fan Control cluster (0x0202) instead of hardcoding fan modes to `[auto, on]`. Devices now correctly expose Low, Medium, High, and Auto fan speeds based on the ZCL spec (Table 6-20).
- Fix the `fan_mode` getter to read the actual `fan_mode` attribute from the Fan Control cluster handler instead of guessing from the thermostat's `running_state` bitmap.
- Fix `async_set_fan_mode` to map all supported fan mode strings (low, medium, high, on, auto) to their corresponding `FanMode` ZCL enum values.
- Expose `HVACMode.FAN_ONLY` as an available HVAC mode when a Fan Control cluster is present on the endpoint, since `SEQ_OF_OPERATION` (derived from `controlSequenceOfOperation`) never includes it per ZCL spec.

## Problem

The `Thermostat` entity hardcodes fan modes to `[auto, on]` and completely ignores the `fan_mode_sequence` attribute (attribute 0x0001) from the Fan Control cluster. Per ZCL 8 (Table 6-20), `FanModeSequenceType` defines which fan modes a device supports:

| Value | Available Modes |
|-------|----------------|
| 0x00 | Low, Medium, High |
| 0x01 | Low, High |
| 0x02 | Low, Medium, High, Auto |
| 0x03 | Low, High, Auto |
| 0x04 | On, Auto |

Devices reporting e.g. `fan_mode_sequence = 0x02` (Low/Medium/High/Auto) were stuck with only Auto/On in the UI.

Additionally, the thermostat never exposes `HVACMode.FAN_ONLY` even when the device supports `SystemMode.Fan_only` (0x07). The mappings `HVAC_MODE_2_SYSTEM` and `SYSTEM_MODE_2_HVAC` already handle `FAN_ONLY <-> Fan_only`, but `hvac_modes` derives its list solely from `SEQ_OF_OPERATION` which never includes it, since `controlSequenceOfOperation` only covers cooling/heating per the ZCL spec.

### `zha/application/platforms/climate/const.py`
- Import `FanMode` from zigpy
- Add `SEQ_FAN_MODES` dict mapping `fan_mode_sequence` values (0x00–0x04) to fan mode string lists
- Add `FAN_MODE_TO_ZCL` dict mapping fan mode strings to `FanMode` enum values
- Add `ZCL_TO_FAN_MODE` reverse mapping dict

### `zha/application/platforms/climate/__init__.py`
- `fan_modes`: Read `fan_mode_sequence` from the fan cluster handler, look up modes via `SEQ_FAN_MODES`, fall back to `[on, auto]` for unknown sequences
- `fan_mode`: Read actual `fan_mode` attribute from the fan cluster handler, map back via `ZCL_TO_FAN_MODE`, fall back to `running_state` heuristic when unavailable
- `async_set_fan_mode`: Use `FAN_MODE_TO_ZCL` mapping instead of hardcoded `On`/`Auto` branch
- `hvac_modes`: Append `HVACMode.FAN_ONLY` when a fan cluster handler is present

## Backwards compatibility

- Devices with `fan_mode_sequence = 0x04` still get `[on, auto]` (unchanged)
- Devices without a Fan Control cluster are completely unaffected
- Unknown `fan_mode_sequence` values fall back to `[on, auto]`
- No changes to `HVAC_MODE_2_SYSTEM`, `SYSTEM_MODE_2_HVAC`, or `async_set_hvac_mode`

## Test plan

- [x] Device with `fan_mode_sequence = 0x02` exposes: Low, Medium, High, Auto
- [x] Device with `fan_mode_sequence = 0x04` (or unknown) exposes: On, Auto (backwards compatible)
- [x] Setting fan mode to "low" sends `FanMode.Low` (0x01) to the device
- [x] Setting fan mode to "high" sends `FanMode.High` (0x03) to the device
- [x] Current fan mode reflects the actual `fan_mode` attribute from the Fan Control cluster
- [x] Thermostat with a Fan Control cluster exposes "Fan only" as an HVAC mode
- [x] Selecting "Fan only" writes `SystemMode.Fan_only` (0x07) to the thermostat
- [x] Thermostat without a Fan Control cluster is unaffected

## Affected devices

Any thermostat with a Fan Control cluster (0x0202) that reports `fan_mode_sequence != 0x04`, such as Tuya HVAC thermostats with cooling/heating/fan modes (e.g. `_TZE204_mpbki2zm`).


zha/application/platforms/climate/const.py
```diff
@@ -141,6 +146,24 @@ class HVACAction(StrEnum):
     PREHEATING = "preheating"
 
 
+SEQ_FAN_MODES: dict[int, list[str]] = {
+    0x00: [FAN_LOW, FAN_MEDIUM, FAN_HIGH],
```
puddly: This should use the zigpy constant: `FanModeSequence.Low_Med_High`

puddly:
To regenerate device diagnostics files (to see what this change would affect in real devices), run the following:

```bash
python -m tools.regenerate_diagnostics
```

It looks like four devices (in the testing DB) change their exposed fan modes.

vmvarga:
Should resolve #226

puddly:
I noticed that one of the test devices was my thermostat. I can confirm this shows up properly:

image

But the fan mode doesn't really do anything for my specific thermostat even when switching the fan mode from "auto" to "on", it doesn't actually support fan mode independent from heat and cool. This is probably just a device-specific bug.

@dmulcahey @TheJulianJES Do you have a device to test this change with?

TheJulianJES:
I'll have a closer look later, but I think the issue is still present in Matter: https://github.com/home-assistant/core/blob/9ddefaaacd1df9da4f5a4f673eea2348b524d0ac/homeassistant/components/matter/climate.py#L121-L159 (assuming it's even the same issue here, which it might not be...)

Arol450:
Hi, I wanted to add a real-world data point in support of this PR.

I've been running a custom ESP32-C6 Zigbee bridge that exposes a Midea split AC unit to Home Assistant via ZHA using standard clusters (Thermostat 0x0201, Fan Control 0x0202, Temperature Measurement 0x0402). The device reports `FanModeSequence.Low_Med_High_Auto (0x02)` and supports `SystemMode.Fan_only (0x07)` and `SystemMode.Dry (0x08)`.

With the current ZHA code:
- Fan modes are hardcoded to `[auto, on]` — the actual sequence reported by the device is ignored
- `fan_only` is not available as an HVAC mode despite the Fan Control cluster being present

Both of these issues are exactly what this PR addresses. I'd be happy to test a build with this change applied against my device if that would help move things forward.

Thanks for working on this.

puddly:
The specific concern I brought above is that "fan only" support isn't reliably signaled by devices. My thermostat (and I imagine many of the others in our testing database) show up as supporting "fan only" but do not actually support it. There is, as far as I can tell, no way to tell from the ZCL alone that a device supports "fan only". We may need an opt-in mechanism via quirks, unfortunately.

vmvarga:
Added opt-in via quirks

vmvarga:
@puddly Can you take a look if latest commit is in line with what you suggested?

grizeldi:
Is there anything I can help with to get this merged? Or is it purely waiting for someone to merge it?

I'm also developing a esp32h2 based air conditioning controller and am hitting the exact same situation Arol described above when running the stable ZHA.

vmvarga:
Fixed merge conflicts

puddly:
I think as an opt-in setting via quirk feature ID this is totally fine, since it's disabled by default.

@TheJulianJES thoughts?

zha/application/platforms/climate/__init__.py
```diff
@@ -670,7 +680,14 @@ def hvac_mode(self) -> HVACMode | None:
     @property
     def hvac_modes(self) -> list[HVACMode]:
         """Return the list of available HVAC operation modes."""
-        return SEQ_OF_OPERATION.get(self._ctrl_sequence_of_oper, [HVACMode.OFF])
+        modes = SEQ_OF_OPERATION.get(self._ctrl_sequence_of_oper, [HVACMode.OFF])
+        if (
+            self._fan_cluster is not None
+            and THERMOSTAT_FAN_ONLY_HVAC in self._device.exposes_features
```
TheJulianJES, line 711: I'm not saying we should, but if we wanted to, we could also have another entity class that overrides the fist one (so likely using a feature group) and then matches on the exposed feature. There, we'd only add `FAN_ONLY` to the list of possible modes. It would keep the base ZCL class clean of the exposed feature, but I think either is fine.

zha/application/platforms/climate/const.py
```diff
@@ -22,6 +28,12 @@
 ATTR_TARGET_TEMP_LOW: Final[str] = "target_temp_low"
 ATTR_TEMPERATURE: Final[str] = "temperature"
 
+# Opt-in quirk feature id: when present in ``device.exposes_features``, the
+# thermostat entity will expose ``HVACMode.FAN_ONLY``. The presence of the Fan
+# cluster (0x0202) alone is not sufficient, because many devices advertise the
+# cluster without actually implementing ``SystemMode.Fan_only`` (0x07).
+THERMOSTAT_FAN_ONLY_HVAC: Final[str] = "thermostat_fan_only_hvac"
```
TheJulianJES: This should really live in zha-quirks at the moment: https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/quirk_ids.py

EDIT: Not anymore after the ZHA/quirks refactor.

TheJulianJES (commented):
I think using an "exposed feature" via quirks is indeed the best we can do here for the `FAN_ONLY` mode.

vmvarga:
THERMOSTAT_FAN_ONLY_HVAC was moved to zhaquirks

vmvarga:
Updated according to #762

vmvarga:
@puddly @TheJulianJES Is there anything else that stops us from merging?

jdeksup:
Hi,
I tested this implementation in my home assistant by doing tweak installation.
Before I had only "Auto | On" options. Now it works like a charm (in french), on "Adapter Zigbee FUJITSU par ATLANTIC GROUP":
image
Any way to have this merged ?

zha/application/platforms/climate/const.py
```diff
@@ -141,6 +147,33 @@ class HVACAction(StrEnum):
     PREHEATING = "preheating"
 
 
+SEQ_FAN_MODES: dict[int, list[str]] = {
+    FanModeSequence.Low_Med_High: [FAN_LOW, FAN_MEDIUM, FAN_HIGH],
+    FanModeSequence.Low_High: [FAN_LOW, FAN_HIGH],
+    FanModeSequence.Low_Med_High_Auto: [FAN_LOW, FAN_MEDIUM, FAN_HIGH, FAN_AUTO],
+    FanModeSequence.Low_High_Auto: [FAN_LOW, FAN_HIGH, FAN_AUTO],
+    FanModeSequence.On_Auto: [FAN_ON, FAN_AUTO],
```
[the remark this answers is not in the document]
TheJulianJES: (Note: You can just run `python -m tools.regenerate_diagnostics` to regenerate diagnostics and see the difference there.)

vmvarga:
@puddly @TheJulianJES Bot seems happy maybe this can be merged?

PR #711, opened 2026-03-17
Clean up default reporting config errors

puddly:
I ran into a few errors in our default reporting config:
1. BinaryInput, BinaryOutput, and BinaryValue should use `REPORT_CONFIG_IMMEDIATE`, not `REPORT_CONFIG_DEFAULT`.
2. The various `state` and `mode` Thermostat cluster attributes should use a discrete reporting config, since the "minimum change" will be ignored.


TheJulianJES (approved):
Looks good.

It will be fun to automatically re-set up reporting for all/affected existing devices, if we ever do that.. 😅

PR #712, opened 2026-03-19
Fix NaN values being recognized as non-values

TheJulianJES:
## AI summary
- Handle `NaN` float values in sensor `_is_non_value` by adding a `math.isnan()` check, since IEEE 754 `NaN != NaN` causes the existing `==` comparison against `data_type.non_value` to silently pass through
  - Note/TODO: Can we fix this in a better way?
- Sensors (e.g. CO concentration) now return `None` instead of propagating `NaN` to Home Assistant (which caused errors because of the state class)

## Test plan
- Added `test_ignore_nan_value` using the frient SCAZB-141 diagnostics file, verifying that a `NaN` attribute report on the carbon monoxide concentration cluster results in `None` state

### DIFF
--- a/zha/application/platforms/sensor/__init__.py
+++ b/zha/application/platforms/sensor/__init__.py
@@ -10,6 +10,7 @@ from datetime import date, datetime, timedelta
 import enum
 import functools
 import logging
+import math
 import numbers
 import typing
 from typing import TYPE_CHECKING, Any, cast
@@ -364,6 +365,9 @@ class Sensor(BaseSensor):
         self, value: int | float, *, attr_def: foundation.ZCLAttributeDef | None = None
     ) -> bool:
         """Ignore non-value numerical values."""
+        if isinstance(value, float) and math.isnan(value):
+            return True
+
         if attr_def is None:
             attr_def = self._attr_def
 
--- a/tests/data/devices/frient-a-s-scazb-141.json
+++ b/tests/data/devices/frient-a-s-scazb-141.json
@@ -1,0 +1,1229 @@
+{
+  "version": 1,
+  "ieee": "ab:cd:ef:12:18:f2:e7:e6",
+  "nwk": "0xEA73",
+  "manufacturer": "frient A/S",
+  "model": "SCAZB-141",
+  "friendly_manufacturer": "frient A/S",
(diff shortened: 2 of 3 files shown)


zha/application/platforms/sensor/__init__.py
```diff
@@ -364,6 +365,9 @@ def _is_non_value(
         self, value: int | float, *, attr_def: foundation.ZCLAttributeDef | None = None
     ) -> bool:
         """Ignore non-value numerical values."""
+        if isinstance(value, float) and math.isnan(value):
+            return True
+
```
TheJulianJES, line 370: An alternative approach – if we want to use the zigpy entries with NaN for non-values – would be this:
```diff
diff --git a/zha/application/platforms/sensor/__init__.py b/zha/application/platforms/sensor/__init__.py
index f7972c71..5750616e 100644
--- a/zha/application/platforms/sensor/__init__.py
+++ b/zha/application/platforms/sensor/__init__.py
@@ -365,9 +365,6 @@ class Sensor(BaseSensor):
         self, value: int | float, *, attr_def: foundation.ZCLAttributeDef | None = None
     ) -> bool:
         """Ignore non-value numerical values."""
-        if isinstance(value, float) and math.isnan(value):
-            return True
-
         if attr_def is None:
             attr_def = self._attr_def
 
@@ -375,7 +372,13 @@ class Sensor(BaseSensor):
             return False
 
         data_type = foundation.DataType.from_type_id(attr_def.zcl_type)
-        return value == data_type.non_value
+        non_value = data_type.non_value
+
+        # NaN != NaN in IEEE 754, so == fails for float non-values
+        if isinstance(non_value, float) and math.isnan(non_value):
+            return isinstance(value, float) and math.isnan(value)
+
+        return value == non_value
 
     def formatter(self, value: Any) -> date | datetime | int | float | str | None:
         """Numeric pass-through formatter."""
```

Thoughts? It does feel like we'd be overcomplicating it though...

puddly (approved):
Ah, `nan`! I completely forgot about it being incomparable to anything...

Zigpy does define that the not-a-value value for Double, Single, and Half is `nan` but it's just informative, I think for a library consuming this information (ZHA) we don't strictly have to abide by it usage-wise. I think this approach is good.

PR #727, opened 2026-03-25
Add OTA last query cmd to diagnostics

TheJulianJES:
## Proposed change

This adds the last cached OTA query command to JSON diagnostics, so we know the `image_type` of a device.


zha/zigbee/device.py
```diff
@@ -1620,6 +1620,19 @@ def get_diagnostics_json(self):
...
+                                "last_query_cmd": {
+                                    "manufacturer_code": cluster.last_query_cmd.manufacturer_code,
+                                    "image_type": cluster.last_query_cmd.image_type,
+                                    "current_file_version": cluster.last_query_cmd.current_file_version,
+                                    "hardware_version": cluster.last_query_cmd.hardware_version,
+                                }
+                            }
+                            if isinstance(cluster, Ota)
+                            and getattr(cluster, "last_query_cmd", None) is not None
+                            else {}
```
TheJulianJES, line 1634: The `getattr` guard is for someone messing up a (quirked) OTA cluster badly, but I think we can just remove it. If someone manages this, congrats.

tests/test_discover.py
```diff
@@ -799,6 +800,55 @@ async def test_get_diagnostics_json_repeated_calls(zha_gateway: Gateway) -> None
...
+        "tests/data/devices/ikea-of-sweden-tradfri-bulb-gu10-ws-400lm.json",
+    )
+
+    ota_cluster = zigpy_device.endpoints[1].out_clusters[Ota.cluster_id]
+    ota_cluster.last_query_cmd = QueryNextImageCommand(
+        field_control=0,
+        manufacturer_code=0x117C,
+        image_type=0x1234,
+        current_file_version=0x00AABBCC,
+    )
```
TheJulianJES, line 816: Side note: In the future, we may wanna consider if we can also restore from diagnostics files to our device object generated from it, so the `last_query_cmd` is available on it then.

zha/zigbee/device.py
TheJulianJES: Eh, actually. We need a `is not None` check anyway, so I guess we can just keep this. It's one line either way.

PR #737, opened 2026-04-04
Fix BEGA color temp channel unsupported check

TheJulianJES:
## Proposed change

This fixes an issue where the "Color temperature channel" select entity for BEGA lights was created, even if the light driver only supports one color temperature channel. On truly "dimmable only" lights with one driver channel only, one attribute still has the correct color temperature (e.g. 3000K), with the other one being set to `0xFFFF`.

To fix this, we just do not create the select entity if either driver channel's color temperature is `0xFFFF`.

## Additional information

Follow-up to:
- https://github.com/zigpy/zha/pull/721

To note again: These lights are **not** tunable color temperature lights. They can only be fixed to either 3000K or 4000K (depends on exact model). So, this cannot be represented in the Home Assistant light entity model and an extra select entity is needed to switch the driver channel. See the above PR for more notes on this.

### DIFF
--- a/tests/test_select.py
+++ b/tests/test_select.py
@@ -325,22 +325,31 @@ async def test_bega_color_temperature_channel_select(zha_gateway: Gateway) -> No
         )
 
 
+@pytest.mark.parametrize(
+    ("temp_1", "temp_2"),
+    [
+        (0xFFFF, 0xFFFF),
+        (0xFFFF, 3000),
+        (3000, 0xFFFF),
+    ],
+)
 async def test_bega_color_temperature_channel_select_unsupported(
     zha_gateway: Gateway,
+    temp_1: int,
+    temp_2: int,
 ) -> None:
-    """Test BEGA select entity is not created when color temps are 0xFFFF."""
+    """Test BEGA select entity is not created when a color temp is 0xFFFF."""
     zigpy_device = await zigpy_device_from_json(
         zha_gateway.application_controller,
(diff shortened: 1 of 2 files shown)


tests/test_select.py
```diff
@@ -325,22 +325,31 @@ async def test_bega_color_temperature_channel_select(zha_gateway: Gateway) -> No
...
 async def test_bega_color_temperature_channel_select_unsupported(
     zha_gateway: Gateway,
+    temp_1: int,
+    temp_2: int,
 ) -> None:
-    """Test BEGA select entity is not created when color temps are 0xFFFF."""
+    """Test BEGA select entity is not created when a color temp is 0xFFFF."""
     zigpy_device = await zigpy_device_from_json(
         zha_gateway.application_controller,
         "tests/data/devices/bega-gantenbrink-leuchten-kg-smart-dimmable-light.json",
```
TheJulianJES, line 344: Note: Almost all BEGA lights are named "Smart Dimmable Light". Some of those are only dimmable, others support two color temperature channels. For the latter ones only, the select entity should be created.

I've not added a second diagnostics file for now, since it's kind of hard to distinguish them. Other than the two color temperature channel attributes, only the `product_code` attribute would differ when read.

I think this test covers it well enough by just updating the attribute (cache), before joining the device.
Side note: It would be nice to have a better way to populate the attribute cache when read for `zigpy_device_from_json`.

Issue #756, opened 2026-05-02
[REQUEST] Add full hue_native_control option for native control of Philips Hue lights


Hedda:
For reference also see this feature request for ZHA Device Handlers which asking about custom device support for this cluster:

* https://github.com/zigpy/zha-device-handlers/issues/2517

Request add full support Philips Hue native control (a.k.a. "Hue Dynamic Scenes" or "Dynamic Lighting Scenes") with decode/encode of binary format used by Philips Hue Zigbee devices which can be used to command custom "Bifrost effects" (atomic Bifrost frames on cluster 0xFC03 to control different firmware-supported special animated scene effect like example "Candle", "Fireplace", and "Colorloop" animation) supported by a few Philips Hue lighting products and support Zigbee frames in a manufacturer specific cluster implementation format on cluster 0xFC03 to zha, zigpy, zha-quirks/zha-device-handlers, and the ZHA integration UI:

* https://github.com/chrivers/bifrost/blob/master/doc/hue-zigbee-format.md

* https://github.com/jtbandes/hue-zigbee-encoding

Philips Hue lights supports this via Zigbee frames in a manufacturer specific cluster implementation format, on cluster 0xFC03.

This is already supported by Zigbee2MQTT and is needed if want to achieve feature-parity with Z2M and zigbee-herdsman, and more importantlly really needed if want to encurage users of the official Philips Hue Bridge and app to move/migrate to ZHA:

* https://github.com/Koenkk/zigbee-herdsman-converters/pull/5192

  * https://github.com/Koenkk/zigbee2mqtt/issues/15699

     * https://community.home-assistant.io/t/a-simple-single-automation-candle-effect-for-hue-white-ambience-and-other-bulbs/235023/

Zigbee2MQTT supports these Hue Effects / animations (effect names):

* `candle` (Soft, flickering yellow/orange glow)
* `fireplace` (Dynamic flickering orange/red colors)
* `colorloop` (Cycles endlessly through the color spectrum)
* `blink`
* `breathe`
* `okay`
* `sunset`
* `sunrise`
* `sparkle`
* `opal`
* `glisten`
* `underwater`
* `cosmos`
* `sunbeam`
* `enchant`
* `none`, `finish_effect`, `stop_hue_effect` or `stop_effect` (Returns the light to its standard state)

These are supported by most Philips Hue series light products with color temperature, but see example these supported devices:
* https://www.zigbee2mqtt.io/devices/046677592530.html
[...]

PR #762, opened 2026-05-11
Migrate quirks out of zigpy and provide direct entity access

### LINKED ISSUE #93: Further shape Zigbee support by implementing missing features in ZHA while retaining choice

puddly:
This is a test PR for inverting the quirks dependency chain. Currently, zigpy is responsible for defining the quirks API, ZHA defines the base entity models and ZCL autodiscovery, and quirks use the zigpy APIs to provide ZHA context on how to create custom entities. This PR inverts it: all quirks constants and definitions will move out of zigpy and into ZHA, quirks will import ZHA itself, and ZHA will unpin quirks as an `==` dependency.

Currently, we have a few APIs for matching devices for the purposes of modifying their clusters at runtime:
1. `CustomDevice`: this matches the full device signature exactly and provides a way to replace the device object directly. Not recommended.
2. `QuirkBuilder`: matches by model and manufacturer, plus a firmware version filter. Also aggregates small operations for modifying clusters and endpoints but also provides a simplified API for exposing entities derived from ZCL attributes.
3. (proposed) `@register_device`: it sort of aggregates the two by having the ZHA `Device` subclass "be" the `QuirkBuilder`. We'd need an API to also allow replacing the zigpy `Device` object as well. This has been yanked out of zigpy by creating a "device resolver" callable that ZHA can provide. It replaces `zigpy.quirks.get_device` by allowing ZHA to itself provide a replacement device, avoiding the zigpy quirks registry entirely.

As a test implementation, I've added support for the manufacturer-specific Hue lighting frame format in a fork of quirks. It look something like this (100% slop that manages to run):

```python
"""Philips Hue native control via cluster 0xFC03 — v3 device-level quirk."""

from __future__ import annotations

import asyncio
import logging
from typing import Any, Final

from zigpy.quirks import CustomCluster
import zigpy.types as t
from zigpy.zcl import ClusterType
from zigpy.zcl.clusters.general import LevelControl, OnOff
from zigpy.zcl.clusters.lighting import Color
from zigpy.zcl.foundation import BaseCommandDefs, ZCLCommandDef

from zha.application import Platform
from zha.application.platforms.light import Light
from zha.zigbee.device import Device, DeviceMatch, Replace, register_device

class HueNativeLight(Light):
    """Light driven by atomic Bifrost frames on cluster 0xFC03."""

    def __init__(self, *args: Any, **kwargs: Any) -> None:
        super().__init__(*args, **kwargs)
        self._hue_cluster_handler = self.cluster_handlers[HueLightCluster.ep_attribute]

    @property
    def effect_list(self) -> list[str] | None:
        return list(EFFECT_NAME_TO_HUE)

    def on_add(self) -> None:
        super().on_add()
        self._hue_cluster_handler.cluster.add_listener(self)

    def cluster_command(self, tsn: int, command_id: int, args: list[Any]) -> None:
        """Handle a server-issued cluster command on 0xFC03."""
        if command_id != HUE_NATIVE_COMMAND_ID:
            return
        payload: t.SerializableBytes = args[0]
        try:
            frame = decode(bytes(payload.serialize()))
        except ValueError:
            _LOGGER.exception("Failed to decode Hue native frame: %s", payload)
            return
        self._apply_frame(frame)

    def _apply_frame(self, frame: HueFrame) -> None:
        if frame.on_off is not None:
            self._state = frame.on_off
        if frame.brightness is not None:
            self._brightness = frame.brightness
        if frame.color_xy is not None:
            self._xy_color = frame.color_xy
        if frame.color_mirek is not None:
            self._color_temp = frame.color_mirek
        if frame.effect_type is not None:
            self._effect = HUE_TO_EFFECT_NAME.get(frame.effect_type, "none")

    async def _async_turn_on_impl(
        self,
        *,
        transition: float | None,
        brightness: int | None,
        effect: str | None,
        flash: Any,
        color_temp: int | None,
        xy_color: tuple[float, float] | None,
        duration: float,
        **_unused: Any,
    ) -> None:
        del flash, _unused  # Flash uses the Identify cluster, not Bifrost.
        frame = HueFrame(
            on_off=True,
            brightness=_clamp_brightness(brightness),
            color_mirek=color_temp,
            color_xy=xy_color,
            fade_speed=_seconds_to_fade_speed(transition or duration),
            effect_type=EFFECT_NAME_TO_HUE.get(effect) if effect else None,
        )
        await self._send_frame(frame)
        if effect and EFFECT_NAME_TO_HUE.get(effect) not in (None, HueEffect.NO_EFFECT):
            asyncio.get_running_loop().call_later(
                _POST_EFFECT_REFRESH_DELAY,
                lambda: asyncio.create_task(self._refresh_state()),
            )

    async def async_turn_off(self, *, transition: float | None = None) -> None:
        await self._send_frame(
            HueFrame(
                on_off=False,
                fade_speed=_seconds_to_fade_speed(transition),
            )
        )

    async def _send_frame(self, frame: HueFrame) -> None:
        payload = encode(frame)
        _LOGGER.debug("Hue native frame: %s", payload.hex())
        await self._hue_cluster_handler.cluster.hue_native(t.SerializableBytes(payload))

    async def _refresh_state(self) -> None:
        # OnOff/Level/Color report standard attributes; trigger a re-poll.
        await self._on_off_cluster_handler.cluster.read_attributes(
            ["on_off"], allow_cache=False
        )

@register_device
class HueLcx004(Device):
    """Hue Play gradient lightstrip (LCX004) — Bifrost-native control."""

    _device_match = DeviceMatch(
        manufacturers=frozenset({PHILIPS, SIGNIFY}),
        models=frozenset({"LCX004"}),
    )
    _operations = (
        Replace(
            endpoint_id=11,
            cluster_id=HUE_FC03_CLUSTER_ID,
            cluster_type=ClusterType.Server,
            replacement=HueLightCluster,
        ),
    )

    def discover_entities(self):
        for entity in super().discover_entities():
            # Skip existing Light entities
            if entity.PLATFORM == Platform.LIGHT:
                continue
            yield entity

        # Register new HueNativeLight entities
        for ep_id, endpoint in self.endpoints.items():
            if ep_id == 0:
                continue

            handlers = endpoint.all_cluster_handlers
            on_off = handlers.get(f"{ep_id}:0x{OnOff.cluster_id:04x}")
            fc03 = handlers.get(f"{ep_id}:0x{HUE_FC03_CLUSTER_ID:04x}")
            if on_off is None or fc03 is None:
                continue

            # Will be cleaned up once device handlers are removed
            cluster_handlers = [on_off, fc03]
            for opt_id in (LevelControl.cluster_id, Color.cluster_id):
                ch = handlers.get(f"{ep_id}:0x{opt_id:04x}")
                if ch is not None:
                    cluster_handlers.append(ch)

            yield HueNativeLight(
                cluster_handlers=cluster_handlers,
                endpoint=endpoint,
                device=self,
            )
```

I think we can greatly minimize the Python code required to implement custom entities with a better API. This has a few benefits:
1. Complex devices can be supported via quirks directly. I originally planned this for a Tuya IR module and it would work the same.
5. Quirks v2 can essentially be inverted to be a functional API built on top of this one. It would still be the preferred way to write quirks for simple devices. The only use case for this API is _complex_ devices that have cross-entity interaction, hook into ZHA internals, etc.
6. Zigpy can slowly shed the quirks APIs that (IMO) don't belong in it.


puddly:
Here's the zigpy side: https://github.com/zigpy/zigpy/compare/dev...puddly:puddly/quirks-v3-resolver
And quirks: https://github.com/zigpy/zha-device-handlers/compare/dev...puddly:puddly/decouple-zigpy

I wouldn't recommend it but you can run ZHA with these three library branches in tandem. Nothing is visibly broken 😅

jaredjxyz:
Hey @puddly I stumbled upon this after doing a little investigating myself into what it would take to get the Hue bulb's 0xFC03 working myself - and came to a lot of the same conclusions you seem to have come to (that the responsibilities are backward and you should be implementing the quirk behavior in the quirk) in this PR, so happily surprised to see this being worked on.

Crossing my fingers that something like this eventually lands, It looks like just a POC for now, lots of changes here. But it looks to my (again, relatively new to the codebase but experienced in SW development) eye to be a promising direction, at least for my use case

Thanks for working on this!

Hedda:
> Hey @puddly I stumbled upon this after doing a little investigating myself into what it would take to get the Hue bulb's 0xFC03 working myself - and came to a lot of the same conclusions you seem to have come to (that the responsibilities are backward and you should be implementing the quirk behavior in the quirk) in this PR, so happily surprised to see this being worked on.

@jaredjxyz For reference, FYI, also see these related requests/discussions asking for zha and ZHA Device Handlers quirks to be able to nativly support and handle these type of "Hue native control" Dynamic Scenes Effects / Animations. Checkout:

* https://github.com/zigpy/zha-device-handlers/issues/2517

and

* https://github.com/zigpy/zha/issues/756

Note! Another device that could also use native control of effect commands via quirks is Lidl Christmas led string HG0646:

* https://github.com/zigpy/zha-device-handlers/issues/1933

Lidl Christmas led string HG0646 uses cluster 0x00FC to invoke invoked native effects commands for their device:

* https://gist.github.com/enyone/e1e0182f41ab9924d33fd24d77765aa6#file-lidltuyaeffects-py-L91

  * https://gist.github.com/enyone/e1e0182f41ab9924d33fd24d77765aa6#file-lidltuyaeffects-py-L98

    * https://github.com/dresden-elektronik/deconz-rest-plugin/issues/3716#issuecomment-735004813

Orange-GT3:
@puddly - I'm putting my hand up to help test when the time is right. If that is out of the normal beta cycle, that's OK too but I'll need a muppet-proof method to install it.

I don't normally engage in testing the betas but I'm so keen to have full Hue support that I will for this :).

puddly:
I've given this approach a rough second pass and I think this might be the final shape?

1. Anything and everything related to "quirks" is completely gone from zigpy. Zigpy provides the base Zigbee device behaviors, a persistence layer for the Zigbee side into an SQLite database, and nothing more. There are import-time shims that redirect imports to `zhaquirks` for backwards compatibility.
   - Zigpy provides two hooks: `device_resolver` for replacing the base zigpy device at runtime and `uninitialized_packet_handler` for exposing the legacy hook we use to get older Aqara devices to join quickly, as they don't respond well to the "normal" device joining process.
2. ZHA provides the base entity model: what a `Light` looks like and does. This is intentionally a 1:1 copy of the Home Assistant Core entity model, with some minor tweaks for when Zigbee functionality extends it (that can maybe trickle back into Core). For now, ZHA will also own discovery itself for mostly-ZCL-compliant entities. Its discovery model is now fairly generic.
   - ZHA's own discovery hooks into zigpy's new `device_resolver`. ZHA itself doesn't use this feature though, it's strictly for quirks.
   - ZHA still has to provide a few hooks to zhaquirks for a few operations (overriding model/manufacturer names, directly mutating existing entity metadata, and filtering entities). Perhaps there's a better way to do this? Or maybe these shouldn't be moved out of ZHA itself (as they were in https://github.com/zigpy/zha/pull/762/commits/810340e945841f6c9cb58f164a1bdf701f9c52a8)?
4. The entire quirk builder API has been moved straight into `zhaquirks`. It registers quirked devices with ZHA using its new discovery/device registration API and will manage this entire API going forward.
   - Under the hood it's all the ZHA device discovery API, `QuirkBuilder()` is just a fluent interface for it. This API can be expanded later to provide better interfaces for things like "just swap out `ZCLLight` with a different class".

zha/application/helpers.py
```diff
@@ -374,6 +374,17 @@ class QuirksConfiguration:
...
+    # quirks provider, typically `zhaquirks.setup`. ZHA itself never imports a
+    # quirks package; when `enabled` is True this must be set or no quirks load.
+    setup_function: Callable[[str | None], None] | None = dataclasses.field(
+        default=None
+    )
+    # Injected by the consumer to short-circuit flaky joins (Xiaomi quick-init);
+    # forwarded to the controller as its uninitialized-packet handler.
+    uninitialized_packet_handler: Callable[..., None] | None = dataclasses.field(
+        default=None
+    )
```
TheJulianJES, line 387: Hmm, I get why we'd do this but as ZHA currently still require the quirks package itself, maybe it should just default to its `setup_function` and `uninitialized_packet_handler` (if it's installed)? I'm not sure..

I guess when we split ZHA out more, we'd have the ZHA server do this, and I think it's completely fine to do there. As HA Core currently plays that role, it might be fine as well. Just a thought.

TheJulianJES (commented):
I've scrolled through all the three branches now and I think I really like this approach now. Awesome job! 😄

`QuirkBuilder` being part of quirks is really good for faster development in that repo. `zha-quirks` now houses a lot of the (old/easy: v1/v2) quirk matching logic and more (e.g. ZHA metadata modifying).
Technically, that makes it harder for someone else to develop a quirks package for use with ZHA, without copying a lot of the base quirks logic (unless they want to build on the slightly more "lower-level" quirks logic in ZHA), but I guess nobody will make their own quirks package that replaces `zha-quirks` anyway... And splitting that off into a separate package would just defeat a lot of the gains we get by having it in the same repo.

The bot comments seem useful. I hope that fixing the re-interview logic doesn't become too much of an issue with this. I think we'll redo some of that in zigpy and ZHA anyways. Without this PR, IIRC HA currently has "ZHA device proxies" which reference ZHA devices and those reference zigpy devices. We can probably handle this better with the ZHA split at the latest.

zha/application/platforms/sensor/__init__.py
TheJulianJES, line 2113: Side note: Now that quirks depends on ZHA, I guess we could move all the exposed feature IDs to ZHA? The entities also live here after all.

puddly: Yeah, there's a lot of cascading cleanup that will need to be done after this PR. Do you want me to roll this cleanup into this one? Or do it in a followup?

TheJulianJES: Eh, that one is so minor, I think it could go into this one. But separate PRs are always good as well – might be nicer? I don't really care for this 😄

puddly:
The device swap out issue will be handled within Core: we can check device object identity to detect this.

There is a circular dependency between zha-device-handlers and this repo (for CI) so this will have to be merged bypassing checks.

tests/data/devices/ubisys-j1-5502.json
```diff
@@ -7,9 +7,11 @@
   "friendly_manufacturer": "ubisys",
   "friendly_model": "J1 (5502)",
   "name": "ubisys J1 (5502)",
-  "quirk_applied": false,
-  "quirk_class": "zigpy.device.Device",
-  "exposes_features": [],
+  "quirk_applied": true,
+  "quirk_class": "zhaquirks.ubisys.cover_j1:(ubisys / J1 (5502))",
```
TheJulianJES: I kinda wonder why this quirk didn't apply before?

TheJulianJES: Oh, the diagnostics just don't have EP 232, which `replaces` is used on. Before, the entire quirk failed to apply. Now it apparently applies successfully – should it?

TheJulianJES: Well, the "half interviewed" state is not something that should happen. But I think this behavior with the warning makes sense. I'll upload all of the more up-to-date Ubisys diagnostics later. Maybe I should also check when that EP got added.

puddly: Hmm. I've regenerated diagnostics for it for now but I'll have to track down where this diagnostics JSON came from.

puddly: Looks like this is the source 😄 https://github.com/home-assistant/core/issues/165380#issuecomment-4047201513

TheJulianJES: .........

puddly: It's a DIY device that someone half-cloned to _look_ like a Ubisys device... I think we can mark this resolved. I've deleted the diagnostics.

TheJulianJES (approved):
Looks good enough to me 😄

PR #767, opened 2026-05-19
Fix on/off-only lights reverting to off after rapid turn_off/turn_on

### LINKED ISSUE #766: `light_transitioning_flag` causes spurious state-revert on rapid turn_off→turn_on for some on/off-only lights
**The text below is generated by Claude (max effort) as the LLM was the one performing the investigation and eventually fixed the issue.
I've read it and it does match what actually happened. I can confirm the "workaround" section is working for me right now.**

The actual device that this is happening on is [MOES ZigBee Star Feather Wall Light Switch, 1 Gang](https://www.amazon.com/MOES-Touch-Way-Smart-Switch/dp/B0FGPP39N9) although Claude thinks its a broader issue, which I can't confirm or deny.
I'm also using this patch to enable actually turning the switch on/off: https://github.com/zigpy/zha-device-handlers/pull/5015
##

### Symptom

On a pure on/off Zigbee light (no Level or Color cluster), a light.turn_off followed within ~1.5 s by light.turn_on results in the entity state correctly transitioning off → on, then silently flipping back to off ~1.5 s after the original turn_off, despite the device being physically on and never receiving any further command.

State history of a turn_on, turn_off, turn_on rapid sequence (200 ms gaps):

T+0      state=on    (cmd 1)
T+0.22s  state=off   (cmd 2)
T+0.45s  state=on    (cmd 3)
T+1.73s  state=off   ← spurious; same context_id as cmd 3's call_service
The recorder's events/states tables show one call_service event tied to two state changes via the same context_id. No additional service call, automation, or Read_Attributes runs in the gap.

### Root cause

async_turn_off unconditionally sets _transitioning_individual = True and schedules async_transition_complete for DEFAULT_ON_OFF_TRANSITION + DEFAULT_EXTRA_TRANSITION_DELAY_SHORT (~1.5 s):

```python
# zha/application/platforms/light/__init__.py — async_turn_off
if self._zha_config_enable_light_transitioning_flag:
    self.async_transition_set_flag()
...
if self._zha_config_enable_light_transitioning_flag:
    self.async_transition_start_timer(transition_time)  # ~1.5 s
While that flag is set, handle_cluster_handler_attribute_updated buffers every on_off event instead of writing state:


if self.is_transitioning:
    if not event.attribute_value:
        self._transition_brightness_buffer = 0
    return
So cmd 2's optimistic on_off=False event sets _transition_brightness_buffer = 0. Cmd 3's on_off=True event is also buffered (and discarded, since the if not event.attribute_value check is False) — but it does not clear the existing 0. Cmd 3 directly writes self._state = True (line 541 of _async_turn_on_impl's simple-on path), but doesn't touch the buffer.
```

When cmd 2's timer fires, async_transition_complete reads the still-zero buffer and overrides _state back to False:

```python
if self._transition_brightness_buffer is not None:
    if self._transition_brightness_buffer == 0:
        # The device is actually off; override the optimistic on-state.
        self._state = False
```

The bug is the asymmetry between async_turn_off (always sets the flag) and async_turn_on (only sets it for brightness/color paths). A turn_off-then-turn_on within the 1.5 s window leaves cmd 2's "off intent" able to override cmd 3's "on" outcome.

### Affected scope

Any on/off-only light when light_transitioning_flag is enabled (it's on by default for many users). The bug doesn't depend on a specific quirk — reproduces with stock _TZ3000_* TS0011/TS0012 switches too. Discovered while debugging a Moes _TZE200_stvgmdjz TS0601, but isolated by disabling the flag globally and re-running the same rapid-toggle test on a _TZ3000_* device.

### Proposed fix

Pick one:

Mirror the asymmetry — in async_turn_off, only set the flag if brightness_supported. Matches the set_transition_flag predicate in async_turn_on. Minimal change; preserves the flag mechanic for the lights that actually need it.
Clear buffer on later opposing event — in the is_transitioning branch of handle_cluster_handler_attribute_updated, when event.attribute_value is truthy, set _transition_brightness_buffer = None. Slightly broader; addresses the root buffering inconsistency.
Skip the buffer override in async_transition_complete for non-dimmable lights — guard the _state = False write with brightness_supported. Safest but most narrow.
I'd lean toward (1).

### Workaround

Set light_transitioning_flag: false in ZHA options (Settings → Devices & Services → ZHA → Configure). The flag has no useful function for on/off-only lights and disabling it fully fixes the symptom.

### Affected version

Latest version installed via Zigbee Home Automation in HomeAssistant as of creation of this issue (couldn't find an actual version number).

TheJulianJES:
## Proposed change

This fixes an issue where on/off-only lights (so ones that do not support brightness) incorrectly revert their state to "off" when quickly turning the light off, and then back on again.

The underlying issue was that we only set the transitioning flag in `turn_on` for lights supporting brightness, but always in `turn_off`, which should not have been the case. Thus, the `turn_on` call didn't cancel/restart the `turn_off` timer.

To fix this, we just do not set the flag for on/off-only lights. They do not need at all, since they have no brightness/transitions, so we can always parse attribute reports.

## Additional information

- Fixes https://github.com/zigpy/zha/issues/766

As a little background: The transitioning flag and timer is needed because Zigbee lights always send their current brightness, even during ongoing transitions.

However, in Home Assistant, we need to show the target brightness instead. Otherwise, you see the brightness slider going back down instantly after moving it up, then slowly moving back up (but only for lights that support attribute reporting). There's also some other issues this fixes (including for group lights that would have very jumpy/weird state otherwise, especially if some lights do not support attribute reporting).

In general, the light class needs to cover a lot(!) of edge-cases, and generally does so successfully. But there are plans to clean up the entire class a bit, e.g. with https://github.com/zigpy/zha/pull/734.

## AI summary

<details><summary>Useful-ish AI summary (click to expand)</summary>
<p>

Fixes #766. `async_turn_off` unconditionally armed the transitioning flag and a ~1.5s timer whenever `_zha_config_enable_light_transitioning_flag` was set. For on/off-only lights this causes a rapid `turn_off → turn_on` sequence to revert to off after the timer fires, even though the device is on and never received another command.

## Root cause

`async_transition_set_flag` (`light/__init__.py:734`) is the only path that resets `_transition_brightness_buffer` and cancels the pending timer. `async_turn_on` calls it only when its `set_transition_flag` predicate is true:

```python
set_transition_flag = (
    brightness_supported or color_temp is not None or xy_color is not None
) and self._zha_config_enable_light_transitioning_flag
```

For an on/off-only light, `brightness_supported` is False and there are no `color_temp` / `xy_color` runtime params, so `async_turn_on` never resets the buffer or cancels the timer left over from `async_turn_off`. The sequence:

1. `turn_off`: flag armed, 1.5s timer scheduled, optimistic `_state = False`
2. Device's `on_off=False` report arrives → `_transition_brightness_buffer = 0` (because `is_transitioning` is True)
3. `turn_on`: `set_transition_flag` is False → `async_transition_set_flag` is **not** called → buffer and timer survive; the only state change is the optimistic `_state = True`
4. Device's `on_off=True` report arrives → ignored: the buffered-attribute handler only resets the buffer when the value is falsy
5. Original timer fires → `async_transition_complete` reads `buffer == 0` and overrides `_state = False` ⇒ revert to off

## Fix

Mirror `async_turn_on`'s predicate in `async_turn_off`: gate the flag set, timer start, and stuck-cleanup on `brightness_supported`. On/off-only lights have no transition to wait for, so the flag had no useful purpose for them.

Brightness-supporting lights are unaffected: any follow-up `turn_on` already arms the flag (since `brightness_supported` is true), which resets the buffer and cancels the previous timer cleanly. That existing path is unchanged.

A narrow theoretical edge case — a color-only light with no Level cluster, where `turn_on` is called with no `color_temp` / `xy_color` — is also covered by gating on `brightness_supported`. Such hardware is rare in practice.

## Regression test

Added `test_on_off_only_rapid_toggle_does_not_revert`: on an on/off-only light, calls `turn_off` then `turn_on`, replays both attribute reports, sleeps past the old 1.5s window, asserts state stays on. Verified the test fails on `dev` HEAD without the fix and passes with it.

</p>
</details>

### DIFF
--- a/tests/test_light.py
+++ b/tests/test_light.py
@@ -2069,6 +2069,61 @@ async def test_turn_on_during_off_transition(zha_gateway: Gateway) -> None:
     assert bool(entity.state["on"]) is False
 
 
+@patch(
+    "zigpy.zcl.clusters.general.OnOff.request",
+    new=AsyncMock(return_value=[sentinel.data, zcl_f.Status.SUCCESS]),
+)
+async def test_on_off_only_rapid_toggle_does_not_revert(
+    zha_gateway: Gateway,
+) -> None:
+    """Test rapid turn_off then turn_on on an on/off-only light does not revert to off.
+
+    On/off-only lights have no brightness transition to wait for, so turn_off
+    must not arm the transitioning flag/timer. Otherwise a buffered on_off=False
+    report from the off command would override the optimistic on-state when
+    the timer fires after the follow-up turn_on.
+    """
+    zigpy_device = create_mock_zigpy_device(zha_gateway, LIGHT_ON_OFF)
+    on_off_cluster = zigpy_device.endpoints[1].on_off
+    on_off_cluster.PLUGGED_ATTR_READS = {"on_off": 1}
+    zha_device = await join_zigpy_device(zha_gateway, zigpy_device)
+
+    entity = get_entity(zha_device, platform=Platform.LIGHT)
+    assert bool(entity.state["on"]) is True
+
+    # Turn off, then immediately turn back on (within the default ~1.5s window).
+    await entity.async_turn_off()
+    await zha_gateway.async_block_till_done()
+    assert bool(entity.state["on"]) is False
+    # On/off-only lights have no transition to wait for.
+    assert not entity.is_transitioning
+
+    # The device's on_off=False report from the off command arrives.
+    await send_attributes_report(
+        zha_gateway,
+        on_off_cluster,
+        {general.OnOff.AttributeDefs.on_off.id: 0},
+    )
+    await zha_gateway.async_block_till_done()
+
+    await entity.async_turn_on()
+    await zha_gateway.async_block_till_done()
+    assert bool(entity.state["on"]) is True
+
+    # The device's on_off=True report from the on command arrives.
+    await send_attributes_report(
+        zha_gateway,
+        on_off_cluster,
+        {general.OnOff.AttributeDefs.on_off.id: 1},
+    )
+    await zha_gateway.async_block_till_done()
+
+    # Wait past the old 1.5s timer window. State must remain on.
+    await asyncio.sleep(2)
+    await zha_gateway.async_block_till_done()
+    assert bool(entity.state["on"]) is True
+
+
 async def test_light_state_restoration(zha_gateway: Gateway) -> None:
     """Test the light state restoration function."""
     device_light_3 = await device_light_3_mock(zha_gateway)
--- a/zha/application/platforms/light/__init__.py
+++ b/zha/application/platforms/light/__init__.py
@@ -633,8 +633,13 @@ class BaseClusterHandlerLight(BaseLight):
             else DEFAULT_ON_OFF_TRANSITION
         ) + DEFAULT_EXTRA_TRANSITION_DELAY_SHORT
 
-        # Start pausing attribute report parsing
-        if self._zha_config_enable_light_transitioning_flag:
+        # Start pausing attribute report parsing, but not for on/off-only
+        # lights (they have no transition to wait for, and a follow-up
+        # turn_on would not clear the buffered off-report).
+        set_transition_flag = (
+            brightness_supported and self._zha_config_enable_light_transitioning_flag
+        )
+        if set_transition_flag:
             self.async_transition_set_flag()
 
         try:
@@ -654,7 +659,7 @@ class BaseClusterHandlerLight(BaseLight):
                 result = await self._on_off_cluster_handler.off()
 
             # Pause parsing attribute reports until transition is complete
-            if self._zha_config_enable_light_transitioning_flag:
+            if set_transition_flag:
                 self.async_transition_start_timer(transition_time)
             self.debug("turned off: %s", result)
             if result[1] is not Status.SUCCESS:
@@ -676,9 +681,7 @@ class BaseClusterHandlerLight(BaseLight):
         finally:
             # If the task was cancelled before the transition timer was started,
             # clean up the transitioning flag so the light does not get stuck.
-            self._async_cleanup_transition_if_stuck(
-                self._zha_config_enable_light_transitioning_flag
-            )
+            self._async_cleanup_transition_if_stuck(set_transition_flag)
 
     async def async_handle_color_commands(
         self,


tests/test_light.py
```diff
@@ -2069,6 +2069,61 @@ async def test_turn_on_during_off_transition(zha_gateway: Gateway) -> None:
...
+    await send_attributes_report(
+        zha_gateway,
+        on_off_cluster,
+        {general.OnOff.AttributeDefs.on_off.id: 1},
+    )
+    await zha_gateway.async_block_till_done()
+
+    # Wait past the old 1.5s timer window. State must remain on.
+    await asyncio.sleep(2)
+    await zha_gateway.async_block_till_done()
```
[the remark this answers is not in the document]
TheJulianJES: We already have this elsewhere and this isn't a real 2 second wait (due to `pytest-looptime`). I think this is fine.

tests/test_light.py
```diff
@@ -2069,6 +2069,61 @@ async def test_turn_on_during_off_transition(zha_gateway: Gateway) -> None:
...
+    on_off_cluster.PLUGGED_ATTR_READS = {"on_off": 1}
+    zha_device = await join_zigpy_device(zha_gateway, zigpy_device)
+
+    entity = get_entity(zha_device, platform=Platform.LIGHT)
+    assert bool(entity.state["on"]) is True
+
+    # Turn off, then immediately turn back on (within the default ~1.5s window).
+    await entity.async_turn_off()
+    await zha_gateway.async_block_till_done()
+    assert bool(entity.state["on"]) is False
```
puddly, line 2097: nit: is `bool(...)` needed? The data type of `on` should be `bool`.

tests/test_light.py
```diff
@@ -2069,6 +2069,61 @@ async def test_turn_on_during_off_transition(zha_gateway: Gateway) -> None:
...
+    # On/off-only lights have no transition to wait for.
+    assert not entity.is_transitioning
+
+    # The device's on_off=False report from the off command arrives.
+    await send_attributes_report(
```
puddly, line 2102: Not sure if it's useful to have as an explicit test but this also passes (as expected) when the reports both come in after the commands succeed (i.e. if you manage to spam the on/off toggle faster than the light reports).

puddly (approved):

PR #773, opened 2026-05-28
Use zigpy's mypy exclude generation tool for stricter types

puddly:
This PR copies the zigpy `python -m tools.regenerate_mypy_ignores` tool for granular typing excludes. I've migrated us to `pyproject.toml` with `ci` and `testing` sections.

As a test, I've fixed a few small `unreachable` and `invalid-type` issues in this PR as well.


.pre-commit-config.yaml
```diff
@@ -1,24 +1,30 @@
-ci:
-  autofix_commit_msg: "Apply pre-commit auto fixes"
-  autoupdate_commit_msg: "Auto-update pre-commit hooks"
-  skip: [mypy]
```
TheJulianJES, line 4: Any reason we remove this? We don't have it in the zigpy repo because the pre-commit CI GitHub action isn't used there, but we do use it for quirks and ZHA.

puddly: We would need to keep the mypy version synced between the pre-commit environment (for pre-commit) and the venv (for the internal tool). Using local pre-commit hooks uses the package straight from the venv but this sort of breaks pre-commit CI, since the tools are no longer self contained.

TheJulianJES: Ah yeah, I see. This wouldn't disable the GH action though. But it looks like you've already done that separately.

puddly: For the quirks repo I definitely see the pre-commit CI useful, since it does auto formatting for things edited with the GH web editor. I don't think it's used much for the ZHA repo, is it?

TheJulianJES: Yeah, I think there was a reason I enabled it, but maybe that was only the auto dependency bump. It's not critical for ZHA.

script/setup
```diff
@@ -27,8 +27,7 @@ if ! [ -x "$(command -v uv)" ]; then
   python3 -m pip install uv
 fi
 
-uv pip install -U pip setuptools pre-commit
-uv pip install -r requirements_test.txt
-uv pip install -e .
+uv pip install -U pip setuptools
+uv pip install -e . --group testing
```
TheJulianJES: We may even just want to use `uv sync` here, similar to what’s done in the quirks repo. I think the GH CI from our shared workflow will already do this now, so might be good to keep the setup script in sync?

TheJulianJES (approved):
Nice!

TheJulianJES (approved):

PR #782, opened 2026-06-02
Reduce number of virtual entities

puddly:
This PR reduces the number of virtual entities by propagating some of the implicit attribute config to their respective entities. The few remaining virtual entities can be moved to quirks in the near future.

### DIFF
--- a/zha/application/platforms/virtual.py
+++ b/zha/application/platforms/virtual.py
@@ -676,7 +676,7 @@ class AqaraMotionAc01Init(_AqaraOppleInitBase):
     )
     _server_cluster_config = {
         AQARA_OPPLE_CLUSTER: ClusterConfig(
-            bind=False,
+            bind=True,
             attributes={
                 "presence": AttrConfig(read_on_startup=False),
             },
@@ -694,7 +694,7 @@ class AqaraThermostatAgl001Init(_AqaraOppleInitBase):
     )
     _server_cluster_config = {
         AQARA_OPPLE_CLUSTER: ClusterConfig(
-            bind=False,
+            bind=True,
             attributes={
                 "system_mode": AttrConfig(read_on_startup=False),
                 "schedule": AttrConfig(read_on_startup=False),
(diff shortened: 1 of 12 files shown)


zha/application/platforms/virtual.py
```diff
@@ -799,51 +732,15 @@ class AqaraCurtainAgl001Init(_AqaraOppleInitBase):
...
-    async def async_initialize_cluster(self, cluster: zigpy.zcl.Cluster) -> None:
-        """Mirror detection_interval into the sibling IAS Zone handler."""
-        interval = cluster.get("detection_interval", cluster.get(0x0102))
-        if interval is None:
-            return
-        ias_zone = getattr(cluster.endpoint, "ias_zone", None)
-        if ias_zone is None:
-            return
-        self.debug("Loaded detection interval at startup: %s", interval)
-        ias_zone.reset_s = int(interval)
```
TheJulianJES, line 844: This logic was always weird to have inside of ZHA. We still need to keep it somewhere (and not just delete it completely). But I don't think we can restore `reset_s` on startup / appdb load entirely in quirks. `__init__` won't have any cached attributes loaded, and I'm not sure where else this could be put to load on startup.

But maybe we can just add a "reset_after" `property` to [the cluster](https://github.com/zigpy/zha-device-handlers/blob/386e5ec3563fdbfce2ed370c64a338e9d92d2a4e/zhaquirks/__init__.py#L304-L307), with it defaulting to the local `reset_s` attribute, but then override it with the attribute access to `detection_interval` in the Aqara quirk?
That way, we don't need to load `reset_s` right on startup / appdb load, but just when it's needed. And we always have the fresh value, so we can also remove the override when changing/writing the attribute ([here](https://github.com/zigpy/zha-device-handlers/blob/386e5ec3563fdbfce2ed370c64a338e9d92d2a4e/zhaquirks/xiaomi/aqara/motion_ac02.py#L64)). I think that's a bit cleaner in general.

TheJulianJES: This is what I imagine to make this work without any special logic in ZHA:
- https://github.com/zigpy/zha-device-handlers/pull/5067

PR #788, opened 2026-06-10
Fix discovery crash from quirk removing ZCL attributes

TheJulianJES:
DRAFT.

## Proposed change

This fixes an issue where a (custom) quirk can remove standard ZCL attributes. Most entity platforms already have guards checking if the attribute even exists, but some do not. This adds them.

## Additional information

I'm not sure if this is something we should add – quirks shouldn't misbehave like this. Or are there valid use-cases for deleting ZCL attributes...? But currently, ZHA startup breaks completely when using [these custom quirks](https://github.com/jacekk015/zha_quirks).

Should address:
- https://github.com/home-assistant/core/issues/173265
- https://github.com/home-assistant/core/issues/178581

This "regression" was introduced with:
- https://github.com/zigpy/zha/pull/657

### AI summary

<details>
<summary>Issue and fix summary (CLICK TO EXPAND)</summary>

### Issue

A quirk can leave a cluster without a definition for an attribute that ZHA entity code reads. zigpy's `Cluster.is_attribute_unsupported()`, `find_attribute()` and `Cluster.get()` all raise `KeyError` for such a name (`get()` resolves the definition *outside* its own `try`, so it raises rather than returning the default).

That `KeyError` propagated through `Device._add_pending_entities()` → `Gateway.load_devices()` → HA's `async_setup_entry`, so **one broken quirk prevented the entire ZHA integration from starting** (`ConfigEntryNotReady` retry loop) — for every device, not just the affected one. Before #657 the cluster-handler-based code tolerated these clusters.

Two shapes of the same bug were reported:

- **home-assistant/core#173265** — a quirk *removes* standard attribute definitions. `ts0601_trv_moes.py` from [jacekk015/zha_quirks](https://github.com/jacekk015/zha_quirks) does `attributes = LocalDataCluster.attributes.copy()` on an `OnOff` cluster, leaving no `on_off` definition, and `Switch._is_supported()` then raised `KeyError: 'on_off'`.
- **home-assistant/core#178581** — a quirks v2 quirk *names* an attribute its cluster never defines (a custom Inovelli VZM32-SN quirk declaring a `relay_click_in_on_off_mode` switch, an attribute only the VZM30/VZM31 clusters have). Quirk entities are constructed with `_attr_always_supported = True`, and `BaseEntity.is_supported()` short-circuits on that *before* `_is_supported()` runs — so per-platform guards can never catch this case.

### Fix

A new `BaseEntity._is_valid()` hook, checked **before** the `_attr_always_supported` short-circuit. `PlatformEntity` implements it as "the attributes this entity is built around resolve on the cluster" (`_attribute_name` and `_inverter_attribute_name`, which `inverted` reads on every state computation). A missing definition is routine for default discovery (debug log — an optional attribute a quirk removed) but an authoring bug when a quirk explicitly asked for the entity (warning log).

That centralises the duplicated `attributes_by_name` membership checks previously scattered across switch/select/number/sensor/binary_sensor.

Guards that `_is_valid()` cannot replace, because they run over cluster configs aggregated *before* `is_supported()` filtering:

- `configure_cluster_configs()` — skips reporting configuration for attributes the cluster does not define.
- `initialize_cluster_configs()` — filters undefined names out of the read batch. `read_attributes()` resolves every name up front, so leaving one in failed the whole batch and left its valid siblings uninitialized.
- `AggregatedClusterPoller.async_update()` — treats an undefined attribute as unsupported instead of raising.

Finally, the crash class is closed at its seam rather than per call site. `recompute_capabilities()` and `is_supported()` read attributes that are *not* the entity's own `_attribute_name` — `min_present_value` on `AnalogOutput`, `description` on `BinaryOutput`, `zone_type` on `IasZone` — and both ran unguarded in `_add_pending_entities()`, so a quirk gutting any of those clusters still failed the whole setup. Both calls now go through `Device._entity_supported()`, which treats an entity that cannot answer the question as unsupported, mirroring the containment `_discover_new_entities()` already gives entity construction. `Device._add_pending_entities()` likewise no longer lets one entity's state computation abort device initialization.

Removal is deliberately not symmetric: a *prospective* entity that raises is dropped, but an *existing* one is kept, because removing it makes consumers delete its registry entry along with the user's customizations.

### Tests

Regression tests for each shape, all verified to fail on unpatched `dev`: a quirk replacing `OnOff` with a cluster whose `AttributeDefs` does not inherit the standard definitions; a quirks v2 switch naming a nonexistent attribute; the same for a nonexistent *inverter* attribute; a valid sibling attribute surviving a poisoned read batch; a quirk gutting `AnalogOutput` (the `recompute_capabilities()` path); and the state-emission and capability-check guards themselves.

`python -m tools.regenerate_diagnostics` produces no snapshot drift, so no currently shipped quirk relies on an entity this now skips.

</details>

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -2494,3 +2494,35 @@ async def test_broken_entity_state_does_not_break_device_init(
         await zha_device.async_initialize(from_cache=True)
 
     assert "Failed to emit state changed event" in caplog.text
+
+
+async def test_recompute_keeps_entity_that_cannot_answer(
+    zha_gateway: Gateway, caplog: pytest.LogCaptureFixture
+) -> None:
+    """An existing entity is kept when its capability check raises.
+
+    Removing it is destructive - consumers delete its registry entry, losing
+    the user's customizations - so a broken check must not be treated as
+    "unsupported" for an entity that already exists.
+    """
+    zigpy_device = create_mock_zigpy_device(
+        zha_gateway,
+        {
+            1: {
+                SIG_EP_INPUT: [general.Basic.cluster_id, general.OnOff.cluster_id],
+                SIG_EP_OUTPUT: [],
+                SIG_EP_TYPE: zigpy.profiles.zha.DeviceType.ON_OFF_SWITCH,
+                SIG_EP_PROFILE: zigpy.profiles.zha.PROFILE_ID,
+            }
+        },
+    )
+    zha_device = await join_zigpy_device(zha_gateway, zigpy_device)
+    entity = get_entity(zha_device, platform=Platform.SWITCH)
+
+    with patch.object(
+        type(entity), "recompute_capabilities", side_effect=KeyError("on_off")
+    ):
+        await zha_device.recompute_entities()
+
+    assert "treating it as supported" in caplog.text
+    assert get_entity(zha_device, platform=Platform.SWITCH) is entity
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -2460,3 +2460,37 @@ async def test_async_reinterview_device_active_coordinator(
 
     assert "Skipping reinterview for active coordinator" in caplog.text
     mock_reinterview.assert_not_called()
+
+
+async def test_broken_entity_state_does_not_break_device_init(
+    zha_gateway: Gateway, caplog: pytest.LogCaptureFixture
+) -> None:
+    """A single entity raising while computing its state must not fail setup.
+
+    Computing a state runs entity (and, for quirk entities, quirk-supplied)
+    code; letting it propagate would abort device initialization and, via
+    `load_devices`, take every other device down with it.
+    """
+    zigpy_device = create_mock_zigpy_device(
+        zha_gateway,
+        {
+            1: {
+                SIG_EP_INPUT: [general.Basic.cluster_id, general.OnOff.cluster_id],
+                SIG_EP_OUTPUT: [],
+                SIG_EP_TYPE: zigpy.profiles.zha.DeviceType.ON_OFF_SWITCH,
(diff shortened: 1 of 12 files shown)


TheJulianJES:
@zigpy-review-bot Update this PR, possibly make changes, and do review rounds before pushing. Check if it should be merged as is then. (You can also update the PR description if it makes sense.)

TheJulianJES:
@zigpy-review-bot Can we improve test coverage? `zha/application/platforms/sensor/__init__.py` seems to have the `except KeyError: return True` uncovered and `continue` for `if not entity.enabled` in `async_update`.

Also, should this PR be split up into multiple ones? Don't do it yet, just think how many we would/should have.

TheJulianJES:
@zigpy-review-bot You are allowed to push the test coverage improvement (and whatever else you have parked).
I'll think about the split suggestions soon.

TheJulianJES:
@zigpy-review-bot Tests failed with this run: https://github.com/zigpy/zha/actions/runs/32916926775/job/98022510877?pr=788. Restarted it and it went through, but create an issue about that for now. Also create one for "Warning: 'asyncio.get_event_loop_policy' is deprecated and slated for removal in Python 3.16", search related PRs (opened) in this repo and also search HA Core for PRs regarding this. I believe balloob or balloobbot may have had one regarding this recently. Open an issue and link everything relevant.

TheJulianJES:
@zigpy-review-bot Create issues, it's explicitly allowed. If the fixes are clear, create PRs for them, do review rounds before opening. If there are already PRs from me for those issues, and only small changes are needed, push follow-up commits to them (but do review rounds before), else, review existing PRs. If there are other PRs that are too big or include other fixes that should go in separate PRs, open separate PRs (with review rounds).

PR #799, opened 2026-06-24
Import device diagnostics with a firmware version filename suffix

puddly:
This PR appends the firmware version string to the diagnostics filename, allowing us to test version-gated quirks.

I built this PR in a specific way to ensure that the recent ZHA rewrite doesn't have any entity-level regressions:
- All existing diagnostics have been re-imported from scratch with ZHA 1.4.1
- All new (and old) diagnostics were imported with ZHA 1.4.1
- `dev` was merged
- Diagnostics were regenerated

The last step succeeded with just `quirk_class` renames.


tools/import_diagnostics.py
```diff
@@ -219,23 +219,21 @@ def zigpy_device_from_legacy_diagnostics(  # noqa: C901
...
-                real_cluster.PLUGGED_ATTR_READS[int(attr_id, 16)] = parse_legacy_value(
-                    attr["value"]
-                )
+                attr_id_int = int(attr_id, 16)
+                parsed = parse_legacy_value(attr["value"])
+                if attr_id_int in real_cluster.attributes:
+                    real_cluster._attr_cache[attr_id_int] = parsed
+                else:
+                    real_cluster._attr_cache.set_legacy_value(attr_id_int, parsed)
+                real_cluster.PLUGGED_ATTR_READS[attr_id_int] = parsed
```
TheJulianJES, line 238: Should we log when we set legacy values? I think they should almost never be used these days (except for maybe the Aqara P1 quirk).

(In some future PR, maybe we can even have regeneration + diagnostics import print file name + what warnings occurred as a summary, but idk.)
(Also related issue of attr ID collisions: https://github.com/zigpy/zha/issues/798. I fear that we may deviate from real devices too much at some point)

tools/import_diagnostics.py
```diff
@@ -340,31 +340,30 @@ async def main(paths: list[str]):
...
 
-            if zigpy_device is None:
-                _LOGGER.debug("Skipping, diagnostics are not valid")
-                continue
+                ha_data = data["home_assistant"].get("data", {})
+                if (
+                    ha_data.get("last_seen") == "2026-03-04T18:21:39.038488+00:00"
+                ) and (ha_data.get("model") == "J1 (5502)"):
+                    _LOGGER.debug("Skipping known-bad DIY device")
+                    continue
```
TheJulianJES, line 385: Should we have some constant(s) for bad imports? But I guess we hopefully don't need to expand this list ever 😅

TheJulianJES (approved):
This is a nice improvement!

tools/import_diagnostics.py
puddly: There are a few hundred 😓 [`66620ab` (this PR)](https://github.com/zigpy/zha/pull/799/commits/66620ab48b418e6cbd68942761baa2c1b3c22331)

TheJulianJES: Oh... I guess we should look into what those unmapped attributes are 😅 (https://github.com/zigpy/zha-device-handlers/issues/5120)

PR #803, opened 2026-06-25
Split off ZHA diagnostics files without an original signature

puddly:
Effectively, device dumps sourced from diagnostics prior to ZHA having `original_signature` are separated out into `legacy_devices/`. Everything in `devices/` should be accurate (for the most part, not accounting for ZCL attribute mutation).

The diff will display strangely but the individual commits should make the changes clear.


zha/application/platforms/binary_sensor/__init__.py
```diff
@@ -111,7 +111,13 @@ def __init__(
             )
 
         super().__init__(endpoint=endpoint, device=device, **kwargs)
-        self._state: bool = self.is_on
+
+        if not self.is_supported():
+            # `self.is_on` can raise if an attribute is missing
+            self._state = False
+        else:
+            self._state = self.is_on
```
TheJulianJES, line 119: I thought this was just https://github.com/zigpy/zha/pull/788 at first but that's something different. Then thought this should be a separate PR. But looking at the code, we never read `self._state` – it's only ever set. I think we can just remove `self._state` entirely then? `is_on` does access the attribute value directly, like all other entities (but that's guarded by `is_supported()` then).

PR #804, opened 2026-06-26
Merge `info_object` into `state` and emit state change diffs

puddly:
This PR replaces the `state` dict with a strongly typed dataclass that absorbs `info_object`. We now also emit state change diffs in `EntityStateChangedEvent`'s new `state_diff` attribute.

For stronger typing and to support the client/server architecture split in the future, we benefit from having more centralized state for entities. This will allow us to send batched state updates to the client instead of the client being notified of a general "update" and then reaching for the `state` dict of the ZHA entity.

The HA Core half is here: https://github.com/home-assistant/core/compare/dev...puddly:core:puddly/zha-state-migration. It uses the `state_diff` values and also adds dynamic capability recomputation.

Note: I intentionally regenerated diagnostics and reverted the regeneration commit so that this diff is minimal. Otherwise, it'll be a hundred of thousand lines of code. This will be un-reverted before merge.

---

Part of https://github.com/OpenHomeFoundation/roadmap/issues/93 and a continuation of https://github.com/zigpy/zha/pull/663.


zha/application/platforms/light/__init__.py
```diff
@@ -357,12 +352,12 @@ def _gateway(self) -> Gateway:
         """Return the gateway."""
 
     @property
-    def state(self) -> dict[str, Any]:
+    def state(self) -> LightState:
         """Return the state of the light."""
-        response = super().state
-        # XXX: for backwards compatibility
-        response["supported_color_modes"] = self._internal_supported_color_modes
```
TheJulianJES, line 364: Maybe we should keep the comment until we remove that here?

zha/application/platforms/sensor/__init__.py
```diff
@@ -213,22 +211,16 @@ def native_unit_of_measurement(self) -> str | None:
...
-            **super().info_object.__dict__,
+    @property
+    def state(self) -> SensorState:
+        """Return the state for this sensor."""
+        return SensorState(
+            **super().state.__dict__,
+            native_value=self.native_value,
             suggested_display_precision=self.suggested_display_precision,
             unit=self.native_unit_of_measurement,
         )
```
TheJulianJES, line 222: This is something we may want to do in a follow-up (but I have no problem with it going here instead):

We're currently not exposing `_attr_options` for the `EnumSensor`. We're adding `suggested_display_precision` + `suggested_display_precision` to the `state` of the `BaseSensor` here, so I wonder if we should also expose the `options` here, or just in `EnumSensor`? If the latter, should the `EnumSensor` remove the unit + suggested display precision? I guess it would be more correct?


Without the changes from this PR, the fix in HA would look something like this (which isn't great but it's currently not exposed in `state` at all, neither without or with this PR): [home-assistant/core@`dev...TheJulianJES:core:tjj/zha-enum-sensor-options`#diff-97adf01229](https://github.com/home-assistant/core/compare/dev...TheJulianJES:core:tjj/zha-enum-sensor-options#diff-97adf01229931f17610655838232182c16a7f10291291d780cfd472be187406eR144-R149)

puddly: The ZHA `BaseSensor` inheritance chain and `sensor.py` in general have very...dubious typing right now. I tried in the past to properly separate sensors into their various categories (integer, floating point, string/enum, datetime, and I think a few others) but it grew into a very large PR. We definitely should look into properly typing them and cleaning up these inconsistencies.

TheJulianJES (approved):
Other than the tests / diagnostics regen, everything looks good to me. This is a really awesome improvement!

The HA Core side also looks good*. I'm not completely sure what to think about partly using `_update_capability_attrs` to update the "shorthand" `_attr_something` vs overriding the methods for those states but it all looks good for the next beta. (Maybe we do want to keep that approach and not cache everything in the library? I'm still a bit torn... 😅)

(*: `test_diagnostics.ambr` needs to be updated)

puddly:
Regarding `_update_capability_attrs`: we effectively recompute the whole state on every event with the current approach in my branch. We could do it more granularly by only calling `_update_capability_attrs` on actual capability attribute updates? Or use properties, as you said, and let Core re-read them and compute changes?

TheJulianJES:
> Regarding `_update_capability_attrs`: we effectively recompute the whole state on every event with the current approach in my branch.

Hmm, I'm not sure that's true? In HA, `self.async_write_ha_state()` does the heavy lifting* and shouldn't be called unnecessarily all the time. But just updating the short-hand `_attrs` from the ZHA state all the time, even if they're unchanged, shouldn't have any performance impact. "Basically nothing happens when you update a normal attribute".

(*: For changed capabilities that need to be updated in the entity registry, `async_write_ha_state` already checks if they really need to be updated: https://github.com/home-assistant/core/blob/b52e8646a819c47e8327a5b68d764b025d191d94/homeassistant/helpers/entity.py#L1217-L1226)

And `async_write_ha_state()` is just called on any "real events", which may have caused the other state attributes to change, so I think that's fine. That's the same behavior as before.

Only some things like `_convert_supported_features` are a bit unnecessary because they actually redo some calculation in HA, even though they're likely unchanged 99% of the time, but I also feel like the performance impact shouldn't be huge there to recalculate the enum. 

(@zigpy-review-bot Verify/check everything and answer.)

Or did you mean something else?

TheJulianJES:
Hmm, well. I guess the recalculations done in `_update_capability_attrs` for fan + light + siren + cover aren't ideal. I think others like select + number + climate are completely fine though – those just set HA attributes to ZHA ones.

I guess an easy way to "fix" this would be to make methods like `_convert_supported_features` (for cover) cached depending on the `zha_features: ZHACoverEntityFeature`, for example. That way, they'd only construct new HA enums from the ZHA ones if it's really necessary. But maybe a different approach might be nicer? (@zigpy-review-bot also consider this comment before answering)

puddly:
I'm surprised the `IntFlag` bitwise operations are so expensive. I think this should work?
```python
class ZHASupportedFeaturesEntity(ZHAEntity):
    """ZHA entity whose state carries `supported_features`."""

    @override
    def _update_capability_attrs(self) -> None:
        self._attr_supported_features = self._convert_supported_features(
            self._zha_state.supported_features
        )

    @staticmethod
    def _convert_supported_features(zha_features: IntFlag) -> IntFlag:
        raise NotImplementedError
```

And then:

```python
# fan.py
class Fan(ZHASupportedFeaturesEntity, FanEntity):
    @staticmethod
    @functools.cache
    @override
    def _convert_supported_features(
        zha_features: ZHAFanEntityFeature,
    ) -> FanEntityFeature:
        features = FanEntityFeature(0)
        if ZHAFanEntityFeature.DIRECTION in zha_features:
            features |= FanEntityFeature.DIRECTION
        # ... unchanged
        return features
```

It speeds things up in my local tests from 4000ns to 73ns.

PR #825, opened 2026-07-13
Fix missing config entity attribute reads

TheJulianJES:
## Proposed change

This PR fixes a regression introduced with the below PR where attributes for several config entities were not read when initially pairing the device, leading to missing entities. This is fixed by adding a cluster config to read the required attributes for those entities.

ZHA refactor PR that caused the issue (note: **not** the recent ZHA 2.0.0 refactor):
- #657

Addresses HA Core issue:
- https://github.com/home-assistant/core/issues/173224

This should also fix a reported issue where the Danfoss entities are missing:
- https://github.com/zigpy/zha/issues/827

### Future

In the future, we may want to move some of the entities to ZHA quirks. I'm not sure about the `ultrasonic_*`-based entities used by the Sonoff though. Maybe we should make them more generic instead? Similar to the `pir_*` attributes?

Either way, an initial branch moving all the Sonoff entities to quirks v2 (keeping the same unique ID) can be found here:
- Quirks: https://github.com/zigpy/zha-device-handlers/compare/dev...zigpy-bot/sonoff-presence-v2-entities
- ZHA: https://github.com/zigpy/zha/compare/dev...zigpy-bot/sonoff-presence-native-removal

I guess we can always move them back to ZHA as generic entities if quirks keeps the cluster ID in the unique ID, so the format is the same as used by ZHA. And we should also move the other entities, like Danfoss and Hue.

### Problem

<details><summary>LLM-generated text describing the problem with more detail (click to expand)</summary>
<p>

`ZCLEnumSelectEntity`, `NumberConfigurationEntity` and `ConfigurableAttributeSwitch` skip entity creation when `cluster.get(attribute_name)` returns `None` in `_is_supported()`. Before #657, the cluster handlers read these attributes during device initialization (`ZCL_INIT_ATTRS`/`REPORT_CONFIG`), so a value was cached and the entity was created. #657 moved the startup reads to per-entity `_server_cluster_config` declarations, but several entities never got one. Their backing attribute is now never read, so the entities are silently skipped on freshly-paired devices. Devices paired before the regression keep their entities only because a value is still in the database from a read performed by an older version.

Reported for the Sonoff SNZB-06P in home-assistant/core#173224: the "Detection sensitivity" and "Presence detection timeout" entities are missing for newly-paired sensors, while identical sensors paired earlier still have them. Debug logs from that issue confirm the interview never sends a read for `ultrasonic_o_to_u_delay`/`ultrasonic_u_to_o_threshold`, after which both entities are skipped with "`… is not supported - skipping … entity creation`".

</p>
</details> 

### Affected entities

All confirmed against the pre-#657 handlers (each attribute was read at startup via `ZCL_INIT_ATTRS`):

- **Sonoff SNZB-06P/03P** (OccupancySensing): `SonoffPresenceSenorTimeout` (`ultrasonic_o_to_u_delay`), `SonoffPresenceDetectionSensitivity` (`ultrasonic_u_to_o_threshold`)
- **Hue SML001/SML002/SML003/SML004** (OccupancySensing): `HueV1MotionSensitivity` and `HueV2MotionSensitivity` (`sensitivity`) — the V1 entity has a `_server_cluster_config` that reads `occupancy` and the PIR delays, but not its own backing attribute
- **Danfoss Ally** (Thermostat): `DanfossExerciseDayOfTheWeek`, `DanfossExerciseTriggerTime`, `DanfossOrientation`, `DanfossAdaptationRunControl`, `DanfossControlAlgorithmScaleFactor`, `DanfossExternalMeasuredRoomSensor`, `DanfossLoadRoomMean`, `DanfossRegulationSetpointOffset`

Other cache-gated entities are unaffected: the IKEA air purifier attributes are read via the fan entity's cluster config, and the Tuya `0xEF00` attributes (`timer_duration`, `window_detection_function`) are seeded into the attribute cache by the quirk and were never read by ZHA itself.

### Fix

<details><summary>LLM-generated text describing the fix in detail (click to expand)</summary>
<p>

Give each affected entity a `_server_cluster_config` that reads its backing attribute on startup. `read_on_startup` mirrors the pre-#657 semantics: attributes marked cached in `ZCL_INIT_ATTRS` use `read_on_startup=False` (allow-cache read, still hits the device when nothing is cached), and the two Danfoss attributes previously marked "can change" (`external_measured_room_sensor`, `load_room_mean`) use `read_on_startup=True`.

Device snapshots are unchanged: the recorded devices already carry cached values for these attributes, so they don't exercise the fresh-pair path.

</p>
</details> 

### Upgrade note

<details><summary>LLM-generated upgrade note (click to expand)</summary>
<p>

Affected mains-powered devices (e.g. the SNZB-06P) pick the values up on the next ZHA restart via startup polling, restoring the missing entities. Battery-powered devices (SNZB-03P, Hue motion sensors, Danfoss TRVs) are initialized from cache on restart, so they may additionally need a re-interview (Reconfigure with the device awake) or an unavailable/available cycle.

</p>
</details>

### DIFF
--- a/zha/application/platforms/select.py
+++ b/zha/application/platforms/select.py
@@ -542,10 +542,6 @@ class HueV1MotionSensitivity(ZCLEnumSelectEntity):
                 OccupancySensing.AttributeDefs.pir_u_to_o_delay: AttrConfig(
                     read_on_startup=False,
                 ),
-                # Hue-specific manufacturer attribute backing this entity;
-                # read on startup so the entity is created on freshly-paired
-                # devices (regression from #657 dropping the cluster handler
-                # `ZCL_INIT_ATTRS` that used to read it).
                 "sensitivity": AttrConfig(read_on_startup=False),
             },
         ),
@@ -581,8 +577,6 @@ class HueV2MotionSensitivity(ZCLEnumSelectEntity):
     _server_cluster_config = {
         OccupancySensing.cluster_id: ClusterConfig(
             attributes={
-                # See HueV1MotionSensitivity: read on startup so the entity is
-                # created on freshly-paired devices (regression from #657).
                 "sensitivity": AttrConfig(read_on_startup=False),
             },
         ),
@@ -1127,10 +1121,6 @@ class DanfossExerciseDayOfTheWeek(ZCLEnumSelectEntity):
                 Thermostat.AttributeDefs.setpoint_change_source_timestamp: AttrConfig(
                     read_on_startup=False,
                 ),
-                # Danfoss proprietary attribute backing this entity; read on
-                # startup so the entity is created on freshly-paired devices
-                # (regression from #657 dropping the Danfoss thermostat cluster
-                # handler `ZCL_INIT_ATTRS` that used to read it).
                 "exercise_day_of_week": AttrConfig(read_on_startup=False),
             },
         ),
--- a/zha/application/platforms/select.py
+++ b/zha/application/platforms/select.py
@@ -542,6 +542,11 @@ class HueV1MotionSensitivity(ZCLEnumSelectEntity):
                 OccupancySensing.AttributeDefs.pir_u_to_o_delay: AttrConfig(
                     read_on_startup=False,
                 ),
+                # Hue-specific manufacturer attribute backing this entity;
+                # read on startup so the entity is created on freshly-paired
+                # devices (regression from #657 dropping the cluster handler
+                # `ZCL_INIT_ATTRS` that used to read it).
+                "sensitivity": AttrConfig(read_on_startup=False),
             },
         ),
     }
@@ -572,6 +577,16 @@ class HueV2MotionSensitivity(ZCLEnumSelectEntity):
         manufacturers=frozenset({"Philips", "Signify Netherlands B.V."}),
         models=frozenset({"SML002", "SML003", "SML004"}),
     )
+
+    _server_cluster_config = {
+        OccupancySensing.cluster_id: ClusterConfig(
+            attributes={
+                # See HueV1MotionSensitivity: read on startup so the entity is
+                # created on freshly-paired devices (regression from #657).
+                "sensitivity": AttrConfig(read_on_startup=False),
+            },
+        ),
+    }
 
 
 class AqaraMonitoringModess(types.enum8):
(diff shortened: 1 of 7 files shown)


zha/application/platforms/select.py
```diff
@@ -542,6 +542,7 @@ class HueV1MotionSensitivity(ZCLEnumSelectEntity):
                 OccupancySensing.AttributeDefs.pir_u_to_o_delay: AttrConfig(
                     read_on_startup=False,
                 ),
```
TheJulianJES, line 544: Huh, I'm not sure how these attributes ended up in the cluster config of this Hue-specific select entity? These are used by a [number entity](https://github.com/zigpy/zha/blob/c38e03e66960e81d27737a70d2f32dfed6a28919/zha/application/platforms/number/__init__.py#L680-L685). (Side note: Why does that number entity also include the `occupancy` attribute for the binary sensor again?)

I'll check to see if other unintentional changes were introduced by that refactor and then remove these in a follow-up PR (to keep this PR as just the fix).

TheJulianJES: And the binary sensor also re-adds all the same attributes yet again? Why do all three platforms configure the exact same attributes in the cluster config? We merge these, right..?

https://github.com/zigpy/zha/blob/c38e03e66960e81d27737a70d2f32dfed6a28919/zha/application/platforms/binary_sensor/__init__.py#L249-L260

TheJulianJES:
The gaps mentioned above should be addressed, probably with something like the below branches:
- https://github.com/zigpy/zha/compare/dev...zigpy-bot/cluster-config-dedup
- https://github.com/zigpy/zha/compare/dev...zigpy-bot/657-followup-fixes

In the future, these manufacturer-specific entities will be moved to zha-quirks anyway. This PR is only for a HA patch release, so it's a more minimal fix. We should also really expand diagnostics here for what we (try to) bind/init/report.

puddly (approved):

zha/application/platforms/select.py
```diff
@@ -542,6 +542,7 @@ class HueV1MotionSensitivity(ZCLEnumSelectEntity):
...
+                "sensitivity": AttrConfig(read_on_startup=False),
```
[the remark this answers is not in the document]
TheJulianJES: Yeah, planned to do basically.

PR #828, opened 2026-07-14
Codex updated: Do not bind clusters whose entities are not created

dmulcahey:
This pull request was created with Codex and is an updated version of the pull request it supersedes.

Original pull request: https://github.com/zigpy/zha/pull/783

See https://github.com/zigpy/zha/issues/781.

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -1200,6 +1200,31 @@ async def test_join_binding_reporting(zha_gateway: Gateway) -> None:
     ]
 
 
+async def test_quirks_v2_prevent_entity_excludes_cluster_configuration(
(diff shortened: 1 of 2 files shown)


dmulcahey:
Closing because this changes `prevent_default_entity_creation` from entity suppression into a binding/reporting configuration policy without sufficient compatibility evidence. The Aqara/AnalogOutput behavior discussed in #783 and #781 makes that distinction consequential, and this implementation still does not suppress virtual IAS enrollment consistently.

A follow-up would need an explicit quirk configuration policy and device-level compatibility evidence for binding, reporting, and IAS enrollment before changing these semantics.

PR #831, opened 2026-07-14
Codex updated: Fix `IndexError` in `issue_cluster_command` for IAS ACE arm response

dmulcahey:
This pull request was created with Codex and is an updated version of the pull request it supersedes.

Original pull request: https://github.com/zigpy/zha/pull/323

Fix issue on sending IasAce:arm_response which does not return a tuple but only a single state but in general to be more fail tolerante as we can't always guarantee the result from the multitude of Zigbee devices

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -18,11 +18,11 @@ import zigpy.profiles.zha
 import zigpy.types
 from zigpy.typing import UNDEFINED
 from zigpy.zcl import ClusterType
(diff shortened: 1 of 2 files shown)


dmulcahey:
Closing as already addressed upstream by #872 (commit f6273ae7). The current implementation reads typed response status by name, fixing the IAS ACE arm-response `IndexError` while preserving more response shapes than this replacement. This PR no longer provides an additional fix on current `dev`.

PR #841, opened 2026-07-17
Keep startup polling failures isolated

dmulcahey:
One unexpected startup device-refresh failure can end the supervising gather while sibling refreshes keep running, and skip re-enabling normal polling. Wait for every device outcome, report failures/cancellations per device, and restore `allow_polling` when the startup operation exits. Cancellation of the overall operation still cancels and awaits its children.

Following the review suggestion, remove ZHA's whole-device concurrency limit and express startup work as LOW-priority radio requests. Keep the ambient priority context and forward explicit LOW through aggregated attribute reads, so packet-priority consumers receive it too. Existing initialization callers retain their default behavior.

Tests cover sibling supervision, failure recovery, cancellation, eligible-device selection, and priority propagation. This is not a measured startup-speed improvement. Radio backends own request scheduling; backends that do not use zigpy's limiter no longer have a separate ZHA whole-device cap.

Validation against `dev` at `66603431339afe37fa0048b70ff31d77dceb8f95`: Python 3.12 full suite, **1385 passed**, coverage above the 95% project gate; full pre-commit (codespell, Ruff, formatting, mypy, lock check) passed. Relevant regression checks fail on the unchanged base. GitHub CI for Python 3.12/3.13/3.14 is reported separately on the PR.

### DIFF
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -57,6 +57,7 @@ from zha.application.platforms.sensor.device_class import (
 from zha.application.platforms.switch import Switch
 from zha.exceptions import ZHAException
 from zha.quirks import DeviceRegistry
+from zha.zigbee.cluster_config import AggregatedAttrConfig, AggregatedClusterConfig
 from zha.zigbee.device import (
     ClusterBinding,
     Device,
@@ -122,6 +123,60 @@ def zigpy_device_mains(zha_gateway: Gateway, with_basic_cluster: bool = True):
             descriptor_capability_field=zdo_t.NodeDescriptor.DescriptorCapability.NONE,
         ),
     )
+
+
+@pytest.mark.parametrize(
+    ("request_priority", "expected_read_attribute_options"),
+    [
+        (None, {}),
+        (
+            zigpy.types.PacketPriority.LOW,
+            {"priority": zigpy.types.PacketPriority.LOW},
+        ),
+    ],
+)
+async def test_initialize_request_priority(
+    zha_gateway: Gateway,
+    request_priority: int | None,
+    expected_read_attribute_options: dict[str, int],
+) -> None:
+    """Test device initialization forwards request priority to attribute reads."""
+    zha_device = zha_gateway.get_or_create_device(zigpy_device_mains(zha_gateway))
+    cluster = mock.MagicMock()
+    cluster.read_attributes = AsyncMock()
+    cluster_configs = {
+        (1, 6, True): AggregatedClusterConfig(
+            cluster=cluster,
+            attributes={
+                "cached_attribute": AggregatedAttrConfig(read_on_startup=False),
+                "fresh_attribute": AggregatedAttrConfig(read_on_startup=True),
+            },
+        )
+    }
+
+    with patch(
+        "zha.zigbee.device.aggregate_cluster_configs",
+        return_value=cluster_configs,
+    ):
+        await zha_device.async_initialize(
+            from_cache=False,
+            request_priority=request_priority,
+        )
+
+    assert cluster.read_attributes.await_args_list == [
+        call(
+            ["cached_attribute"],
+            allow_cache=True,
+            only_cache=False,
+            **expected_read_attribute_options,
+        ),
+        call(
+            ["fresh_attribute"],
+            allow_cache=False,
+            only_cache=False,
+            **expected_read_attribute_options,
+        ),
+    ]
 
 
 async def _send_time_changed(zha_gateway: Gateway, seconds: int):
--- a/tests/test_device.py
+++ b/tests/test_device.py
@@ -62,6 +62,7 @@ from zha.application.platforms.sensor.device_class import (
 from zha.application.platforms.switch import Switch
 from zha.exceptions import ZHAException
 from zha.quirks import DeviceRegistry
+from zha.zigbee.cluster_config import AggregatedAttrConfig, AggregatedClusterConfig
 from zha.zigbee.device import (
     ClusterBinding,
     Device,
@@ -127,6 +128,60 @@ def zigpy_device_mains(zha_gateway: Gateway, with_basic_cluster: bool = True):
             descriptor_capability_field=zdo_t.NodeDescriptor.DescriptorCapability.NONE,
         ),
     )
+
+
+@pytest.mark.parametrize(
+    ("request_priority", "expected_read_attribute_options"),
+    [
+        (None, {}),
+        (
+            zigpy.types.PacketPriority.LOW,
+            {"priority": zigpy.types.PacketPriority.LOW},
+        ),
+    ],
+)
+async def test_initialize_request_priority(
+    zha_gateway: Gateway,
+    request_priority: int | None,
+    expected_read_attribute_options: dict[str, int],
+) -> None:
+    """Test device initialization forwards request priority to attribute reads."""
+    zha_device = zha_gateway.get_or_create_device(zigpy_device_mains(zha_gateway))
+    cluster = mock.MagicMock()
+    cluster.read_attributes = AsyncMock()
+    cluster_configs = {
+        (1, 6, True): AggregatedClusterConfig(
+            cluster=cluster,
+            attributes={
+                "cached_attribute": AggregatedAttrConfig(read_on_startup=False),
+                "fresh_attribute": AggregatedAttrConfig(read_on_startup=True),
+            },
+        )
+    }
+
+    with patch(
+        "zha.zigbee.device.aggregate_cluster_configs",
+        return_value=cluster_configs,
+    ):
+        await zha_device.async_initialize(
+            from_cache=False,
+            request_priority=request_priority,
+        )
+
+    assert cluster.read_attributes.await_args_list == [
+        call(
+            ["cached_attribute"],
+            allow_cache=True,
+            only_cache=False,
+            **expected_read_attribute_options,
+        ),
+        call(
+            ["fresh_attribute"],
+            allow_cache=False,
+            only_cache=False,
+            **expected_read_attribute_options,
+        ),
+    ]
 
 
 async def _send_time_changed(zha_gateway: Gateway, seconds: int):
--- a/tests/test_gateway.py
+++ b/tests/test_gateway.py
@@ -59,6 +59,26 @@ ZIGPY_DEVICE_BASIC = {
 }
 
 
+def create_mains_powered_startup_device(
+    gateway: Gateway,
+    *,
+    ieee: str,
+    nwk: int,
+) -> Device:
+    """Create a recent mains-powered device for startup polling tests."""
+    zigpy_device = create_mock_zigpy_device(
+        gateway,
+        ZIGPY_DEVICE_BASIC,
+        ieee=ieee,
+        nwk=nwk,
+    )
+    assert zigpy_device.node_desc is not None
+    zigpy_device.node_desc.mac_capability_flags |= (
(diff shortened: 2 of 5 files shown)


puddly:
An alternative approach here would be to disable concurrency limiting entirely and make sure that device state polling is done with low priority:

```python
async with self.request_priority(t.PacketPriority.LOW):
    # code to poll a device
```

Maybe we can add this as a kwarg to `async_initialize` or something?

Zigpy has a mechanism to control concurrency if a request is properly tagged with the above context manager.

dmulcahey:
@puddly Thanks — I updated the PR to take this approach. ZHA no longer limits complete device-initialization pipelines during startup. It starts every eligible initializer as LOW-priority work and delegates actual radio-request scheduling to zigpy and the radio backend.

The revised call site is [here](https://github.com/zigpy/zha/blob/28b6884941d917c4d369ccfe00d3d2c2d248d6f1/zha/application/gateway.py#L336-L396). In abbreviated form, it is now:

```python
startup_polling_priority = t.PacketPriority.LOW

async with self.request_priority(startup_polling_priority):
    initialization_results = await asyncio.gather(
        *(
            device.async_initialize(
                from_cache=False,
                request_priority=startup_polling_priority,
            )
            for device in online_devices
        ),
        return_exceptions=True,
    )
```

### Why removing the ZHA-level limiter is the better boundary

The previous implementation inspected zigpy's private `_concurrent_requests_semaphore.max_concurrency`, subtracted four, and used that result to limit entire `Device.async_initialize()` calls. That mixed two different kinds of concurrency:

- device initialization also includes local discovery, cache processing, and entity work; and
- radio request concurrency is the scarce resource that actually needs scheduling.

Zigpy now owns the latter concern. Its [request-priority context](https://github.com/zigpy/zigpy/blob/dbfb1a6141776709831f541252ea0d2ba8f79071/zigpy/application.py#L155-L163) supplies the ambient priority, and its [global request limiter](https://github.com/zigpy/zigpy/blob/dbfb1a6141776709831f541252ea0d2ba8f79071/zigpy/application.py#L1023-L1060) resolves that context when a backend submits a request without an explicit priority. The limiter itself is a [priority-aware cascading limiter](https://github.com/zigpy/zigpy/blob/dbfb1a6141776709831f541252ea0d2ba8f79071/zigpy/datastructures.py#L349-L503), rather than a single undifferentiated semaphore.

With zigpy's [default concurrency of 8](https://github.com/zigpy/zigpy/blob/dbfb1a6141776709831f541252ea0d2ba8f79071/zigpy/config/defaults.py#L12-L15) and [25% LOW / 50% NORMAL / 25% HIGH capacity fractions](https://github.com/zigpy/zigpy/blob/dbfb1a6141776709831f541252ea0d2ba8f79071/zigpy/config/defaults.py#L53-L57), the cumulative admission thresholds are 2 for LOW, 6 for LOW plus NORMAL, and 8 when HIGH is included. These are cascading priority thresholds, not a strict total-concurrency ceiling for every possible arrival ordering, but the important distinction is that startup traffic now enters the LOW tier instead of consuming an arbitrary number of whole-device slots.

For common configured values, the structural difference looks like this:

| Configured concurrency | Old ZHA initializer cap (`max(1, C - 4)`) | Zigpy LOW request tier (25%) |
| ---: | ---: | ---: |
| 4 | 1 complete initializer | 1 radio request |
| 8 | 4 complete initializers | 2 radio requests |
| 12 | 8 complete initializers | 3 radio requests |

Those columns are intentionally not performance equivalents: the old limit covered the entire initialization pipeline, while the new tier governs individual radio requests. The latter is the abstraction we actually want. The old ZHA reservation heuristic came from [ZHA #510](https://github.com/zigpy/zha/pull/510); zigpy subsequently gained its priority-aware request limiting in [zigpy #1635](https://github.com/zigpy/zigpy/pull/1635), so the scheduling responsibility can now live at the lower layer.

> **Important — zigpy-ziggurat needs explicit priority propagation, not only the context manager.**

There is a subtle distinction between the priority used to acquire a local limiter and the priority stored on the outgoing packet:

1. The LOW `ContextVar` is inherited by the tasks created inside the `async with` block. Zigpy defines its default request context [here](https://github.com/zigpy/zigpy/blob/dbfb1a6141776709831f541252ea0d2ba8f79071/zigpy/application.py#L107-L110).
2. If a cluster call leaves `priority=None`, [`zigpy.device.Device.request()`](https://github.com/zigpy/zigpy/blob/dbfb1a6141776709831f541252ea0d2ba8f79071/zigpy/device.py#L609-L692) resolves the ambient value for its per-device limiter, but still forwards the original `None` to `ControllerApplication.request()`.
3. [`ControllerApplication.request()` builds the packet](https://github.com/zigpy/zigpy/blob/dbfb1a6141776709831f541252ea0d2ba8f79071/zigpy/application.py#L1077-L1151) with that value, so ambient LOW alone can still produce `ZigbeePacket.priority is None`.
4. The [zigpy-ziggurat 1.0.1 send path](https://github.com/zigpy/zigpy-ziggurat/blob/635c76acc1d390652056edb35cad8ea4b4b7cc25/zigpy_ziggurat/zigbee/application.py#L899-L942) does not re-resolve zigpy's ambient context. It serializes `packet.priority`, and deliberately converts `None` to `0` (NORMAL). The relevant [`SendAps` priority field](https://github.com/zigpy/zigpy-ziggurat/blob/635c76acc1d390652056edb35cad8ea4b4b7cc25/zigpy_ziggurat/zigbee/commands.py#L217-L235) therefore receives NORMAL unless the packet was tagged explicitly.

That would make a context-only version look correct inside ZHA while Ziggurat still received the startup reads as normal interactive traffic. The revised implementation closes that gap:

- [`Device.async_initialize()` now has an optional keyword-only `request_priority`](https://github.com/zigpy/zha/blob/28b6884941d917c4d369ccfe00d3d2c2d248d6f1/zha/zigbee/device.py#L1207-L1226).
- [`initialize_cluster_configs()` adds the priority to both cache-permitted and fresh attribute reads when one is supplied](https://github.com/zigpy/zha/blob/28b6884941d917c4d369ccfe00d3d2c2d248d6f1/zha/zigbee/cluster_config.py#L245-L292). This matters because a cache-permitted read can still fall through to the radio when the value is absent.
- Existing callers pass no priority, and the helper omits the keyword entirely when it is `None`, preserving their current behavior.
- The ambient LOW context remains in place for nested request paths and backends that resolve it themselves.

[`PacketPriority.LOW` is `-1`, while NORMAL is `0`](https://github.com/zigpy/zigpy/blob/dbfb1a6141776709831f541252ea0d2ba8f79071/zigpy/types/named.py#L583-L590). On the server side, the current Ziggurat source carries the signed [wire priority](https://github.com/zigpy/ziggurat/blob/4df43c959a1952a62ff4e7763b574ce02ed6dd9e/crates/ziggurat-protocol/src/wire.rs#L365-L378) through its [bridge conversion](https://github.com/zigpy/ziggurat/blob/4df43c959a1952a62ff4e7763b574ce02ed6dd9e/crates/ziggurat-protocol/src/bridge.rs#L263-L297), maps `-1` to `UserLow` and `0` to `UserNormal` in the [Zigbee stack](https://github.com/zigpy/ziggurat/blob/4df43c959a1952a62ff4e7763b574ce02ed6dd9e/crates/ziggurat-driver/src/zigbee_stack.rs#L101-L125), and schedules transmissions through its [priority heap](https://github.com/zigpy/ziggurat/blob/4df43c959a1952a62ff4e7763b574ce02ed6dd9e/crates/ziggurat-driver/src/zigbee_stack.rs#L333-L459) when [APS work is enqueued](https://github.com/zigpy/ziggurat/blob/4df43c959a1952a62ff4e7763b574ce02ed6dd9e/crates/ziggurat-driver/src/zigbee_stack/aps.rs#L310-L371). I am describing the current server source separately here; the pinned zigpy-ziggurat client commit above does not imply that it pins this server revision.

### Backend behavior

| Radio library | How startup priority is handled after this change |
| --- | --- |
| [bellows](https://github.com/zigpy/bellows/blob/4ee82da31b2cbdab9ccdb10c9d4ddd28f8645d39/bellows/zigbee/application.py#L957-L1015) | Its send path uses zigpy's global priority-aware limiter. Ambient LOW is available, and the revised reads also carry explicit LOW. |
| [zigpy-znp](https://github.com/zigpy/zigpy-znp/blob/614f74e0a02f7b7b89c589f6fa0a134edc1b377a/zigpy_znp/zigbee/application.py#L861-L899) | Same global-limiter behavior. |
| [zigpy-deconz](https://github.com/zigpy/zigpy-deconz/blob/d7d73ba0d1019b97490cecfe0fda391370ede7fc/zigpy_deconz/zigbee/application.py#L545-L557) | Same global-limiter behavior. |
| **[zigpy-ziggurat](https://github.com/zigpy/zigpy-ziggurat/blob/635c76acc1d390652056edb35cad8ea4b4b7cc25/zigpy_ziggurat/zigbee/application.py#L899-L942)** | **The explicit argument is essential: it puts `-1` on the packet so the server receives `UserLow`, rather than converting a missing value to NORMAL.** |
| [zigpy-zigate](https://github.com/zigpy/zigpy-zigate/blob/3d82221ac54b8d851ac6d07737eb6970b9334d5e/zigpy_zigate/zigbee/application.py#L294-L316) / [zigpy-xbee](https://github.com/zigpy/zigpy-xbee/blob/f5f9d6f92c76b9c7835a6c307658ca419a8f8452/zigpy_xbee/zigbee/application.py#L257-L310) | These send paths do not currently use zigpy's global priority limiter. We are intentionally not retaining a duplicate ZHA whole-device limiter solely for these backends. |

The explicit propagation currently covers the radio-producing work in `Device.async_initialize()`: the aggregated attribute reads. The current entity-level `async_initialize_cluster()` override is local/cache-only. If a future entity hook adds network I/O, that new path should either accept the explicit priority as well or zigpy-ziggurat should resolve the ambient priority before serializing a packet. That is a future-proofing consideration, not a reason to broaden this focused change.

### Concrete behavior on a larger network

Consider 50 recently seen mains-powered devices during startup:

- ZHA now starts all 50 initialization pipelines, so local discovery and cache work are not artificially serialized behind a radio-derived number.
- On bellows, ZNP, or deCONZ, each actual startup request enters zigpy's LOW tier. Normal and high-priority work retain their priority-aware capacity instead of waiting behind an undifferentiated ZHA batch.
- On Ziggurat, the startup attribute requests carry `priority=-1`; the server classifies them as `UserLow`, so normal interactive work can outrank queued startup transmissions.
- Starting all 50 initializers therefore does **not** mean that a priority-aware backend transmits 50 frames concurrently. Scheduling happens where the individual requests are visible.

### F4 failure and cancellation semantics remain intact

Removing the outer concurrency helper does not undo the original corrective behavior:

- `asyncio.gather(..., return_exceptions=True)` retains one ordered outcome per device and waits for every sibling initializer, so one quirk or device failure cannot orphan the rest of the batch.
- A child that cancels independently is returned as that device's result and logged separately from an ordinary exception.
- Cancellation of the overall startup operation still propagates through `gather`, cancels unfinished child tasks, and is re-raised by the wrapper.
- [`allow_polling` is restored in `finally`](https://github.com/zigpy/zha/blob/28b6884941d917c4d369ccfe00d3d2c2d248d6f1/zha/application/gateway.py#L398-L425), so an unexpected device or orchestration failure cannot leave polling globally disabled for the rest of the gateway lifetime.

So I agree with the suggested direction and have changed the implementation accordingly: ZHA expresses startup polling as LOW-priority work, zigpy/backends own radio scheduling, and Ziggurat receives the priority explicitly on the packet rather than accidentally treating startup reads as NORMAL.

PR #849, opened 2026-07-28
Properly tear down group entities on membership changes and group removal

TheJulianJES:
Group entity teardown was skipped in several places, leaking member entity subscriptions: member entities kept references to dead group entities and kept invoking their debounced updates.

### What was broken

- `discover_group_entities()` cleared `group_entities` with a bare `.clear()` when a group dropped below 2 members, without calling `on_remove()` or `unregister_group_entity()` — so the `Debouncer` and every member subscription stayed alive.
- It also never removed a *platform's* group entity when that platform alone dropped below 2 eligible members (e.g. a mixed light+switch group losing one of its two switches). The stale entity stayed registered and functional, backed by a single member.
- `Gateway.group_removed()` popped the group without calling `Group.on_remove()`, unlike `shutdown()` and `device_removed()`.
- `Group.on_remove()` only removed the group entities themselves, leaving the member entity subscriptions in `_entity_unsubs` dangling.

### What this changes

- `discover_group_entities()` now tears stale group entities down properly — `unregister_group_entity()` plus a scheduled `on_remove()` — both when the whole group drops below 2 members and when a single platform loses eligibility. Coordinator members are already excluded from the per-platform counts, so a "coordinator + one light" group now correctly loses its group entity too.
- `Gateway.group_removed()` schedules `Group.on_remove()`, so removing a group tears down its entities even when the members were not removed first.
- `Group.on_remove()` releases any remaining member entity subscriptions, and unregisters a group entity even if that entity's own `on_remove()` raised — otherwise `_group_entities` and `_entity_unsubs` could desync and a later `unregister_group_entity()` would raise `KeyError`.
- `unregister_group_entity()` verifies the exact entity instance is registered, so a delayed removal cannot unregister a *recreated* group entity that has the same unique id.

### Notes

- Removing a member *device* (rather than a member endpoint) still leaves a stale group entity behind: zigpy's `_remove_device()` does not remove the device's endpoints from groups, so no `group_member_removed` event fires and `Group.members` keeps its cached value. That is pre-existing and a different code path (`Gateway.device_removed`), so it is deliberately out of scope here.
- This is **not** a prerequisite for home-assistant/core#162307 ("Create device registry entries for ZHA group entities"). Home Assistant rebuilds its group entity set from the `GroupInfo.entities` snapshot on every group event, and that snapshot is correct either way. The two are complementary: without this PR, a platform that drops below two eligible members keeps a stale group entity that Home Assistant faithfully re-adds.

### DIFF
--- a/tests/test_gateway.py
+++ b/tests/test_gateway.py
@@ -41,6 +41,7 @@ from zha.application.gateway import (
 from zha.application.helpers import ZHAData
 from zha.application.platforms import GroupEntity
 from zha.application.platforms.light.const import EFFECT_OFF, LightEntityFeature
+from zha.const import STATE_CHANGED
 from zha.quirks import DeviceMatch, DeviceRegistry, ModelInfo, QuirkRegistryEntry
 from zha.zigbee.device import Device, DeviceEntityAddedEvent, DeviceEntityRemovedEvent
 from zha.zigbee.group import Group, GroupMemberReference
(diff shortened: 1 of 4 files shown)


TheJulianJES:
@zigpy-review-bot Review and check if anything else is needed, or if this should be merged as is. Push any changes needed. Consider the linked HA Core PR https://github.com/home-assistant/core/pull/162307 and this branch: https://github.com/home-assistant/core/compare/dev...TheJulianJES:core:zigpy-bot/zha-group-devices.

Check if that is complete as well. Do review rounds on both this PR and the HA Core PR, but with Opus agent (not Fable) and also with Copilot: 5.6 Sol. You can also push more changes to the linked HA Core branch, but do not force-push that one. If needed, create another branch based off that one. You can force-push this ZHA PR if needed though (or create other issues or PRs for other things discovered along the way). Do verify everything with review rounds before pushing. You can also update this PR description (but again: verify before).

Also, is this PR required for the Core PR?

PR #857, opened 2026-08-03
Import new device diagnostics

puddly:


tests/data/devices/somfy-ysia-variation-5c-0xffffffff.json
```diff
@@ -0,0 +1,708 @@
...
+        "entity_category": "config",
+        "entity_registry_enabled_default": true,
+        "enabled": true,
+        "primary": false,
+        "extra_state_attribute_names": [],
+        "device_ieee": "ab:cd:ef:12:ec:84:26:cd",
+        "endpoint_id": 232,
+        "available": true,
+        "group_id": null,
+        "installed_version": "0xffffffff",
```
TheJulianJES, line 695: 🤨

puddly: There I guess are two sources for this: the `current_file_version` attribute (technically this is a valid value!) and as part of a `QueryNextImage` command (invalid). I wonder which it is?

Issue #725, opened 2026-03-24
Entity discovery issues


TheJulianJES:
### `_compute_primary_entity` can get stuck after entity removal

Weight-election losers are set to `_attr_primary = False`, which permanently excludes them from future elections (filtered by `_attr_primary is not False`). If the winner is later removed, no remaining entity can become primary.

Fixing this requires distinguishing quirk-set `_attr_primary` from weight-computed values, since both use the same field. Needs a separate PR.

### `_discover_new_entities()` in `async_configure` leaves entities half-added

`async_configure()` calls `_discover_new_entities()`, which runs `entity.on_add()` (e.g. RSSI sensor registers a global updater listener) and queues entities in `_pending_entities` — but never calls `_add_pending_entities()`. These entities are cleaned up when `async_initialize()` runs later, but until then they have active side effects without being finalized.

### `is_supported_in_list` receives `self` in the entity list

`recompute_entities` passes the full entity list including the entity being evaluated. Implementations (RSSI sensor, button) already filter out `self` defensively. A prior commit on this branch explicitly ensured this works, but the caller could exclude `self` to avoid requiring defensive checks in every implementation.

_Originally posted by @TheJulianJES in https://github.com/zigpy/zha/issues/517#issuecomment-4121937584_

PR #861, opened 2026-08-06
Separate computed primary entity state from `_attr_primary`

TheJulianJES:
Fixes the first item of #725.

The primary entity election previously stored its result in `_attr_primary`, the same field entity classes and quirks use to explicitly mark an entity as (not) primary. Conflating the two caused the election to get stuck:

- Election losers were set to `_attr_primary = False`, permanently excluding them from all future elections (the candidate filter skips `_attr_primary is False`). If the winner was later removed, no remaining entity could ever become primary again.
- A previous winner (`_attr_primary = True`) looked "explicitly primary" to later elections, which then short-circuited, so a stronger candidate appearing later could never take over. If a genuinely explicit primary entity appeared alongside a previous winner, the `assert not explicitly_primary` sanity check would crash instead.

### Changes

Election state is a device concern, not an entity concern (design settled in the review discussion below — the first revision stored a per-entity computed flag instead):

- `_attr_primary` now only ever holds the explicit declaration (entity class attribute or quirk `primary=True/False`) and is never written by the election.
- The election result is a single device-owned reference, `Device._primary_entity` (exposed as `Device.primary_entity`), set for both the explicit and the weight-based paths and cleared when the referenced entity is removed. The election recomputes it from scratch, so stale per-entity primary state is no longer representable.
- `PlatformEntity.primary` returns `_attr_primary` when set, falling back to `device.primary_entity is self` — so an explicit `True`/`False` from a quirk or entity class can never be overridden by the election. Group entities (no device, no election) report their explicit value or `False`.
- The `primary` setter is removed; nothing outside the election ever wrote primary state.

### Tests

- `test_primary_entity_reelection`: a mock smart plug with `OnOff` + `IasZone`. Marking `on_off` as unsupported in the ZCL attribute cache makes the switch natively unsupported, so `recompute_entities()` removes it through the real removal path — the IAS zone (previous election loser) now wins the re-election. Writing an `on_off` value clears the unsupported flag, and the rediscovered switch wins back primary.
- `test_primary_entity_election_disabled_winner`: a disabled previous winner loses its computed primary state on the next recomputation, so only the runner-up is primary. After being re-enabled, it wins back the election.
- `test_primary_entity_election_explicit_primary_takes_over`: an entity marked explicitly primary (as a quirk would) takes over from a previously computed winner on the next recomputation, instead of tripping the sanity assert.

All three tests fail on `dev` without the fix, and passed unchanged across the switch from per-entity flags to the device-owned reference — they pin behavior, not storage.

### Possible future TODOs for other PRs

1. **Trim the `primary_weight` property:** pure pass-through to `_attr_primary_weight` with a single caller (the election), kept for now to match the repo-wide `_attr_*`-plus-property idiom.
2. **Multiple explicit primaries still assert:** two quirk entities with `primary=True` hit `assert not explicitly_primary` and crash device init. Pre-existing; could be downgraded to a warning + tie handling.


zha/zigbee/device.py
```diff
@@ -1613,13 +1613,19 @@ def _compute_primary_entity(self, entities: Sequence[PlatformEntity]) -> None:
...
 
         if len(explicitly_primary) == 1:
             self.debug(
                 "Device has a single explicitly primary entity,"
                 " not performing weight matching"
             )
+
+            # Clear any previously computed primary state, so the winner of an
+            # earlier election cannot coexist with the explicitly primary entity
+            for entity in entities:
```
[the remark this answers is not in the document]
TheJulianJES: Fixed: [`ed9d98f` (this PR)](https://github.com/zigpy/zha/pull/861/commits/ed9d98f6b8ea08157e01bcb2c483bb8f8e3aafcc)

TheJulianJES:
Hmm, I think enabled and disabled entities should behave the same, in that a disabled entity could be the primary entity still and not allow others to take its primary spot? But on the other side, I do kind of see some benefit to excluding (user-)disabled entities from the primary entity election?

TheJulianJES:
The below PR will fix that enabled/disabled issue and should be rebased + merge after this PR is merged.
- https://github.com/zigpy/zha/pull/862

The current logic of checking `enabled` doesn't make any sense because of multiple reasons, including that we don't recompute the primary entity when enabling/disabling an entity and HA disables the LQI/RSSI entities too late on the ZHA side when the primary entity computation is already done. So effectively, the computation never really used `enabled` anyway, hence the above PR.

puddly:
Ah, you're very right, this attribute does indeed perform double duty.

I think the TODO you listed is a good alternative approach: denormalization is the problem here because primary entity computation is a device problem, not an entity problem, and IMO should be something the device figures out and maintains. This would retain `_attr_primary` as an entity-level hint and then let the device object itself hold a `PlatformEntity | None` reference to a primary entity. What do you think?

TheJulianJES:
Yeah, I think that would be the better approach. I can have a look at that later.

zha/application/platforms/__init__.py
```diff
@@ -642,6 +629,14 @@ def device(self) -> Device:
         """Return the device."""
         return self._device
 
+    @property
+    def primary(self) -> bool:
+        """Return if the entity is the primary device control."""
+        if self._attr_primary is not None:
+            return self._attr_primary
```
puddly, line 636: Is this `primary` property used anywhere?

TheJulianJES: Yes, [in `BaseEntity.state`](https://github.com/TheJulianJES/zha/blob/79a230d8f32a878622ffc93a88dd5d0a4f9fea29/zha/application/platforms/__init__.py#L423) (`BaseEntityState.primary`). That's used by Home Assistant then.

This [`PlatformEntity.primary`](https://github.com/TheJulianJES/zha/blob/79a230d8f32a878622ffc93a88dd5d0a4f9fea29/zha/application/platforms/__init__.py#L632-L638) property overrides the [`primary` property in `BaseEntity`](https://github.com/TheJulianJES/zha/blob/79a230d8f32a878622ffc93a88dd5d0a4f9fea29/zha/application/platforms/__init__.py#L333-L336). Groups don't have a device reference.

puddly: Ah, right, we do keep it as entity-level state.

TheJulianJES: Yeah, I think that's also what we want, right? HA needs it at entity level for the name, so if we keep it at the device level only, HA would have to reference entity <-> device, which may be slightly more complicated (and also because we'd need device updates to update HA entities then).

But I guess both variants generally make sense.

tests/test_device.py
```diff
@@ -973,6 +973,122 @@ async def test_primary_entity_weight_0_not_elected(zha_gateway: Gateway) -> None
...
+
+    assert not switch.primary
+    assert ias_zone.primary
+
+    # When the switch is re-enabled, it wins back the election
+    switch.enable()
+    await zha_device.recompute_entities()
+
+    assert switch.primary
+    assert not ias_zone.primary
```
TheJulianJES, line 1064: Note: This current/pre-existing behavior, of disabled/enabled entities (trying to) influence the primary entity, is bad. The test just asserts it. After this PR, https://github.com/zigpy/zha/pull/862 will adjust it to remove that (will need to be rebased on this one – small PR then).

With that PR, it's also accurate to how it currently works in HA. HA calls `disable()` on disabled entities too late on startup. So already currently, there's no practical difference for the primary entity when an entity is enabled or disabled. But it's a bug it already works that way in Core (currently, as we don't ever call `recompute_entities()` automatically yet).

If we ever do want enabled/disabled entities to intentionally affect the primary entity (which we likely don't because it doesn't really make sense, as names would just swap "randomly" if the user enables/disables entities), we need to rework the logic of HA enabling/disabling entities. Either way, that's not for this PR, nor #862, which just converts the current ZHA behavior to work more like how Core uses it.

zha/zigbee/device.py
```diff
@@ -1141,6 +1147,8 @@ async def _remove_entity(
             # rediscovers the same unique_id would skip the replacement and
             # leave the stale entity shadowing it indefinitely.
             del self._platform_entities[key]
+            if entity is self._primary_entity:
+                self._primary_entity = None
```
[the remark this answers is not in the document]
TheJulianJES: I also took a second look at this before pushing, so I guess a comment makes sense then: https://github.com/zigpy/zha/pull/861/commits/5487b837b0ea5944a6264a138433f3582446fe80

PR #888, opened 2026-09-04
Ignore unavailable members when computing light group state

RReverser:
`LightGroup.update()` derives on/off from every member's last reported state, with no availability check. A member that drops off the network while on keeps voting "on" indefinitely. The group then stays marked as "on" even once every reachable member is off, and only clears once the stale member itself reports again or the whole group entity is force refreshed some other way.

`_make_members_assume_group_state()` already skips unavailable members when writing optimistic state after a command. `update()` should apply the same rule when reading it back.

### DIFF
--- a/tests/test_light.py
+++ b/tests/test_light.py
@@ -2324,3 +2324,64 @@ async def test_turn_off_cancellation_cleans_up_transition_flag(
         await task
 
     assert entity.is_transitioning is False
+
+
+async def test_group_state_ignores_unavailable_members(zha_gateway: Gateway) -> None:
+    """Test that an unavailable member does not keep the group on."""
+
+    coordinator = await coordinator_mock(zha_gateway)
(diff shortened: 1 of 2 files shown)


RReverser:
Ping? This one should be a straightforward fix.

RReverser:
Is this good to go then?

puddly (approved):

RReverser:
So this has couple of approvals now, can someone merge please?

puddly:
Please be patient. We'll merge it in time for the next Core beta.

RReverser:
Fair enough, sorry for the pings.

Being an OSS maintainer myself, I've been on both sides of this - both when the pings become too frequent and annoying, and when I genuinely forget about open PRs for longer than they deserve heh. So I just wasn't sure if these PRs are stuck and whether they need either me or the maintainers to do anything. 

I'll shut up now :)

PR #889, opened 2026-09-04
Add opt-in option to adjust only lit members of a light group

RReverser:
Brightness and colour changes on a `LightGroup` always go through `Move_To_Level_With_On_Off` / turn every member on, even members that are currently off. A group-bound wall dimmer sends the plain `Move_To_Level` instead, so its brightness/colour changes only affect lamps that are already lit and leave the rest off. There is no way to get that behaviour from a `light.turn_on` service call today, which makes any HA slider widget that controls brightness and/or colour (such as mushroom light cards) turn on the entire group on every adjustment.

This PR adds a new `group_adjust_only_lit_members` option (default `False`) that makes brightness- or colour-only calls on a group that already has a lit member behave the same way: `Move_To_Level` and `Move_To_Color_Temp`/`Move_To_Color`, no On command. A bare `light.turn_on()`, or one sent while the whole group is off, keeps turning every member on as before, so "turn on at 40%" still works from a fully off group.

### DIFF
--- a/tests/test_light.py
+++ b/tests/test_light.py
@@ -2324,3 +2324,138 @@ async def test_turn_off_cancellation_cleans_up_transition_flag(
         await task
 
     assert entity.is_transitioning is False
+
+
+async def test_group_adjust_only_lit_members(zha_gateway: Gateway) -> None:
+    """Test the opt-in group_adjust_only_lit_members option."""
+
+    # Read once by recompute_capabilities() when the group entity is created, so
+    # it must be set before the group exists; toggling it afterwards has no
+    # effect without an explicit recompute_capabilities() call.
+    zha_gateway.config.config.light_options.group_adjust_only_lit_members = True
+
+    await coordinator_mock(zha_gateway)
+    device_light_1 = await device_light_1_mock(zha_gateway)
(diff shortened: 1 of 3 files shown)


puddly:
Thanks!

As an alternative, we could also expose this as a group-specific config entity once https://github.com/home-assistant/core/pull/162307 lands. This would allow you to configure behavior per-group.

RReverser:
> As an alternative, we could also expose this as a group-specific config entity once [home-assistant/core#162307](https://github.com/home-assistant/core/pull/162307) lands. This would allow you to configure behavior per-group.

No harm in doing that, but I'm genuinely unsure of any situations where I _wouldn't_ want this behaviour - it feels a lot more natural and in line with how zigbee buttons already work. The only reason I kept this option as off-by-default is backward compatibility for existing setups.

I worked around this limitation by creating fake HA entries for every zigbee group in my own setup, but it got cumbersome enough to maintain that I decided upstreaming is worth it.

When that other PR lands, we could surely add per-group overrides for both "assume state of group" and "adjust only lights which are already On" but I'd rather not block on it as a dependency.

RReverser:
Anything else I can do on this one to make it more likely to land? It would fix such a common annoyance.

